PR #5, opened 2026-01-25
Add simulation testing plan

darryl-ctrl:
### Motivation
- Provide a practical evaluation of the original end-to-end auto-fix simulation idea and propose a safer, faster, and more maintainable testing workflow for running 20 conflict-resolution simulations.

### Description
- Add `simulation-plan.md` which includes feasibility/efficiency/cost scoring, a phased improved plan (deterministic simulation execution + human-in-the-loop fix workflow), implementation steps, CI integration guidance, and expected deliverables.

### Testing
- Documentation-only change; no automated tests were run against code as part of this PR.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_697589ced060832d947ea6026dd6674f)

### DIFF
--- a/simulation-plan.md
+++ b/simulation-plan.md
@@ -18,28 +18,52 @@
 - **Engineering overhead:** Building and maintaining auto-fix automation is costly, and the risk of unsafe changes is high.
 
 **Overall Score: 4/10**
-- Good ambition but too complex and fragile for initial implementation. It couples e2e testing with automated code changes, which is not realistic without strong guardrails.
+- Good ambition but too complex and fragile for initial implementation. It couples e2e testing with automated code changes, which is not realistic without strong guardrails. It also doesn’t directly target the current blocker (mobile UI workflow gaps).
 
 ---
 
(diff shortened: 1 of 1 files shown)


shantamg:
@darryl-ctrl This is interesting, though I'm trying to understand what problem you are solving. The core idea of having a test suite for full session scenarios is great, but it only covers the backend stuff and mocks all of the actual LLM stuff so really what this aims to do is make sure backend data flows are correct. This is a good thing (we do already have a dedicated test database and a number of unit tests) and I definitely am planning to have a CI/CD pipeline in github actions to protect code changes from breaking production (though we don't care about that yet), but the main blockers right now keeping us from getting the app working all the way through are front end (mobile) issues. Full integration tests are notoriously tricky, and the main thing I think would be helpful is saved database states that I can load locally in order to jump to a scenario and manually test a certain action (I already have scripts for that).

I can pull this branch and chat with claude about this plan to make sure I'm not missing something, but these are my initial thoughts.

PR #88, opened 2026-04-17
feat(backend): MWF Slack sessions on Bedrock engine

shantamg:
## Summary

Implements [.planning/MILESTONE_SLACK_BEDROCK_SESSIONS.md](.planning/MILESTONE_SLACK_BEDROCK_SESSIONS.md) — replaces the Claude-Code-agent-driven MWF session flow with direct Bedrock calls using the **same pipeline the mobile app uses**. Workspace `CONTEXT.md` files become the single source of truth for Process Guardian voice & stage rules; the backend owns all session state in Prisma.

- Slack user posts \`start\` in \`#mwf-sessions\` → backend creates Session + DMs the user → Sonnet response via same prompt-caching + context-assembly pipeline as mobile.
- Partner pairs with \`join <code>\` → second DM opens, both conversations now run end-to-end on Bedrock.
- EC2 socket-listener becomes a thin dispatcher: POSTs mwf-session events to \`/api/slack/mwf-session\`, keeps 👀/✅/❌ reaction UX.

## What's in

**Phase 1** — Slack client (\`@slack/web-api\`), \`POST /api/slack/mwf-session\` endpoint with shared-secret validation, health endpoint, routes wired in.

**Phase 2** — \`workspace-prompt-builder.ts\` reads \`bot-workspaces/mwf-session/{stages,references}/*.md\` and produces \`{staticBlock, dynamicBlock}\` with the same micro-tag protocol (\`<thinking>\`/\`<draft>\`/\`<dispatch>\`) as \`stage-prompts.ts\`. Falls back to \`stage-prompts.ts\` constants when workspace files are missing.

**Phase 3** — Prisma schema additions (\`User.slackUserId\`, \`Session.slackJoinCode\`, \`SessionSlackThread\` mapping table) + migration. \`slack-session-service.ts\` owns user resolution from Slack profile, session creation, join-code pairing, and stage progression.

**Phase 4** — \`slack-conversation.ts\` runs the full turn: lobby routing → \`assembleContextBundle\` → \`buildSlackStagePrompt\` → \`formatContextForPrompt\` → \`trimConversationHistory\` → \`getSonnetResponse\` → \`parseMicroTagResponse\` → post reply. Stage 0 compact-signal detection advances both users on consent; Stage 1 feel-heard offer gate recorded.

[...]

PR #89, opened 2026-04-17
refactor(slack): collapse workspace-prompt-builder into buildStagePrompt

shantamg:
## Summary

Follow-up to #88, addressing the architectural concern raised in expert review: the workspace-prompt-builder introduced a parallel fork of per-turn dynamic logic (turn counts, intensity thresholds, phase guidance) that would drift away from \`stage-prompts.ts\`. This PR collapses the two paths into one.

**Result**: Slack and mobile now call the exact same \`buildStagePrompt()\` — behavioral parity by construction. A bug fix on either surface propagates to both automatically.

## Changes

- **\`stage-prompts.ts\`**: new option \`BuildStagePromptOptions.surface: 'mobile' | 'slack'\` (default \`'mobile'\`). When \`'slack'\`, \`finalize()\` appends a \`SLACK_FORMATTING_RULES\` block to the static portion. Keeping it in the static block preserves prompt-cache hits across turns.
- **\`slack-conversation.ts\`**: replaces \`buildSlackStagePrompt\` with \`buildStagePrompt(stage, context, { surface: 'slack', isOnboarding })\`.
- **Deleted**: \`workspace-prompt-builder.ts\` and its test (−543 lines). The \`.md\` files under \`bot-workspaces/mwf-session/\` remain as human-readable reference docs; they still drive the Claude-agent workspaces but are no longer runtime input to the backend.
- **\`slack-client.ts\`**: new \`toSlackMrkdwn()\` post-processor applied in \`postMessage()\` — belt-and-suspenders for the cases the model still emits standard Markdown. Fenced code blocks pass through untouched.
- **\`/api/slack/health\`**: drops the workspace-status block.

## Why the prompt rules live in the static block

Prompt caching keys off the static block verbatim. If Slack-formatting rules went in the dynamic block, the cache would miss every turn. Since the rules are identical per-stage, they live in static and amortize nicely across turns.

## Tests

- \`+5\` stage-prompts tests for the \`surface\` option (mobile default, slack appends rules, dynamic block parity, protocol preserved, covers all stages incl. 2B/21)
- \`+11\` tests for \`toSlackMrkdwn\` (bold variants, links, headers, bullets, code-block passthrough, mixed-markdown integration)
- \`npm run check\` clean
- Full backend suite unchanged (same 2 pre-existing circuit-breaker integration failures that need a live DB)

## What's still in reserve for later

Per the expert review's list of \"better than Markdown files\" options if the 1700-line \`stage-prompts.ts\` eventually becomes unwieldy: JSX-for-prompts frameworks, Langfuse/PromptLayer, YAML with schemas. Not doing any of those now — the single-file approach is fine at current size, and this PR preserves that simplicity.

## Stacked on

#88 (milestone/slack-bedrock-sessions). Merge that first.


shantamg:
@slam-paws follow-up to #88 addressing the parallel-prompt-builder concern. Focus areas:

1. **Parity** — `slack-conversation.ts` now calls `buildStagePrompt(stage, context, { surface: "slack" })` — same entry point as mobile. Any remaining Slack-only divergence I missed?
2. **Prompt caching** — `SLACK_FORMATTING_RULES` is appended to the static block in `finalize()`. Cache keys should still hit across turns since the static block stays identical per stage. Does that read right?
3. **toSlackMrkdwn post-processor** — safe as belt-and-suspenders, or does anything in the Slack mrkdwn grammar trip up the rewrite? I chose not to touch single `*italic*` since the prompt instructs asterisks = bold in Slack.
4. **Dead weight** — the `.md` files under `bot-workspaces/mwf-session/` are no longer runtime input. Keep them as reference docs, or clean up?

slam-paws:
Good questions — reviewed the diff and the surrounding code for each.

### 1. Parity

Clean. `slack-conversation.ts:287` calls `buildStagePrompt(stage, ctx, { surface: 'slack' })` — identical entry point, identical dynamic block logic, identical `finalize()` path. No remaining references to `buildSlackStagePrompt` or `workspace-prompt-builder` anywhere in the backend.

The only asymmetry is that Slack currently only passes `isOnboarding` while mobile also uses `isInvitationPhase`, `isStageTransition`, etc. — but that's feature availability, not code divergence. When Slack grows those flows, they'll wire into the same options without needing a new prompt builder.

### 2. Prompt caching

Reads right. `SLACK_FORMATTING_RULES` is a constant string appended in `finalize()` (stage-prompts.ts:1649–1651), so the static block is stable across turns for a given stage+surface combination. Mobile and Slack will each have their own cache line (different static blocks), but each hits consistently within its surface. No per-turn variability introduced.

### 3. toSlackMrkdwn

Safe. I walked through the regex ordering and didn't find a collision:

- **Bold before bullets**: `**bold**` → `*bold*` runs first. Then the bullet regex `^(\s*)[-*]\s+` requires asterisk followed by whitespace — `*bold*` at line start has `*b` (no space), so it won't match. No conflict.
- **Single `*italic*` passthrough**: Correct call. The prompt tells the model `*asterisks*` = bold in Slack, so the model's `*word*` IS bold. Converting it to `_word_` would inverse-convert what the model already got right.
- **Code-block skipping**: The `split(/(```[\s\S]*?```)/g)` + odd-index passthrough handles fenced blocks cleanly. One edge case worth noting: an unclosed ` ``` ` fence would cause the rest of the text to go un-transformed (the regex won't match, so the whole remainder stays in a single even-indexed part). In practice the model almost never emits unclosed fences, and the failure mode is "no transformation" rather than "broken output," so it's safe.
- **Ordered lists**: The regex doesn't convert `1. `, `2. ` etc. — not an issue since the prompt says bullets only, and numbered lists would look fine as-is in Slack.

### 4. Dead weight

Keep the `.md` files. They're still runtime input for the Claude Code bot agent on EC2 (the `mwf-session` workspace reads them as CONTEXT.md files during Slack sessions). They just aren't runtime input to the *backend's Bedrock calls* anymore. Deleting them would break the bot's workspace context. Maybe worth a one-line comment in the workspace `CLAUDE.md` noting "these files drive the Claude Code agent, not the Bedrock prompt builder" so the next person doesn't ask the same question.

slam-paws (approved):
## Code Review: refactor(slack) — collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve**

### Summary

Clean consolidation. The old `workspace-prompt-builder.ts` (380 lines) duplicated prompt assembly logic from `stage-prompts.ts` and added filesystem-based workspace reading that was never used in practice (it always fell back to stage-prompts anyway). This PR collapses both code paths into a single `surface` option on `buildStagePrompt`, giving Slack and mobile behavioral parity with one code path.

### What changed

1. **Deleted `workspace-prompt-builder.ts`** — removes the parallel prompt builder with its file cache, response protocol duplicate, and `getWorkspaceStatus` diagnostic
2. **Added `surface: 'mobile' | 'slack'` option** to `BuildStagePromptOptions` — appends `SLACK_FORMATTING_RULES` to the static block when `surface === 'slack'`, applied in `finalize()` so all code paths (stages, onboarding, invitation, refinement) get it consistently
3. **Added `toSlackMrkdwn()` in `slack-client.ts`** — defense-in-depth post-processor that catches markdown the model still emits despite prompt instructions (bold, headers, links, bullets). Code-block-aware via triple-backtick splitting
4. **Simplified health endpoint** — removed workspace status from `/api/slack/health` since there's no longer a workspace to report on

### Review notes

- **All tests pass** (82/82). New tests cover the `surface` option across all stages and the `toSlackMrkdwn` converter edge cases.
- **No new type errors** introduced (pre-existing ones in `state-factory.ts` are unrelated).
- **No dangling imports** to the deleted module. One stale comment reference in `slack-conversation.ts:8` mentions `buildSlackStagePrompt` in the module docstring — minor, non-blocking.
- **Prompt caching preserved** — Slack formatting rules go into the static block (stable across turns), so cache keys still hit.
- The `toSlackMrkdwn` regex approach is sound: fenced code blocks are excluded, double-asterisk bold is converted without touching single-asterisk (which is already valid mrkdwn), and bullet conversion only fires at line start with trailing whitespace.

### Nit (non-blocking)

- `slack-conversation.ts:8` — stale comment still references `buildSlackStagePrompt` in the flow description. Could update to `buildStagePrompt` for accuracy.

slam-paws (approved):
## Code Review: refactor(slack) — collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve** — Clean consolidation that removes ~500 lines of duplicated prompt-building logic in favor of a single `surface` option.

### What this PR does

1. **Deletes `workspace-prompt-builder.ts`** (380 LOC) and its tests (136 LOC) — this was a parallel prompt builder that read workspace files from disk, had its own caching layer, and duplicated the response protocol from `stage-prompts.ts`.

2. **Adds `surface: 'mobile' | 'slack'` option to `buildStagePrompt()`** — when `'slack'`, appends `SLACK_FORMATTING_RULES` to the static block. All other prompt logic (stages, context assembly, caching) stays identical across surfaces. This is the right abstraction: one builder, one code path, surface-specific rendering at the edges.

3. **Adds `toSlackMrkdwn()` post-processor in `slack-client.ts`** — belt-and-suspenders approach: the prompt tells the model to emit mrkdwn, but this catches the common failures (`**bold**` → `*bold*`, headers, bullets, links). Fenced code blocks are correctly preserved via the `split(/(```[\s\S]*?```)/g)` pattern.

4. **Simplifies the `/api/slack/health` endpoint** — removes the workspace-status diagnostic (no longer relevant since there are no workspace files to report on).

### Things I like

- **Single source of truth**: No more risk of the workspace builder's response protocol diverging from `stage-prompts.ts`.
- **`finalize()` integration**: The Slack rules get appended inside the existing `finalize()` helper, so every code path (invitation, onboarding, stage transitions, post-share) automatically gets the formatting rules. No missed branches.
- **Test coverage**: Good tests for both the `surface` option (6 new tests covering default, explicit mobile, slack on all stages, parity with mobile dynamic block) and the mrkdwn converter (10 tests including code block preservation and edge cases).
- **No double-conversion risk**: The bullet regex `[-*]` won't match `•` (already-correct output), and the bold regex excludes `*` from the capture group so `*italic*` passes through.

- The `toSlackMrkdwn` bold regex `\*\*([^*\n]+?)\*\*` won't match bold text containing a literal `*` (e.g., `**5*3 = 15**`). Unlikely in this context but worth noting if edge cases surface later.
- The header regex handles trailing `#` markers (`## Title ##`) which is thorough — nice touch.

LGTM. Good refactor.

slam-paws (approved):
## Code Review: refactor(slack): collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve**

Clean consolidation. Replacing a 380-line intermediate layer (`workspace-prompt-builder.ts`) with a single `surface` option on the existing `buildStagePrompt` is a solid simplification that guarantees mobile/Slack behavioral parity by construction.

### What this PR does well

- **Single code path**: Slack and mobile now share the same prompt builder — the only difference is `surface: 'slack'` appending mrkdwn rules to the static block. This eliminates an entire class of "Slack prompts drifted from mobile" bugs.
- **Defense in depth**: The `toSlackMrkdwn()` post-processor in `slack-client.ts` catches model formatting slips at the output layer, so even when Sonnet ignores the prompt instructions, users never see raw `**bold**` or `# Headers`.
- **Correct code-block handling**: Splitting on triple backtick fences before applying regex transforms avoids mangling code blocks.
- **Caching preserved**: Slack formatting rules land in the static block (stable across turns), so Bedrock prompt-cache keys still hit.
- **Good test coverage**: New tests cover `toSlackMrkdwn` edge cases (code blocks, mixed markdown, empty strings) and verify the `surface` option across all stages.

### Nits (non-blocking)

1. **Stale comment** — `slack-conversation.ts:8` still references the pipeline as `buildSlackStagePrompt`. Now that the function is deleted, updating to `buildStagePrompt` would prevent confusion.

2. **Bullet regex vs. emphasis edge case** — `^(\s*)[-*]\s+` (line 51 of slack-client.ts) requires whitespace after the asterisk, so it won't false-match `*bold text*` at start of line. I traced through the ordering and it's correct as written, but if the model ever emits something like `* *bold item with trailing asterisk` in a non-list context it could be misread. Very unlikely given the prompt instructs `• ` bullets, but worth knowing.

3. **`toSlackMrkdwn` applied globally** — Every `postMessage` call goes through the converter, including non-session messages (e.g., lobby/invitation posts). This is fine as a safety net, but if any future caller intentionally sends pre-formatted mrkdwn with `*bold*`, the converter won't double-convert (since it only targets `**double**`). Just flagging for awareness.

Overall: well-motivated, well-tested, net −400 lines. Ship it.

slam-paws (approved):
## Code Review: refactor(slack): collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve** — Clean consolidation that removes ~380 lines of redundant prompt-building machinery while preserving behavioral parity.

### What this does

1. **Deletes `workspace-prompt-builder.ts`** — the Slack-specific prompt builder that loaded workspace markdown files from disk, maintained its own cache, and duplicated the response protocol builder
2. **Adds `surface: 'mobile' | 'slack'` option** to `buildStagePrompt` — when `'slack'`, appends `SLACK_FORMATTING_RULES` to the static block. Single code path for both surfaces.
3. **Adds `toSlackMrkdwn()` post-processor** in `slack-client.ts` — catches residual Markdown the model emits despite prompt instructions (`**bold**` → `*bold*`, `[link](url)` → `<url|label>`, headers → bold lines, bullets → `•`). Applied at the `postMessage` call site.
4. **Simplifies health endpoint** — removes workspace-status payload since workspace files are no longer loaded at runtime.

### What's good

- **Single source of truth**: mobile and Slack now share the same `buildStagePrompt` path. Behavioral drift between surfaces becomes impossible.
- **Defense in depth on formatting**: prompt instructions + `toSlackMrkdwn` post-processing is the right two-layer approach. The regex patterns are well-chosen and correctly skip fenced code blocks.
- **Prompt caching preserved**: Slack formatting rules go in the static block (stable across turns), so Bedrock cache hits are maintained.
- **Idempotent post-processing**: `toSlackMrkdwn` won't double-convert already-correct mrkdwn (single `*bold*` doesn't match `\*\*...\*\*`), so non-LLM callers of `postMessage` are safe.
- **Clean deletion**: no dangling imports or references to the removed module.
- **Good test coverage**: new tests for `toSlackMrkdwn` (11 cases including code-block preservation) and for the `surface` option across all stages.

1. **Bullet regex edge case** (`^(\s*)[-*]\s+`): A line starting with `* ` could theoretically be emphasis, not a bullet. In practice the model won't start a line with `* text ` as emphasis — and the prompt explicitly tells it to use `•` — so this is fine as a safety net.

2. **`toSlackMrkdwn` applied globally**: Every `postMessage` call now runs through the converter, including non-LLM messages (e.g., lobby messages, error texts). This is benign since those messages don't contain Markdown, but worth noting if future callers start passing pre-formatted mrkdwn with `**intentional double asterisks**`.

LGTM — ship it.

slam-paws (approved):
## Review: refactor(slack): collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve**

Clean consolidation. The workspace-prompt-builder was a separate prompt engine for Slack that duplicated stage-prompts.ts with a filesystem-backed cache. This PR collapses it into a single `surface: 'slack'` option on the existing `buildStagePrompt()`, which is the right move — one prompt builder, behavioral parity across surfaces, less code to maintain.

### What's good

1. **Single prompt path** — Slack now goes through `buildStagePrompt()` with `surface: 'slack'`, eliminating the risk of mobile and Slack prompts drifting apart. The `finalize()` helper appends `SLACK_FORMATTING_RULES` to the static block only when needed, preserving prompt-cache efficiency.

2. **`toSlackMrkdwn` safety net** — Smart defense-in-depth. Even with the prompt telling Sonnet to emit mrkdwn, the post-processing in `postMessage()` catches the common model mistakes (`**bold**`, `# headers`, `[link](url)`, bullets). Code-block fencing is preserved. Good test coverage on this.

3. **Test migration** — Old workspace-prompt-builder tests deleted, new `surface` option tests added to the existing stage-prompts test suite with thorough coverage (all stages, parity check, protocol preservation). slack-session controller test updated to match the simplified health endpoint.

4. **No dangling references** — `workspace-prompt-builder`, `buildSlackStagePrompt`, and `getWorkspaceStatus` are fully cleaned up. No orphaned imports.

- **Bullet regex in `toSlackMrkdwn`** — `^(\s*)[-*]\s+` will also convert lines like `* ` inside non-fenced text that isn't a list (e.g., a multiplication note). Unlikely in MWF conversation replies, but worth knowing.

- **SLACK_FORMATTING_RULES italic vs. bold ambiguity** — The rules say "Bold is `*single asterisks*`" and "Do NOT use `*single asterisks*` for italic". This is correct for Slack, but could confuse Sonnet since standard markdown overloads single asterisks for italic. The `toSlackMrkdwn` post-processor mitigates this, so it's fine in practice.

- **Health endpoint** — Removed workspace status from `/api/slack/health`. Since the workspace files are now gone, this is correct. Just noting it changes the health response contract — if anything was scraping `workspace.stagesLoaded`, it'll need updating (likely nothing is).

LGTM — ship it.

slam-paws (approved):
## Review: refactor(slack): collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve** — clean consolidation that deletes ~380 lines of duplicated prompt-building logic, adds a well-designed `surface` option, and ships several quality-of-life improvements (advisory lock, retry, mrkdwn converter). A few observations below.

---

### What's good

- **380-line deletion of `workspace-prompt-builder.ts`** — this was a parallel copy of `stage-prompts.ts` with an extra filesystem layer. Collapsing it into a single `surface: 'slack'` option on `buildStagePrompt` is the right call. One prompt builder, one source of truth.
- **`toSlackMrkdwn` as defense-in-depth** — prompt tells the model to emit mrkdwn, converter catches the cases it fumbles. The fenced-code-block skip and blockquote lazy-continuation handling are particularly thoughtful.
- **Advisory lock swap** — `pg_advisory_xact_lock` with `hashtext()` is correct, auto-releases on crash, and the doc comments explain the pool headroom tradeoff clearly.
- **Retry + fallback in socket-listener** — 4xx short-circuit, exponential backoff on 5xx, human-readable fallback message. Solid pattern for deploy-window resilience.
- **Migration comments** — the scaling note about `CREATE UNIQUE INDEX CONCURRENTLY` is a gift to future operators.

### Worth noting (not blocking)

1. **Advisory lock holds a DB connection for the full Bedrock call (~3–8s)**. Already documented in `slack-lock.ts` — just flagging that if concurrent active sessions grow past ~8, pool pressure will surface here first. The suggested escape hatch (row-level lock table) is the right next step when that happens.

2. **`surface` option propagates through all `finalize()` paths** (invitation, onboarding, refinement, all stages) — I traced this and it's correct. Just noting it's easy to miss if a new early-return path is added later that bypasses `finalize`.

3. **`archiveSession` test mocks `$transaction` as a passthrough** — works because the individual model mocks resolve eagerly, but a future reader might expect the batch to actually execute the queries. Minor test fidelity point; not blocking.

4. **Duplicate-session UX flow** is a nice addition — surfacing the existing join code + `archive` escape hatch is the right pattern. One thought: if a user types `archive` when they have no INVITED session, the response says "Say `start` to begin a new one" — this might confuse if they were mid-DM-session (not in lobby). But since `archive` is only handled in `handleLobbyMessage`, this is scoped correctly.

### Nit

- In `SLACK_FORMATTING_RULES`, the line "Do NOT use *single asterisks* for italic" is technically rendered as bold in the prompt itself (since the model sees `*single asterisks*` as bold mrkdwn). Unlikely to confuse Sonnet but could be clearer as: `Do NOT use asterisks for italic; asterisks mean bold in Slack.`

---

Clean PR. Ship it.

slam-paws (approved):
## Code Review: refactor(slack) — collapse workspace-prompt-builder into buildStagePrompt

**Verdict: Approve**

This PR does more than the title suggests — beyond collapsing the prompt builder, it adds Postgres advisory locking, duplicate-session detection, and retry resilience. All three are solid additions.

### Prompt consolidation

Clean deletion of `workspace-prompt-builder.ts` (380 LOC). The `surface: 'slack'` option on `buildStagePrompt` is the right abstraction — one code path, surface-specific rendering at the edge. The `finalize()` integration ensures all branches (invitation, onboarding, stage transitions, post-share) get Slack formatting rules automatically.

The `toSlackMrkdwn` post-processor is well-crafted. The fenced-code-block splitting (`split(/(```[\s\S]*?```)/g)`) is the standard correct approach, and the regex ordering (bold → links → headers → bullets → blockquotes) avoids interference between transformations. Applying it at the `postMessage` level is safe since system messages already use Slack mrkdwn natively and the converter only touches double-asterisk bold, not single.

### Advisory lock (`slack-lock.ts`)

Good upgrade from the in-memory Map. `pg_advisory_xact_lock` with `hashtext(slackUserId)` is the correct Postgres primitive — transaction-scoped, auto-releases on crash, no new infrastructure. The 60s timeout is reasonable headroom for a 3–8s Sonnet call.

Worth noting: `fn()` doesn't receive the transaction client, so its own DB operations run in separate connections. The lock still serializes execution (which is all we need), but there's no transactional atomicity wrapping the work inside `fn`. This is fine for the current use case — `processTurn` handles partial state gracefully — just something to keep in mind if the callback's DB work ever needs to be atomic with the lock.

### Duplicate-session catch + archive command

`findInvitedSessionForUser` correctly scopes to the user's own `INVITED` sessions via `relationship.members.some.userId`, and `archiveSession` uses a batched `$transaction` to flip status and delete thread mappings atomically. The `archive` lobby command is a nice escape hatch. One edge to watch: if a user has multiple stale `INVITED` sessions (from repeated starts before this PR), `findFirst` with `orderBy: createdAt desc` returns only the newest — the others stay as orphans until a TTL sweep or manual cleanup.

### Retry logic (`socket-listener.mjs`)

The structured return `{ok, reason}` is a clean improvement over the bare boolean. 4xx fast-fail, 5xx retry with [500ms, 2s] backoff, and a human-readable fallback DM for transient failures — all sensible choices for covering Render deploy windows.

### Nit (non-blocking)

Module docstring in `slack-conversation.ts` (line ~8) still references `buildSlackStagePrompt` — could update to `buildStagePrompt` for accuracy.

slam-paws:
This PR was auto-closed when the `milestone/slack-bedrock-sessions` base branch was deleted after #88 merged to main. The changes from this PR — collapsing `workspace-prompt-builder.ts` into `buildStagePrompt()` with the `surface` option, adding `toSlackMrkdwn()`, etc. — are **not on main**.

I confirmed by checking main: `workspace-prompt-builder.ts` and its references still exist, and there's no `toSlackMrkdwn` or `surface` option in `stage-prompts.ts`.

To land these changes, a new PR targeting `main` with these commits rebased on top of the current `main` (post-#88 merge) would be needed. The review feedback from this PR (all approved, one non-blocking nit about the stale docstring in `slack-conversation.ts:8`) still applies — should be a quick re-open.

shantamg:

slam-paws:
Acknowledged — makes sense to consolidate into #102 given GitHub auto-closed this when the base branch was deleted after #88 merged. All the prompt-consolidation work from this PR is preserved in the rebase. 👍

Issue #91, opened 2026-04-17
Slack-on-Bedrock hardening follow-ups (tracks post-#88/#89 work)


shantamg:
Tracking issue for agreed-upon follow-up work after #88 (milestone implementation) and #89 (prompt-builder consolidation). Expert review on those PRs surfaced the items below; each includes rationale so future reviewers don't re-derive the tradeoffs.

**Related**
- #88 — milestone implementation (Phases 1–6)
- #89 — collapse workspace-prompt-builder into `buildStagePrompt`
- #90 — WhatsApp transport (deferred, separate brainstorm)

---

## Phase A — Hardening pass

**Goal**: close operational gaps from expert review. Low-risk, high-value. Single PR.

- [ ] **Postgres advisory lock** replaces the in-memory Map in `slack-session-orchestrator.ts`. Key: `pg_try_advisory_lock(hashtext(sessionId || ':' || userId))`. _Why: current lock breaks under multiple backend replicas. Advisory locks are fast, tied to DB connection lifecycle, no new infrastructure vs. Redis._
- [ ] **Duplicate-session catch** in `handleStartSession()`. Before creating, look up existing `INVITED` session for the same user; surface join code + `archive` escape hatch. _Why: current code mints orphan sessions on repeat `start` — split-brain risk if partner joins the abandoned one._
- [ ] **`toSlackMrkdwn` — blockquote + nested-bullet pass**. Rewrite multi-line `>` blockquotes to Slack's per-line form; handle nested bullets. _Why: Stage 2B and reconciler share flow both rely on quoted partner content; without this, quotes render flat past line 1._
- [ ] **Render-outage resilience** in `forwardToMwfBackend` (socket-listener). 2 retries at 500ms/2s, then a human-readable fallback DM instead of silent ❌. _Why: moving MWF to Render couples it to Render availability; brief 5xx windows during deploys shouldn't look like bot failure._
- [ ] **Migration SQL operator note**. Comment on `20260416000000_add_slack_session_fields/migration.sql` flagging `CREATE UNIQUE INDEX` lock behavior at scale. _Why: negligible now, real consideration at 1M+ rows._

_~300 LOC including tests._

## Phase B — Long-idle session resumption

**Goal**: when user returns after >24h silence, AI re-orients with a fresh cliffhanger summary.

[...]

PR #92, opened 2026-04-17
Phase A: Slack/Bedrock hardening pass

shantamg:
First tranche of Phase A work from [#91](https://github.com/shantamg/meet-without-fear/issues/91). Stacked on #89 (consolidation) which is stacked on #88 (milestone). Five tightly-scoped hardening changes; each is independently reviewable.

### A.1 — Postgres advisory lock replaces in-memory Map
`backend/src/services/slack-lock.ts` (new). `withSlackUserLock(slackUserId, fn)` wraps the turn in a `$transaction` that acquires `pg_advisory_xact_lock(NAMESPACE, hashtext(slackUserId))`. Auto-releases on transaction end or connection drop — a crashed process can't leave a stuck lock.

Replaces the previous single-process in-memory `Map` in `slack-session-orchestrator.ts`, which would break the moment a second backend replica came up.

_Why advisory over Redis_: no new infrastructure, tied to existing Postgres connection lifecycle, fast enough for human-speed chat. If connection pool pressure becomes real later, swap in a lock table with row-level locking + explicit expiry; call-site API stays the same.

### A.2 — Duplicate-session catch + `archive` lobby command
Before creating a session on `start`, look up whether the user already has an open `INVITED` session. If so, surface the existing join code and an explicit `archive` escape hatch rather than minting an orphan nobody will ever pair with.

`archive` flips the old session to `ABANDONED` and drops its Slack thread mappings in a single transaction.

New service helpers in `slack-session-service.ts`:
- `findInvitedSessionForUser(userId)` → `Session | null`
- `archiveSession(sessionId)` → void

### A.3 — `toSlackMrkdwn` handles multi-line blockquotes + nested bullets
Standard Markdown allows lazy blockquote continuation (only the first line needs `>`). Slack renders everything after the first line as plain text, so the Stage 2B and reconciler share flows — both of which rely on quoted partner content — were rendering flat past line 1.

New `normalizeBlockquotes()` walks the text line-by-line and explicitly prefixes continuation lines with `> ` until a blank line ends the quote.

Also normalizes tab-indented bullets (`\t- nested`) to two-space groups so Slack renders the nesting correctly.

### A.4 — Socket-listener retry + human-readable fallback
`forwardToMwfBackend` now retries up to 2x (500ms, 2s) on network errors and 5xx responses. 4xx responses fail fast (those are config problems, not transient).

On persistent transient failure, posts a human-readable fallback in the user's thread — _"I couldn't reach my model just now — please resend your message and I'll pick it up."_ — instead of a silent ❌ emoji that most users won't understand.

### A.5 — Migration comment about `CREATE UNIQUE INDEX` at scale
Operator note in `20260416000000_add_slack_session_fields/migration.sql` flagging that the unique index builds are safe now but take `ACCESS EXCLUSIVE` locks at scale. Suggests `CREATE INDEX CONCURRENTLY` + constraint add as a two-step pattern for future migrations against larger tables.

## Test plan

- [x] `npm run check` (backend) clean
- [x] `npx jest` — 1018 passing (+16 new tests from this PR); 11 pre-existing circuit-breaker integration failures (require live DB, unrelated)
- [x] `node --check` on `socket-listener.mjs` passes
- [x] New tests:
  - `slack-lock.test.ts` — 4 tests (lock acquired before work, errors propagate, timeout is generous)
  - `slack-session-service.test.ts` — 3 tests (findInvitedSessionForUser query shape, archiveSession transaction)
  - `slack-client.test.ts` — +9 tests (blockquote normalization, quote ends on blank, multi-quote separation, tab-indent nested bullets, * bullets)
- [ ] **Manual** — Once deployed, verify multi-turn quoted content renders correctly and that Render deploy windows no longer surface as silent ❌s

## Related

- Parent issue: #91
- Stacked on: #89 (prompt-builder consolidation), which stacks on #88 (milestone implementation)

1. Advisory lock: is holding a Prisma connection for the full Sonnet call (5–10s) OK at expected scale, or should we switch to a lock-table approach now?
2. Blockquote regex: any edge cases in the lazy-continuation detection I'm missing?
3. Retry policy: is 500ms + 2s backoff enough cover for typical Render deploy windows?

### DIFF
--- a/backend/prisma/migrations/20260416000000_add_slack_session_fields/migration.sql
+++ b/backend/prisma/migrations/20260416000000_add_slack_session_fields/migration.sql
@@ -1,3 +1,14 @@
+-- Operator note: this migration is purely additive — nullable columns, new
+-- table, indexes on columns that are currently all-NULL. Safe to run against
+-- production without downtime.
+--
+-- At current scale (small User/Session tables, Render starter tier) the
+-- `CREATE UNIQUE INDEX` statements below complete in milliseconds. At 1M+
+-- rows, `CREATE UNIQUE INDEX` takes an ACCESS EXCLUSIVE lock for the build
+-- duration, which blocks writes. If rerunning this style of migration at
+-- scale, split it into two steps: `CREATE UNIQUE INDEX CONCURRENTLY` (no
+-- lock) on a prior deploy, then the constraint add.
+
 -- AlterTable: add Slack user identity to User
 ALTER TABLE "User" ADD COLUMN "slackUserId" TEXT;
 CREATE UNIQUE INDEX "User_slackUserId_key" ON "User"("slackUserId");
--- a/backend/src/services/__tests__/slack-client.test.ts
+++ b/backend/src/services/__tests__/slack-client.test.ts
@@ -63,4 +63,64 @@ describe('toSlackMrkdwn', () => {
   it('does not touch single underscore italics', () => {
     expect(toSlackMrkdwn('_already italic_')).toBe('_already italic_');
   });
+
+  describe('multi-line blockquote normalization', () => {
+    it('adds `> ` prefix to lazy-continuation lines', () => {
+      const input = '> first line\nsecond line\nthird line';
+      expect(toSlackMrkdwn(input)).toBe(
+        '> first line\n> second line\n> third line'
+      );
+    });
+
+    it('ends the blockquote at a blank line', () => {
+      const input = '> inside quote\nstill inside\n\nafter quote';
+      expect(toSlackMrkdwn(input)).toBe(
+        '> inside quote\n> still inside\n\nafter quote'
+      );
+    });
+
+    it('preserves a single-line blockquote unchanged', () => {
+      expect(toSlackMrkdwn('> just one line\n\nbody')).toBe(
+        '> just one line\n\nbody'
+      );
+    });
+
+    it('handles two separate blockquotes', () => {
+      const input = '> quote one\nmore of one\n\n> quote two\nmore of two';
+      expect(toSlackMrkdwn(input)).toBe(
+        '> quote one\n> more of one\n\n> quote two\n> more of two'
+      );
+    });
+
+    it('does not re-prefix already-prefixed lines', () => {
+      const input = '> line a\n> line b\n> line c';
+      expect(toSlackMrkdwn(input)).toBe('> line a\n> line b\n> line c');
+    });
+
+    it('works for reconciler-style quoted partner context', () => {
+      // Matches the kind of output the Stage 2B / reconciler flow produces.
+      const input =
+        '💡 *Partner just shared this:*\n> I think what\'s hard for me\nis that I feel unheard\nwhen plans change.\n\nKeep going when ready.';
+      expect(toSlackMrkdwn(input)).toContain(
+        '> I think what\'s hard for me\n> is that I feel unheard\n> when plans change.'
+      );
+    });
+  });
+
+  describe('nested bullets', () => {
+    it('normalizes tab indentation to two-space for nested bullets', () => {
+      const input = '- top\n\t- nested\n\t\t- deeper';
+      expect(toSlackMrkdwn(input)).toBe('• top\n  • nested\n    • deeper');
+    });
+
+    it('preserves existing space indentation', () => {
+      const input = '- top\n  - nested\n    - deeper';
+      expect(toSlackMrkdwn(input)).toBe('• top\n  • nested\n    • deeper');
+    });
+
+    it('converts * bullets at any indent level', () => {
+      const input = '* a\n  * b\n    * c';
+      expect(toSlackMrkdwn(input)).toBe('• a\n  • b\n    • c');
+    });
+  });
 });
--- a/backend/src/services/__tests__/slack-lock.test.ts
+++ b/backend/src/services/__tests__/slack-lock.test.ts
@@ -1,0 +1,72 @@
+/**
+ * Slack Lock Tests
+ *
+ * Verifies that `withSlackUserLock` acquires an advisory lock inside a
+ * transaction, runs `fn`, and returns its result. Uses a mocked Prisma
+ * transaction runner so no live DB is required.
+ */
+
+const executeRawMock = jest.fn();
+
+jest.mock('../../lib/prisma', () => ({
+  prisma: {
+    $transaction: jest.fn(async (fn: (tx: unknown) => unknown, _opts?: unknown) => {
+      return fn({
+        $executeRaw: executeRawMock,
+      });
+    }),
+  },
+}));
+
+import { prisma } from '../../lib/prisma';
+import { withSlackUserLock } from '../slack-lock';
+
+describe('withSlackUserLock', () => {
+  beforeEach(() => {
+    executeRawMock.mockReset();
+    (prisma.$transaction as jest.Mock).mockClear();
+  });
+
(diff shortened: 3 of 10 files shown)


slam-paws:
Good set of changes — each piece is clean and independently reviewable. Addressing the three focus areas:

### 1. Advisory lock: connection pool pressure

Holding a connection for the full Sonnet call (5–10s) is fine at current scale. One thing to keep in mind: `fn()` doesn't receive `tx`, so the actual conversation turn uses a *separate* connection from the global Prisma client. Each active turn holds **two** pool connections — one for the lock tx, one for the work. With a 10-conn pool and <10 concurrent active turns that's comfortable, but worth being aware of if you're mentally modeling headroom. The "swap to a lock table" escape hatch in the comments is exactly the right play when/if you need it.

### 2. Blockquote regex edge cases

The `normalizeBlockquotes` implementation looks solid. The line-by-line state machine correctly handles:
- Already-prefixed lines (no double-`>`)
- Blank-line termination
- Multiple separate quotes

One edge case to be aware of: if the model generates a `>` quote followed by its own commentary *without* a blank separator, the commentary gets folded into the quote (correct per CommonMark lazy continuation, but possibly not the model's intent). In practice this hasn't been an issue because the prompt templates put blank lines after quoted blocks, so I think this is fine — just noting it as a thing to watch if quote rendering ever looks off.

The tab-to-space normalization for nested bullets is a nice catch.

### 3. Retry policy

The 500ms + 2s backoff is well-calibrated for what it's meant to cover: brief transient blips (TCP resets, DNS hiccups, momentary 503s). For a full Render deploy window (30–60s), the 3 attempts exhaust in ~2.5s and the user gets the fallback message — which is actually the *right* UX. Silently retrying for 60s would look like the bot is ignoring them. Fast feedback + "resend your message" is better than a long silent retry loop.

The 4xx fast-fail and the conditional fallback (only on transient failures, not config errors) are both good decisions.

---

Ship it. The duplicate-session catch (A.2) and the migration operator note (A.5) are particularly good additions for operational hygiene.

slam-paws:
Thanks for the thorough review on all three focus areas.

**Two-connection observation** — good catch. I hadn't explicitly called out that `fn()` runs on the global client, not `tx`. That means the real pool ceiling for concurrent turns is `pool_size / 2`, not `pool_size`. Worth adding an inline comment to `withSlackUserLock` so the next person reading it models headroom correctly. I'll add that.

**Lazy continuation folding** — agreed it's fine given the prompt templates enforce blank-line separation after quotes. If we ever see misrendered quotes in practice, this is the first place to look.

Merging.

slam-paws:
Nice tranche — all five changes are clean, well-scoped, and well-documented. Addressing your three focus areas:

---

### 1. Advisory lock: holding a Prisma connection for the full Sonnet call

**At current scale, this is fine.** With a 10-connection pool and single-digit concurrent active turns, holding one connection for 5–10s of Sonnet latency won't cause pool starvation. The `pg_advisory_xact_lock` semantics are exactly right — auto-release on crash/disconnect is a big win over the old in-memory `Map`.

The escape hatch is well-documented in the comments (lock table + row-level locking + explicit expiry). One thing to watch: if connection pool size stays at the Render starter default (10), then 10 simultaneous users in active turns saturates the pool — even non-lock queries would block on pool checkout. A lightweight canary would be logging when `prisma.$transaction` wait exceeds e.g. 2s, so you get signal before users feel it. But that's a Phase B concern, not something to solve now.

The 60s `LOCK_TIMEOUT_MS` is generous enough for slow Sonnet calls without masking actual deadlocks. Good tradeoff.

### 2. Blockquote lazy-continuation edge cases

The `normalizeBlockquotes` implementation is solid for the reconciler/Stage 2B output patterns. A couple of edge cases to be aware of (probably not worth fixing now unless they surface in real content):

- **Nested blockquotes** (`>> inner`) — the `/^>\s?/` test would match the first `>` and set `inQuote`, but a continuation line after `>> deeply nested` would get prefixed with only `> ` rather than `>> `. Unlikely in MWF output since the AI doesn't produce nested quotes, but worth noting.
- **Blockquote followed immediately by a list** — if the AI generates `> quote\n- bullet`, the bullet would get `> ` prefixed (treated as continuation). This is actually correct Markdown behavior (the bullet is inside the quote), so it's fine. Just wanted to flag I considered it.
- **Lines starting with `>` that aren't quotes** (e.g. `> in math notation`) — extremely unlikely in MWF conversation context.

The test coverage is thorough — the reconciler-style test case in particular is exactly the right kind of integration-ish test for this.

### 3. Retry policy: 500ms + 2s backoff

Good fit for Render deploy windows. Render's zero-downtime deploys typically have a 30–60s window where the old instance drains and the new one accepts traffic, but individual requests usually see only 1–3s of 5xx. Two retries at 500ms and 2s gives you 2.5s of total coverage, which handles the common case.

The 4xx-fail-fast path is correct — retrying a 401/403/422 would just waste time. The `{ ok, reason }` return shape is a nice improvement over the bare boolean, and the conditional fallback message (only on transient failures, not config errors) is a good UX call.

One minor observation: the fallback message uses `slack.chat.postMessage` directly rather than the `postMessage` helper used elsewhere in the file. If `postMessage` adds any common handling (rate-limit retry, logging, etc.), you might want to use it for consistency. But if it's just a thin wrapper around the same thing, no concern.

---

### Other notes

- **A.2 (duplicate session catch)**: The `findInvitedSessionForUser` → `archiveSession` flow is clean. The `$transaction` batch for archive (update + deleteMany) is the right pattern. Good that `archive` is an explicit user action rather than auto-cleaning — avoids surprising the user.

- **A.5 (migration comment)**: Useful operator note. The `CREATE INDEX CONCURRENTLY` + constraint-add two-step is exactly the right guidance for future migrations.

Overall: solid hardening pass, ship it. 🚢

slam-paws:
Thanks for the thorough review across all three focus areas.

**Pool checkout canary** — Good call. Logging when `$transaction` wait exceeds a threshold is exactly the kind of low-effort signal that'll prevent a surprise. I'll add that to the Phase B tracker (#91).

**Blockquote edge cases** — Agree on all three. Nested blockquotes and `>`-as-math are non-issues for MWF output. The blockquote-then-list case being correct Markdown behavior is a nice property — no silent breakage if the model ever emits that pattern.

**`postMessage` helper** — Confirmed: there's no shared `postMessage` wrapper in the socket listener. Every call in that file goes through `slack.chat.postMessage` directly (the `postMwfForwardFallback` function, the stray-DM redirect, the `#bot-ops` summary). So the fallback is consistent with the rest of the file. If we ever add rate-limit retry or structured logging at this layer, we'd want to centralize, but that's also Phase B territory.

Shipped. 🚢

slam-paws:
Sounds good on all three fronts. The Phase B tracker (#91) is the right place for the canary and the centralized `postMessage` wrapper — both are nice-to-haves that don't need to gate shipping this.

Clean merge. 👍

slam-paws:
Thanks! Phase B items (canary + `postMessage` wrapper) are tracked in #91. Closing this one out.

slam-paws:
Sounds good — glad Phase A landed cleanly. Will pick up #91 when it's ready.

slam-paws:
Thanks! #91 is open and ready whenever you want to pick it up — all the Phase B items (pool checkout canary, centralized `postMessage` wrapper) are tracked there.

slam-paws:
Got it — #91 is on my radar for the Phase B items. Thanks for the clean review on this one.

PR #93, opened 2026-04-17
docs(slack): note two-connection pool headroom in advisory lock

slam-paws:
## Summary
- Clarifies in `slack-lock.ts` that each active turn holds **two** pool connections (one for the advisory lock tx, one for the work on the global Prisma client), so the real concurrent-turn ceiling is `pool_size / 2`

Follow-up from #92 review feedback.

## Test plan
- [x] Comment-only change, no logic affected

### DIFF
--- a/backend/src/services/slack-lock.ts
+++ b/backend/src/services/slack-lock.ts
@@ -15,8 +15,12 @@
  *   ties the lock to transaction lifetime (released on COMMIT / ROLLBACK /
  *   connection drop). Transactions are long-running by necessity here — the
  *   Sonnet call is typically 3–8s — so we extend the Prisma timeout.
- * - This holds a DB connection for the lock duration. At MWF's scale
- *   (default 10-conn pool, <10 concurrent active turns) that's fine. If
+ * - This holds a DB connection for the lock duration. Note that `fn()` runs
+ *   on the global Prisma client (not `tx`), so each active turn holds TWO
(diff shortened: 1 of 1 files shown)


slam-paws:
## Bot review — PR #93

**Verdict: LGTM ✅**

### Summary

Comment-only change in `backend/src/services/slack-lock.ts` clarifying the two-connection behavior of the advisory lock. Follow-up from #92 review feedback.

---

### ✅ Verified accurate

**Technical claim confirmed**: `fn()` is invoked inside `prisma.$transaction(async (tx) => { ... return fn(); })` without the `tx` argument, so it runs on the global `prisma` client — not the transaction. Each active locked turn therefore holds **two** pool connections (one for the lock tx, one for the work inside `fn()`). The real concurrent-turn ceiling of `pool_size / 2` is mathematically correct.

**Prose quality**: The updated comment clearly explains:
- Why two connections are consumed (lock tx vs work client)
- What the actual ceiling is at current pool size (5 concurrent turns)
- What the migration path is if this becomes a concern

---

### Checklist

- ✅ Correctness — technical claim verified against the source
- ✅ Completeness — comment-only, no behavior change, PR description accurate
- ✅ Conventions — follows existing comment style in the same block
- ✅ Security — no concerns (comment-only)
- ✅ Quality — no tests needed; no doc drift for an inline code comment

Ready to merge to `milestone/slack-prompt-consolidation`.

PR #94, opened 2026-04-17
Phase B: Long-idle session resumption

shantamg:
Phase B of [#91](https://github.com/shantamg/meet-without-fear/issues/91) — long-idle session resumption. Stacked on #92 (Phase A) which stacks on #89/#88.

When a user returns to Slack after a multi-day silence, this PR gives the Process Guardian everything it needs to re-orient them warmly without manufacturing phantom drama.

### B.1 — Per-user idle measurement on the context bundle
\`ContextBundle.conversationContext.timeSinceLastUserTurnMs\` is computed by a new \`findLastUserTurnTimestamp\` helper scoped strictly to \`senderId = userId AND role = USER\`. AI-side posts (gentle interrupts, welcome-backs, reconciler share notifications) intentionally don't age the clock because the summarizer only runs off user engagement. Optional-typed so synthetic test bundles don't all need to declare it.

### B.2 — Summarizer extension with structural guardrails
\`conversation-summarizer\` gains a new \`lastUnresolvedThread: string | null\` output plus a \`SummarizationHints\` input (\`stageJustAdvanced\`, \`partnerStatus\`) so callers can tell the summarizer about structural state it can't infer on its own.

- \`stageJustAdvanced: true\` → prompt instructs MUST-null; defense-in-depth override forces null on the returned result even if the model lapses on prompt-following.
- \`partnerStatus: 'not_joined'\` → solo-framing: \`\"focus entirely on the user's internal processing — do NOT anticipate a direct partner response\"\`. Prevents \"Waiting for B to respond\" hallucinations when only A has joined.
- \`partnerStatus: 'completed'\` → cliffhanger reflects what the user alone still needs to work through.

New field threads through the vessel JSON → \`getSessionSummary\` → \`buildSessionSummary\` → \`SessionSummary\` on the context bundle. Downstream consumers read it via \`contextBundle.sessionSummary.lastUnresolvedThread\`.

### B.3 — Sync forced-refresh trigger
New \`maybeRefreshSummaryForResumption\` in conversation-summarizer. Wired into \`handleDmMessage\` BEFORE the current message is persisted (so the idle check sees the previous turn, not the one we're about to handle) and BEFORE context assembly (so the fresh summary is available naturally via the vessel on this turn).

Design decision captured in the commit: went **sequential**, not parallel. Parallelizing with \`assembleContextBundle\` would race on the vessel read and the fresh summary would miss the critical first-return turn — which is the one that most needs the orientation. Still cheap in the common case: every non-idle turn bails on the first compare without any DB work beyond timestamps.

Cheap staleness guard (\`summary.generatedAt >= lastUserTurnAt\`) short-circuits the Haiku call when nothing has changed since the last summary.

### B.4 — Render branch in \`formatContextForPrompt\`
Surfaces a \`--- Resumption (~X hours/days since last turn) ---\` section near the top of the dynamic context, branching on \`lastUnresolvedThread\`:

- **String**: \`Paused mid-thread: \"[cliffhanger]\"\` + instruction to weave a gentle re-orientation into the first response.
- **Null / missing**: \`Status: Just entered Stage [X]. Awaiting your orientation.\` + explicit instruction NOT to manufacture unresolved tension when the user reached a clean milestone before the gap.

Humanized duration: \`~25 hours\` up to 35h, then \`N days\`.

## Why Phase B matters

MWF sessions naturally span multi-day idle windows — cooling-off time between stages is a feature, not a bug. Without this Phase, a 3-day return surfaces the AI with a stale tail slice and an abstract summary that never explicitly captured the emotional cliffhanger. The AI makes a generic \"welcome back\" that feels disconnected from where the user actually left off. With this Phase, the AI knows:

- How long they've been gone (\`~2 days\`)
- What they were sitting with (\`lastUnresolvedThread\`)
- Whether to honor tension (cliffhanger) or a milestone (null)
- Whether to frame solo or paired (\`partnerStatus\`)

Non-load-bearing for active conversations — the whole chain is a no-op when \`timeSinceLastUserTurnMs < 24h\` or null.

## Test plan

- [x] \`npm run check\` clean
- [x] \`npx jest\` — same pre-existing failures on main (\`circuit-breaker\` integration tests need live DB; \`time-language.test.ts\` has a flaky midnight-boundary case on line 30 that fails identically on clean \`main\`). My changes add 19 passing tests with zero regressions to existing tests.
- [x] **New tests** (19 total across 2 files):
  - \`conversation-summarizer.test.ts\` (+12): hint composition for each \`partnerStatus\` value, defense-in-depth null override, Haiku-lapse handling, stale/fresh detection logic, threshold constant, no-activity edge case
  - \`context-assembler.test.ts\` (+7): no-op on active conversations, no-op on first-turn-ever, cliffhanger branch with thread text, milestone branch with null, milestone branch with missing summary, hours-vs-days duration formatting
- [ ] **Manual** — after deploy, verify a >24h-gap return produces a re-orientation that references the cliffhanger (or the milestone, as appropriate)

## Related

- Parent issue: #91
- Stacked on: #92 (Phase A), which stacks on #89 → #88

1. Sequential vs parallel trade-off in B.3: is the added ~500–1500ms on the first-return turn acceptable, or should we explore the parallel-with-re-read pattern?
2. Null guardrail on \`stageJustAdvanced\`: is there a case where forcing \`lastUnresolvedThread\` to null loses real value? (The belt-and-suspenders code path catches model-lapse cases; I can't think of a case where the model's output would be better than null when the caller knows the stage just advanced, but worth a second look.)
3. Duration humanization: \`~25 hours\` vs \`1 day\` — is the 36h boundary right?

### DIFF
--- a/backend/src/services/__tests__/ai-orchestrator.test.ts
+++ b/backend/src/services/__tests__/ai-orchestrator.test.ts
@@ -65,6 +65,7 @@ const mockContextBundle = {
     recentTurns: [],
     turnCount: 0,
     sessionDurationMinutes: 5,
+    timeSinceLastUserTurnMs: null,
   },
   emotionalThread: {
     initialIntensity: null,
--- a/backend/src/services/__tests__/context-assembler.test.ts
+++ b/backend/src/services/__tests__/context-assembler.test.ts
@@ -15,6 +15,7 @@ describe('Context Assembler', () => {
         recentTurns: [],
         turnCount: 0,
         sessionDurationMinutes: 0,
+        timeSinceLastUserTurnMs: null,
       },
       emotionalThread: {
         initialIntensity: null,
--- a/backend/src/services/__tests__/context-assembler.test.ts
+++ b/backend/src/services/__tests__/context-assembler.test.ts
@@ -227,5 +227,129 @@ describe('Context Assembler', () => {
         expect(formatted).not.toContain('ABOUT THIS USER (from previous sessions)');
       });
     });
+
+    describe('Resumption framing (long-idle)', () => {
+      const HOUR = 60 * 60 * 1000;
+
+      it('does not render resumption section for active conversations', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 5,
+            sessionDurationMinutes: 10,
+            timeSinceLastUserTurnMs: 5 * 60 * 1000, // 5 min
+          },
+        });
+        const formatted = formatContextForPrompt(bundle);
+        expect(formatted).not.toContain('Resumption');
+      });
+
+      it('does not render resumption section when user has never spoken', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 0,
+            sessionDurationMinutes: 0,
+            timeSinceLastUserTurnMs: null,
+          },
+        });
+        const formatted = formatContextForPrompt(bundle);
+        expect(formatted).not.toContain('Resumption');
+      });
+
+      it('renders cliffhanger branch when summary has a lastUnresolvedThread', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 8,
+            sessionDurationMinutes: 30,
+            timeSinceLastUserTurnMs: 48 * HOUR,
+          },
+          sessionSummary: {
+            keyThemes: [],
+            emotionalJourney: '',
+            currentFocus: '',
+            userStatedGoals: [],
+            lastUnresolvedThread: 'Alice was still sitting with the feeling that trust had been eroding.',
+          },
+        });
+        const formatted = formatContextForPrompt(bundle);
+        expect(formatted).toContain('Resumption');
+        expect(formatted).toContain('Paused mid-thread');
+        expect(formatted).toContain('trust had been eroding');
+        expect(formatted).toContain('2 days'); // humanized duration
+      });
+
+      it('renders milestone branch when lastUnresolvedThread is null', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 8,
+            sessionDurationMinutes: 30,
+            timeSinceLastUserTurnMs: 26 * HOUR,
+          },
+          stageContext: {
+            stage: 2,
+            gatesSatisfied: {},
+          },
+          sessionSummary: {
+            keyThemes: [],
+            emotionalJourney: '',
+            currentFocus: '',
+            userStatedGoals: [],
+            lastUnresolvedThread: null,
+          },
+        });
+        const formatted = formatContextForPrompt(bundle);
+        expect(formatted).toContain('Resumption');
+        expect(formatted).toContain('Just entered Stage 2');
+        expect(formatted).toContain('without manufacturing unresolved tension');
+      });
+
+      it('renders milestone branch when sessionSummary is missing entirely', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 8,
+            sessionDurationMinutes: 30,
+            timeSinceLastUserTurnMs: 30 * HOUR,
+          },
+          stageContext: {
+            stage: 1,
+            gatesSatisfied: {},
+          },
+          // no sessionSummary
+        });
+        const formatted = formatContextForPrompt(bundle);
+        expect(formatted).toContain('Resumption');
+        expect(formatted).toContain('Just entered Stage 1');
+      });
+
+      it('formats short idle windows (24-35h) in hours', () => {
+        const bundle = createMinimalBundle({
+          conversationContext: {
+            recentTurns: [],
+            turnCount: 5,
+            sessionDurationMinutes: 15,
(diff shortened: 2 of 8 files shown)


slam-paws:
Good design. The four pieces compose cleanly — idle measurement feeds the refresh trigger, refresh feeds the summarizer, summarizer feeds the renderer. Each piece no-ops independently when it doesn't apply.

On your three focus areas:

### 1. Sequential vs parallel (B.3)

Sequential is the right call. The whole value proposition of this PR is that the *first return turn* gets the fresh cliffhanger — parallelizing with `assembleContextBundle` defeats that. The 500–1500ms penalty only hits users who've been away >24h with a stale summary; for them the re-orientation quality matters far more than sub-second latency. If this ever needs optimization, the cheapest path would be caching the `findLastUserTurnTimestamp` result (you already query it in `findPreviousUserTurnAt` in `slack-conversation.ts` and again inside `assembleContextBundle` via the parallel `Promise.all`). That's a minor optimization though — not worth the complexity now.

Speaking of which: `findPreviousUserTurnAt` in `slack-conversation.ts` and `findLastUserTurnTimestamp` in `context-assembler.ts` are identical queries. In the common path (non-idle), the assembler's copy runs in the `Promise.all` and the slack-conversation copy already ran and returned early. On the idle path, both run sequentially. Not a bug — just a note that if you ever consolidate, passing `previousUserTurnAt` into `assembleContextBundle` would save one DB round-trip on every turn.

### 2. Null guardrail on `stageJustAdvanced`

The belt-and-suspenders approach is correct. I can't think of a case where forcing null loses real value after a stage advance — the user crossed a gate, which by definition resolves the tension that gates track. The only theoretical edge: a user crosses a stage gate but has a *separate* unresolved emotional thread unrelated to the gate criteria (e.g., they satisfied "I feel heard" but are still sitting with anger about something tangential). In practice, the summarizer would need to distinguish gate-related tension from tangential tension, which is asking too much of a single-sentence cliffhanger. Forcing null is the safer default — the AI can still pick up tangential threads from the rolling summary and recent turns without needing an explicit `lastUnresolvedThread` to anchor on.

### 3. Duration humanization / 36h boundary

The 36h cutoff (`hours < 36`) makes sense. `~25 hours` is more informative than `1 day` for the 24–35h window because it communicates "they slept on it" vs "they disappeared for days." Past 36h, the precision doesn't add value and `2 days` reads more naturally than `~40 hours`. One edge: at exactly 36h, `Math.round(36/24)` = `2 days` which is slightly generous but fine for a "humanized" duration. The asymmetry is intentional — you'd rather the AI treat a 36h gap as "multi-day" (warmer re-orientation) than "just over a day" (might skip the orientation). Looks right.

**One small thing**: the `LONG_IDLE_THRESHOLD_MS` constant in `context-formatters.ts` has a comment saying it's "duplicated from conversation-summarizer to avoid a circular import." Both are 24h so they match today, but if someone changes one without the other, the refresh trigger and the render gate would disagree silently. Worth a `// grep: LONG_IDLE_RESUMPTION_THRESHOLD_MS` breadcrumb on the formatter's copy so a future editor finds both.

Tests look thorough — the 19 new tests cover the critical branches well, especially the defense-in-depth null override and the stale/fresh detection logic.

slam-paws:
Thanks for the thorough review — especially the nuance on the stage-gate edge case (tangential unresolved threads). You're right that asking the summarizer to distinguish gate-related vs tangential tension is beyond what a single-sentence cliffhanger should attempt. The rolling summary + recent turns will carry tangential threads naturally.

Good catch on the `findPreviousUserTurnAt` / `findLastUserTurnTimestamp` duplication. You're right that passing `previousUserTurnAt` into `assembleContextBundle` would save one DB round-trip on every turn — not worth the added parameter threading now but noted for when we consolidate.

On the grep breadcrumb for `LONG_IDLE_THRESHOLD_MS` — agreed, silent drift between the two constants would be a nasty subtle bug. I'll add the cross-reference comment on the formatter's copy pointing at `LONG_IDLE_RESUMPTION_THRESHOLD_MS` in conversation-summarizer.

slam-paws:
Sounds good on all three. The cross-reference comment on the formatter's `LONG_IDLE_THRESHOLD_MS` pointing back to `LONG_IDLE_RESUMPTION_THRESHOLD_MS` will close the last loose thread from the review. LGTM once that's in.

slam-paws:
Done — added the `grep: LONG_IDLE_RESUMPTION_THRESHOLD_MS` breadcrumb on the formatter's copy in 9cfe4f6. A future editor changing either constant will now find both via a simple grep.

slam-paws:
Verified in 9cfe4f6 — the breadcrumb is in place and reads clearly. `grep LONG_IDLE_RESUMPTION_THRESHOLD_MS` now surfaces both copies as intended.

All three review focus areas resolved + the cross-reference concern addressed. LGTM — ready to merge once the stack below (#92 → #89 → #88) lands.

slam-paws:
Thanks for verifying! Glad the breadcrumb reads well and the grep surfaces both copies cleanly. Will merge this once the stack below (#92 → #89 → #88) lands.

slam-paws:
Sounds good — standing by for the stack to land. No further changes needed on this branch.

shantamg:

slam-paws:
Noted — PR superseded by #102. The review from earlier in this thread covered all three focus areas (sequential vs parallel, null guardrail, duration humanization) and the cross-reference breadcrumb was addressed in 9cfe4f6. No further action needed here.

PR #95, opened 2026-04-17
Phase C: Session decay (TTL) — invite sweep + active/archived retention

shantamg:
Phase C of [#91](https://github.com/shantamg/meet-without-fear/issues/91). Stacked on #94 (Phase B) → #92 (Phase A) → #89 → #88.

Implements the full two-tier session decay policy. Stale sessions no longer accumulate in Postgres, and the 7-day lobby-invite TTL, 90-day active-session retention, and 180-day hard-purge are all wired up without introducing any new scheduling infrastructure.

### C.1 + C.2 — Opportunistic INVITED TTL + dead-session lookup exclusion
- \`INVITED_SESSION_TTL_MS\` = 7 days, \`INVITED_SESSION_NUDGE_THRESHOLD_MS\` = 5 days (constants, exported for tests + reuse).
- All three Slack session-lookup helpers (\`findSessionByThread\`, \`findSessionByJoinCode\`, \`findInvitedSessionForUser\`) now:
  1. Archive INVITED sessions past their 7-day TTL inline on lookup (opportunistic — no cron).
  2. Unconditionally exclude \`ABANDONED\` / \`ARCHIVED\` from results. A stale \`SessionSlackThread\` row can't resurrect a dead session.
- **Why 7 days over 3–4**: captures at least one full weekend cycle regardless of invite day, so a partner finally summoning courage on a Saturday doesn't hit \"session expired\" and spark a fresh fight.

### C.3 — Day 5–6 nudge via dynamic prompt block
- New \`PromptContext.invitedSessionNudge?: string | null\` — mobile-safe (ignored), Slack populates when relevant.
- \`finalize()\` in \`buildStagePrompt\` appends \`OPERATIONAL NUDGE:\` + the nudge text to the **dynamic** block (not static — keeps the prompt cache intact; nudge is per-turn).
- \`buildInvitedSessionNudge(session)\` in slack-conversation returns a conversational hint for INVITED sessions aged 5–7 days, including the join code + \`archive\` escape hatch. Explicit instruction to the AI: keep it conversational, don't lead with it.

### C.4 + C.5 — 90-day archive + 180-day hard-purge (DB-side)
- **Investigation findings**: Prior \"90-day sweeper\" commits (19da935, d4846a5) are both EC2 filesystem sweepers operating on \`data/mwf-sessions/\`. Nothing backend-side exists. Built it.
- New \`services/session-retention.ts\` with \`enforceSessionRetention(now)\` — pure function, two phases:
  1. Archive \`ACTIVE\` / \`WAITING\` \`CONFLICT_RESOLUTION\` sessions with no message activity for 90 days.
  2. Hard-delete \`ARCHIVED\` sessions whose \`updatedAt\` is another 90+ days old (180-day total window, gives users a window to read their own history before it's purged).
- Pre-filter on \`Session.updatedAt\` (cheap) to bound the per-candidate \`max(Message.timestamp)\` lookup. \`updatedAt\` doesn't reflect message writes, so we confirm staleness via the authoritative signal.
- New CLI wrapper \`scripts/sweep-session-retention.ts\` — invokable via \`npx tsx\`. **Scheduling deliberately left to deploy config** (wire into Render cron, EC2 crontab, or GitHub Actions Schedule as preferred). No new runtime infrastructure.

## Test plan

- [x] \`npm run check\` clean
- [x] \`npx jest\` — 1059 passing. Same three pre-existing failures on clean main: \`circuit-breaker\` (needs live DB) + \`time-language.test.ts\` (flaky midnight-boundary). Zero regressions.
- [x] **New tests** (+23 across 3 files):
  - \`slack-session-service.test.ts\` +12: lookup TTL + dead-status exclusion permutations (5 per lookup × 3 lookups, collapsed to the distinct cases; plus TTL constant sanity).
  - \`stage-prompts.test.ts\` +4: \`invitedSessionNudge\` present / absent / null / static-block-isolation.
  - \`session-retention.test.ts\` +7: archive happy path, recency-wins-over-stale-\`updatedAt\`, no-messages fallback, hard-delete phase, query scope, inspected count, constants.
- [ ] **Manual** — after deploy: wire \`sweep-session-retention.ts\` into a daily cron and confirm it reports sensible counts on the first run.

## Why this is safe to ship

- \`enforceSessionRetention\` is idempotent. Safe to run repeatedly; a session that's already ARCHIVED stays put, already-deleted sessions can't reappear.
- The opportunistic INVITED sweep on lookups doesn't block any user traffic — it archives and returns null in the same tx as the lookup.
- Hard-delete cascades are scoped by schema FKs (\`onDelete: Cascade\`) — no orphan rows.
- All policy constants are named + exported so future adjustments are one-line changes.

## Related

- Parent issue: #91
- Stacked on: #94 (Phase B) → #92 (Phase A) → #89 → #88

1. \`session-retention.ts\`: is \`Session.updatedAt\` as the archive-time proxy for the hard-delete cutoff the right call, or should we add an explicit \`archivedAt: DateTime?\` column? Current design relies on the status flip touching \`updatedAt\`, which works but couples two concerns.
2. C.3 nudge framing: is \`OPERATIONAL NUDGE:\` in the dynamic block the right structural layer, or should it live on the payload the AI's \`<dispatch>\` layer already handles?
3. Deploy scheduling: the CLI wrapper is scheduler-agnostic by design, but do you have a preference for where to wire the daily run (Render cron vs. EC2 crontab host vs. GitHub Actions)?

### DIFF
--- a/backend/src/scripts/sweep-session-retention.ts
+++ b/backend/src/scripts/sweep-session-retention.ts
@@ -1,0 +1,28 @@
+/**
+ * Sweep Session Retention
+ *
+ * CLI entrypoint for the session retention policy. Invoke via:
+ *   npx tsx src/scripts/sweep-session-retention.ts
+ *
+ * Designed to be wired into any scheduler (Render cron, EC2 crontab, GitHub
+ * Actions schedule). The policy itself is in `services/session-retention.ts`.
+ */
+
+import { enforceSessionRetention } from '../services/session-retention';
+
+async function main(): Promise<void> {
+  const result = await enforceSessionRetention();
+  // Emit to stdout so cron hosts capture results in their log aggregation.
+  // eslint-disable-next-line no-console
+  console.log(JSON.stringify({ ok: true, ...result }, null, 2));
+}
+
+main()
+  .catch((err) => {
+    // eslint-disable-next-line no-console
+    console.error('[sweep-session-retention] failed:', err);
+    process.exit(1);
+  })
+  .finally(() => {
+    process.exit(0);
+  });
--- a/backend/src/scripts/sweep-session-retention.ts
+++ b/backend/src/scripts/sweep-session-retention.ts
@@ -23,6 +23,9 @@ main()
     console.error('[sweep-session-retention] failed:', err);
     process.exit(1);
   })
+  // `.finally()` only fires on the success path here — `process.exit(1)`
+  // in `.catch()` terminates the process before `.finally()` executes, so
+  // exit code 0 is never reached on failure.
   .finally(() => {
     process.exit(0);
   });
--- a/backend/src/services/__tests__/session-retention.test.ts
+++ b/backend/src/services/__tests__/session-retention.test.ts
@@ -1,0 +1,139 @@
+/**
+ * Session Retention Tests
+ */
+
+const sessionFindMany = jest.fn();
+const sessionUpdate = jest.fn();
+const sessionDeleteMany = jest.fn();
+const messageFindFirst = jest.fn();
+
+jest.mock('../../lib/prisma', () => ({
+  prisma: {
+    session: {
+      findMany: sessionFindMany,
+      update: sessionUpdate,
+      deleteMany: sessionDeleteMany,
+    },
+    message: {
+      findFirst: messageFindFirst,
+    },
+  },
+}));
+
+import {
+  enforceSessionRetention,
+  ACTIVE_SESSION_RETENTION_MS,
+  ARCHIVED_SESSION_PURGE_MS,
+} from '../session-retention';
+
+describe('enforceSessionRetention', () => {
+  const DAY = 24 * 60 * 60 * 1000;
+  const NOW = new Date('2026-04-17T00:00:00Z');
+
+  beforeEach(() => {
+    sessionFindMany.mockReset();
+    sessionUpdate.mockReset();
+    sessionDeleteMany.mockReset();
+    messageFindFirst.mockReset();
+    sessionUpdate.mockResolvedValue({});
+    sessionDeleteMany.mockResolvedValue({ count: 0 });
+  });
+
+  it('archives ACTIVE sessions whose most recent message is older than 90 days', async () => {
+    sessionFindMany.mockResolvedValue([
+      { id: 's1', updatedAt: new Date(NOW.getTime() - 120 * DAY) },
+    ]);
+    messageFindFirst.mockResolvedValue({
+      timestamp: new Date(NOW.getTime() - 95 * DAY),
+    });
+
+    const result = await enforceSessionRetention(NOW);
+
+    expect(result.archived).toBe(1);
+    expect(sessionUpdate).toHaveBeenCalledWith({
+      where: { id: 's1' },
+      data: { status: 'ARCHIVED' },
+    });
+  });
+
+  it('does NOT archive a session whose message is newer than 90d even if updatedAt is stale', async () => {
+    // updatedAt is 120d old (session was last status-changed then), but a
+    // recent message proves the session is active. We should NOT archive.
+    sessionFindMany.mockResolvedValue([
+      { id: 's1', updatedAt: new Date(NOW.getTime() - 120 * DAY) },
+    ]);
+    messageFindFirst.mockResolvedValue({
+      timestamp: new Date(NOW.getTime() - 5 * DAY),
+    });
+
+    const result = await enforceSessionRetention(NOW);
+
+    expect(result.archived).toBe(0);
+    expect(sessionUpdate).not.toHaveBeenCalled();
+  });
+
+  it('uses updatedAt as a fallback when a session has no messages', async () => {
+    sessionFindMany.mockResolvedValue([
+      { id: 's1', updatedAt: new Date(NOW.getTime() - 200 * DAY) },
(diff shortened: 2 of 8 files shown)


slam-paws:
Nice phase — the opportunistic TTL sweep is clean, and I like that it doesn't introduce any new scheduling infrastructure. Some thoughts on the three focus areas:

### 1. `updatedAt` vs explicit `archivedAt` column

`updatedAt` works here because `@updatedAt` auto-fires on the status flip, so the 90-day hard-delete window starts exactly when you'd expect. The coupling risk Shantam flagged is real in theory — if anything else touches a session after it's been archived (a migration, a manual DB fix, a future feature that updates a field), the delete clock resets silently. But for MWF's current scale and session lifecycle, this feels like the right tradeoff. Adding `archivedAt` means a migration, updating every archive path, and carrying a nullable column that's only meaningful for one status.

If you do want a belt-and-suspenders later, an `archivedAt` column would make the intent explicit and decouple the two concerns. But I wouldn't block on it — ship with `updatedAt` and revisit if the coupling bites.

### 2. `OPERATIONAL NUDGE` in dynamic block vs `<dispatch>`

Dynamic block is the right layer. The dispatch pattern is *reactive* — the AI sees user input, decides to emit `<dispatch>TAG</dispatch>`, and a handler generates a specialized response. The nudge is *proactive* — the system injects context based on session state, and the AI weaves it in. These are different information flows:

- Dispatch = user triggers → AI routes → handler generates
- Nudge = system injects → AI incorporates

Putting it in the dynamic block is cheaper (no extra AI call), keeps it per-turn/ephemeral (won't pollute the prompt cache), and doesn't require the AI to "decide" anything — the nudge is present when relevant, absent when not. The `OPERATIONAL NUDGE:` label in the dynamic block gives the AI enough structural signal to treat it differently from conversation context.

### 3. Deploy scheduling

The CLI wrapper being scheduler-agnostic is correct. The EC2 bot instance already runs daily sweeps via crontab (`scripts/ec2-bot/crontab.txt`), so wiring `npx tsx src/scripts/sweep-session-retention.ts` into a daily slot there (e.g. `0 5 * * *` after the existing 3/4 AM jobs) is the path of least resistance. Render doesn't have cron configured today and adding it just for this feels like scope creep.

### Minor code note

In `sweep-session-retention.ts` — the `.catch(() => process.exit(1)).finally(() => process.exit(0))` pattern works because `process.exit(1)` terminates before `.finally()` runs, but it reads like `.finally()` would override the exit code. A comment clarifying that `.finally()` only fires on the success path (since `process.exit(1)` preempts it) would prevent a future reader from "fixing" it.

Overall this is solid and safe to ship — idempotent, no new infrastructure, good test coverage.

slam-paws:
Thanks for the thorough walk-through of all three focus areas.

### 1. `updatedAt` — agreed, shipping as-is

The coupling risk is theoretical at current scale and the alternative (`archivedAt` column) pulls in migration ceremony for a single-status concern. If we ever need to decouple, the migration is small and the retention logic is already isolated in `session-retention.ts` — easy to swap the proxy column later.

### 2. Dynamic block — glad the layering reads right

The dispatch-vs-nudge distinction you drew (reactive routing vs proactive injection) is a useful mental model for where future operational signals should land. Will keep that framing.

### 3. Deploy scheduling — EC2 crontab it is

`0 5 * * *` after the existing filesystem sweepers makes sense. Will wire that up as a follow-up once this merges and the base stack is on the bot host.

### Minor code note — fixed

Good catch on the `.catch/.finally` readability. Pushed d43cc6d with an inline comment clarifying that `process.exit(1)` preempts `.finally()` so the exit-code-0 path only fires on success.

slam-paws:
Sounds good on all fronts. PR is ready to merge whenever you're happy with the stack below it (#94 → #92 → #89 → #88). Once #94 lands and this rebases clean, should be good to go.

Will leave the `0 5 * * *` crontab wiring to you as the follow-up — the script is ready to drop in.

slam-paws:
Noted — will wire up the `0 5 * * *` crontab as a follow-up once this stack lands and the rebase is clean.

slam-paws:
Sounds good. The `0 5 * * *` daily-at-5am-UTC schedule lines up well — runs after overnight quiet and before any morning Slack activity. The CLI wrapper (`scripts/sweep-session-retention.ts`) is already idempotent so safe to wire into either the EC2 crontab or a GitHub Actions schedule depending on where the rest of the infra lands.

slam-paws:
Agreed — leaning toward EC2 crontab since the bot instance is already running and the script just needs `DATABASE_URL`. GitHub Actions would work too but adds a secrets-management step for the DB connection string. Either way the idempotency means we can switch later without risk. Will wire it up once the stack lands.

shantamg:

slam-paws:
Acknowledged — review was completed across the earlier thread and all three focus areas were resolved:

1. **`updatedAt` proxy** — shipping as-is; `archivedAt` column can be added later if the coupling becomes a real problem.
2. **`OPERATIONAL NUDGE` in dynamic block** — correct layer (input-side per-turn context, not output-side dispatch).
3. **Deploy scheduling** — EC2 crontab at `0 5 * * *` UTC, wiring up as a follow-up once #102 lands.

Nothing further from my side. Commits preserved via rebase in #102.

PR #96, opened 2026-04-17
Phase D: Stage 2B over Slack — classifier, Gentle Interrupt, share command

shantamg:
Phase D of [#91](https://github.com/shantamg/meet-without-fear/issues/91). Stacked on #95 (Phase C) → #94 (Phase B) → #92 (Phase A) → #89 → #88.

Implements the conversational shell for Stage 2B over Slack. The backend state machine (EmpathyDraft, EmpathyAttempt, ReconcilerShareOffer, etc.) was already transport-agnostic; this PR adds the Slack-specific primitives needed to drive it from a DM.

### D.1 — Haiku classifier for accept/revise/decline
`services/empathy-reply-classifier.ts` — `classifyEmpathyValidationReply()` returns \`{ intent, correction }\` from free-text replies. Haiku over Block Kit buttons because async Stage 2B sessions span days and button-expiry races are a UX cliff; Haiku over keyword heuristics because humans qualify (\"Acceptable, but I was more disappointed than angry\" is \`revise\` with a specific correction, not \`accept\`).

Safe fallback: \`{ intent: 'revise', correction: null }\` on any error or malformed Haiku output — keeps the conversation moving without falsely advancing a gate.

### D.2 — Gentle Interrupt for reconciler share flow
`services/slack-reconciler-notify.ts::notifyGuesserOfShareViaSlack()` posts a standalone DM to the guesser when the subject shares context via the reconciler. Fire-and-forget hook next to the existing Ably notify in `refinementFinalizeHandler`.

**Why NOT queue the context until the guesser's next turn**: if the guesser is mid-compose on a vulnerable empathy draft and we hold the share, they send an uninformed draft, then the bot replies \"BTW your partner shared this 5 min ago…\" — forcing a redo of emotional labor. Catastrophic UX cliff in the one flow we most need to get right. Posting async with explicit \"no need to reply\" framing keeps it low-pressure: the guesser sees it pop up above their text box and chooses whether to incorporate.

### D.3a — Share command detection + ReadyShare nudge (subject side)
`services/slack-empathy-exchange.ts` — \`detectShareCommand(text)\` and \`markDraftReadyToShare(sessionId, userId)\`. Wired into `handleDmMessage`: Stage-2 subjects typing \`share\` / \`send it\` / \`ship it\` short-circuit the Sonnet turn, flip their EmpathyDraft's \`readyToShare\` flag, and get a tailored confirmation (marked_ready / already_ready / no_draft / empty_draft).

Regex is lead-word-only by design — rejects mid-sentence uses like \"I want to share something\" to avoid accidentally advancing a gate during conversational turns.

**Scope note**: this iteration is the command-detection layer only. Full Stage 2 orchestration — EmpathyAttempt creation, reconciler handoff, cross-user Ably notify — stays in the existing mobile controllers. A follow-up PR (D.3b) will wire the full handoff path for Slack-only sessions, including routing guesser replies through the D.1 classifier to the existing validation controller. Keeping this PR atomic to match the other Phase commits.

### D.4 — Partner empathy rendering with blockquote
`postEmpathyRevealToSlack()` — sibling to D.2's Gentle Interrupt but framed actively: empathy reveal IS the moment that invites a reply, so the text explicitly asks \"does this land for you? Reply \`accept\`, \`revise\`, or \`decline\`\" using the D.1 classifier's vocabulary.

Relies on A.3's multi-line blockquote handling (every line gets \`>\` explicitly for determinism + testability).

### D.5 — Race test for Gentle Interrupt
Simulates guesser-mid-compose when subject shares. Asserts:
1. Interrupt posts in <100ms regardless of any concurrent operation
2. Neither operation blocks or throws due to the other
3. Two simultaneous interrupts to different guessers both land

If this test ever fails, the Queue-until-next-turn pattern we rejected during expert review is creeping back in — the failure message points back to D.2's commit for why.

## Test plan

- [x] \`npm run check\` clean
- [x] \`npx jest\` — 1102 passing, +43 new in Phase D. Same 3 pre-existing failure suites on clean main.
- [x] **New tests** (43):
  - \`empathy-reply-classifier.test.ts\` +10
  - \`slack-reconciler-notify.test.ts\` +12 (was 6; 2 for empathy-reveal helper, 4 edge cases, 2 for race scenarios, 2 re-org)
  - \`slack-empathy-exchange.test.ts\` +21
- [ ] **Manual**: after deploy, drive a Slack Stage 2 flow end-to-end, confirm accept/revise/decline classification on varied free text

## What's explicitly deferred (D.3b)

Full EmpathyAttempt orchestration for Slack-only sessions + guesser-side classifier routing. Requires service-level extraction from the existing stage2 / reconciler controllers so Slack calls the same business logic as mobile. Tracked as a follow-up on #91 — scope there is substantial enough that it warrants its own review cycle rather than bloating this PR.

## Related

- Parent issue: #91
- Stacked on: #95 (Phase C) → #94 (Phase B) → #92 (Phase A) → #89 → #88

1. D.3a scope: happy with shipping the command-detection primitives and deferring the full EmpathyAttempt orchestration to D.3b, or would you rather hold this PR until the full flow is in?
2. Classifier fallback: \`{ intent: 'revise', correction: null }\` as the safe default on any error — does that match how you'd want ambiguous replies to behave?
3. D.5 race test — the 100ms ceiling is arbitrary. Should we bind it to something more meaningful, or is a contract that the interrupt completes \"promptly\" enough?

### DIFF
--- a/backend/src/controllers/reconciler.ts
+++ b/backend/src/controllers/reconciler.ts
@@ -1029,6 +1029,19 @@ export async function refinementFinalizeHandler(
         empathyStatus: partnerEmpathyStatus,
         triggeredByUserId: user.id,
       }, { excludeUserId: user.id });
+
+      // Gentle Interrupt for Slack: no-op for mobile users, posts an async
+      // heads-up DM to Slack guessers (non-blocking so a Slack outage can't
+      // affect the mobile delivery path above).
+      const { notifyGuesserOfShareViaSlack } = await import('../services/slack-reconciler-notify');
+      notifyGuesserOfShareViaSlack({
+        sessionId,
+        guesserUserId: guesserId,
+        subjectName,
+        sharedContent: content,
+      }).catch((err) => {
+        logger.warn('[refinementFinalizeHandler] Slack gentle interrupt failed:', err);
+      });
     }
 
     successResponse(res, {
--- a/backend/src/services/__tests__/empathy-reply-classifier.test.ts
+++ b/backend/src/services/__tests__/empathy-reply-classifier.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Empathy Reply Classifier Tests
+ *
+ * Haiku is mocked so no live Bedrock. Focus: the classifier's input
+ * structuring, output validation, and fallback behavior.
+ */
+
+const getHaikuJsonMock = jest.fn();
+
+jest.mock('../../lib/bedrock', () => ({
+  getHaikuJson: (...args: unknown[]) => getHaikuJsonMock(...args),
+  BrainActivityCallType: { INTENT_DETECTION: 'INTENT_DETECTION' },
+}));
+
+import {
+  classifyEmpathyValidationReply,
+  type EmpathyReplyClassification,
+} from '../empathy-reply-classifier';
+
+const BASE_INPUT = {
+  replyText: 'yes, exactly',
+  empathyAttempt: 'It sounds like you were scared after the argument.',
+  sessionId: 's1',
+  turnId: 't1',
+};
+
+describe('classifyEmpathyValidationReply', () => {
+  beforeEach(() => {
+    getHaikuJsonMock.mockReset();
+  });
+
+  it('returns accept when Haiku says accept', async () => {
+    getHaikuJsonMock.mockResolvedValue({ intent: 'accept', correction: null });
+    const result = await classifyEmpathyValidationReply(BASE_INPUT);
+    expect(result).toEqual({ intent: 'accept', correction: null });
+  });
+
+  it('returns revise + correction when Haiku captures a qualifier', async () => {
+    getHaikuJsonMock.mockResolvedValue({
+      intent: 'revise',
+      correction: 'more disappointed than angry',
+    });
+    const result = await classifyEmpathyValidationReply({
+      ...BASE_INPUT,
+      replyText: 'Close, but actually I was more disappointed than angry.',
+    });
+    expect(result.intent).toBe('revise');
+    expect(result.correction).toBe('more disappointed than angry');
+  });
+
+  it('returns decline when Haiku says decline', async () => {
+    getHaikuJsonMock.mockResolvedValue({ intent: 'decline', correction: null });
+    const result = await classifyEmpathyValidationReply({
+      ...BASE_INPUT,
+      replyText: "no, that's not it at all",
+    });
+    expect(result.intent).toBe('decline');
+  });
+
+  it('normalizes whitespace-only correction strings to null', async () => {
+    getHaikuJsonMock.mockResolvedValue({ intent: 'revise', correction: '   ' });
+    const result = await classifyEmpathyValidationReply(BASE_INPUT);
+    expect(result.correction).toBeNull();
+  });
+
+  it('falls back to revise when Haiku returns null', async () => {
+    getHaikuJsonMock.mockResolvedValue(null);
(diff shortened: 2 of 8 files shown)


slam-paws (approved):
Clean Phase D — the pieces compose well across D.1–D.5. Addressing the three focus areas:

**1. D.3a scope: ship command-detection now, defer orchestration to D.3b — yes, good call.** The `detectShareCommand` + `markDraftReadyToShare` primitives are self-contained, well-tested, and the `handleShareCommand` switch in `slack-conversation.ts` already gives users clear feedback for all four states. Bundling the full EmpathyAttempt creation + reconciler handoff + cross-user Ably notify into this PR would have doubled the review surface for no user-visible gain until the guesser-side routing is also ready. The `readyToShare` flag is the clean seam — D.3b picks it up without touching anything here.

**2. Classifier fallback `{ intent: 'revise', correction: null }` — correct safe default.** `revise` without a correction keeps the loop open: the bot asks "what would you change?" and the user gets another shot. `accept` would falsely advance a gate, `decline` would discard emotional labor. The fallback coverage is thorough — null response, malformed intent, thrown error, whitespace-only correction all collapse to the same safe state. One minor note: the `isValidClassification` function accepts `correction: undefined` (line 675 of the diff) but the normalization block on lines 658-660 only checks `typeof result.correction === 'string'` — an explicit `undefined` falls through to `null` via the `: null` branch, so it works, but the `correction === undefined` check in the validator is load-bearing for that path. Worth a comment if you touch this again, but not blocking.

**3. D.5 race test — 100ms ceiling is fine as-is.** The test isn't asserting latency SLA, it's asserting *non-blocking behavior*: the interrupt resolves without waiting on `pendingDraft`. Sub-100ms for an in-memory mock with no I/O is effectively asserting "didn't block." If you wanted to be more precise you could assert `interruptElapsed < 10` since there's no real I/O, but 100ms already communicates the intent clearly and won't flake on CI. The failure message pointing back to D.2's commit rationale is a nice touch — turns a regression into a design-review breadcrumb.

**Other observations:**

- The dynamic import in `refinementFinalizeHandler` (`await import('../services/slack-reconciler-notify')`) + `.catch()` is the right pattern here — keeps the Slack module out of the mobile-only critical path and ensures a Slack outage can't affect the Ably delivery. Fire-and-forget done properly.

- D.4's empathy reveal CTA (`accept`, `revise <your correction>`, `decline`) matches the classifier's vocabulary exactly, which means the classifier will see clean input in the common case. Good alignment between the prompt and the parser.

- The `SHARE_COMMAND_RE` lead-word-only regex is deliberately conservative — good. The test cases for mid-sentence rejection (`"I want to share something"`, `"share this with them"`) document the boundary clearly.

LGTM — merge when #95 lands.

slam-paws:
Thanks for the thorough review and the LGTM.

Good catch on the `correction === undefined` path in `isValidClassification` (line 674). You're right that it's load-bearing — without it, `{ intent: 'revise' }` (no `correction` key at all) would fail validation and fall through to the fallback, which happens to produce the same `{ intent: 'revise', correction: null }` result anyway, but via the wrong branch. I'll add a comment noting that coupling if I touch this file again.

Will merge once #95 lands per the stack order.

slam-paws:
Sounds good — no rush on the inline comment, just flagging it while the context was fresh. Merge order after #95 makes sense given the stack.

slam-paws:
👍 Noted — will merge this once #95 lands.

shantamg:

PR #132, opened 2026-04-19
fix(mobile): don't sign out on transient Clerk token-refresh errors

shantamg:
## Symptom
Typing in a chat session at \`app.meetwithoutfear.com\`, the app suddenly redirects to the "Get Started" welcome screen (still on \`app.meetwithoutfear.com\`, not the marketing site — the web app's own signed-out screen has the same layout, easy to mistake). Console shows \`[UserSessionUpdates] Subscription error: Error: Ably connection timeout\` and \`[Realtime] Subscription setup error\`.

## Root cause
\`mobile/src/lib/api.ts\` 401-interceptor treats *all* failed token refreshes equivalently:

\`\`\`ts
try {
  const freshToken = await tokenProvider.getToken({ forceRefresh: true });
  // ... use it if valid
} catch (refreshError) {
  console.warn('[API] Failed to refresh auth token:', refreshError);
}
// If we couldn't refresh or token is invalid...
await handleAuthFailure();  // → Clerk.signOut()
\`\`\`

When \`getToken\` itself *throws* (network blip, Clerk API 5xx, brief connectivity loss — exactly the kinds of things happening when Ably also times out), we fall through to \`signOut()\`. \`(auth)/_layout.tsx\` sees \`!isSignedIn\` and \`<Redirect href="/(public)" />\` bounces the user to the welcome screen mid-chat.

## Fix
Split the refresh outcome into three cases — only authoritative signals sign out:

| Outcome | Meaning | Action |
|--|--|--|
| \`getToken\` throws | Transient (network / Clerk hiccup) | Reject this request only; keep session alive |
| Returns null | Clerk says "no session" | Sign out |
| Returns invalid-format token | Impossible state | Sign out |

Also adds a distinct \`TOKEN_REFRESH_FAILED\` error-tracking code so we can see how often case 1 fires in production.

## Out of scope
- The Ably timeouts themselves are a separate symptom of the same underlying network/Clerk instability. This fix doesn't address the realtime layer — but once we stop signing users out on transient refresh errors, short hiccups should be recoverable rather than session-ending.
- The COOP warnings in the console are Clerk OAuth popup artefacts and unrelated to this sign-out path.

## Test plan
- [x] \`npm --workspace mobile run check\` passes.
- [x] \`npm --workspace mobile run test -- --testPathPattern lib/__tests__/api\` — all 14 existing api tests pass.
- [ ] After deploy via OTA: reproduce the scenario (type in a chat session for several minutes); expect to stay signed in through any console "refresh errored (transient)" warnings rather than being redirected to the welcome screen.

### DIFF
--- a/mobile/src/lib/api.ts
+++ b/mobile/src/lib/api.ts
@@ -210,37 +210,53 @@ apiClient.interceptors.response.use(
       ) {
         originalRequest[RETRY_KEY] = true;
 
-        // Attempt to get a fresh token from Clerk (force refresh to bypass cache)
+        // Attempt to get a fresh token from Clerk (force refresh to bypass cache).
+        // We distinguish three outcomes here because only one of them should
+        // sign the user out:
+        //   1. getToken throws (network hiccup, Clerk API 5xx): transient —
+        //      reject the current request but keep the session alive so the
+        //      user doesn't get booted mid-chat.
+        //   2. getToken returns an invalid-format token: impossible state —
+        //      sign out.
+        //   3. getToken returns null (authoritative "no session"): sign out.
         if (tokenProvider) {
+          let refreshThrew = false;
+          let freshToken: string | null = null;
           try {
-            const freshToken = await tokenProvider.getToken({ forceRefresh: true });
-            if (freshToken) {
-              // Validate the fresh token before retrying
-              if (isValidJwtFormat(freshToken)) {
-                // Update the request with fresh token and retry
-                originalRequest.headers.Authorization = `Bearer ${freshToken}`;
-                return apiClient(originalRequest);
-              } else {
-                // Fresh token is also invalid - don't retry, user needs to re-auth
-                console.warn(
-                  '[API] Fresh token from provider is also invalid. User needs to sign in again.'
-                );
-                originalRequest[INVALID_TOKEN_KEY] = true;
-              }
-            }
+            freshToken = await tokenProvider.getToken({ forceRefresh: true });
           } catch (refreshError) {
-            console.warn('[API] Failed to refresh auth token:', refreshError);
+            refreshThrew = true;
+            console.warn('[API] Token refresh errored (transient):', refreshError);
+            trackError('auth', 'TOKEN_REFRESH_FAILED', originalRequest?.url || 'unknown');
           }
+
+          if (!refreshThrew) {
+            if (freshToken && isValidJwtFormat(freshToken)) {
+              originalRequest.headers.Authorization = `Bearer ${freshToken}`;
+              return apiClient(originalRequest);
+            }
+            if (freshToken) {
+              // Got a token back but it's malformed — treat as unrecoverable.
+              console.warn(
+                '[API] Fresh token from provider is invalid. User needs to sign in again.'
+              );
+              originalRequest[INVALID_TOKEN_KEY] = true;
+            }
+            // Either freshToken was null (Clerk says no session) or invalid.
+            // Fall through to handleAuthFailure below.
+            console.warn('[API] Unauthorized - session expired, signing out');
+            trackError('auth', 'SESSION_EXPIRED', originalRequest?.url || 'unknown');
+            await handleAuthFailure();
+          }
+          // If refreshThrew: do NOT sign out. Fall through and reject this
+          // request with the original 401 so callers can surface an error;
+          // the next retriable API call will try the refresh again.
+        } else {
+          // No token provider wired up at all — nothing we can do.
+          console.warn('[API] No token provider; cannot refresh auth');
+          trackError('auth', 'SESSION_EXPIRED', originalRequest?.url || 'unknown');
+          await handleAuthFailure();
         }
-
-        // If we couldn't refresh or token is invalid, the user needs to re-authenticate
-        console.warn('[API] Unauthorized - session may have expired, please sign in again');
-
-        // Track auth error
-        trackError('auth', 'SESSION_EXPIRED', originalRequest?.url || 'unknown');
-
-        // Sign out the user to force re-authentication
-        await handleAuthFailure();
       }
 
       // Create standardized API error


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes the 401 response interceptor so that transient `getToken()` failures (network blips, Clerk API 5xx) no longer sign the user out. Only authoritative "no session" (null) or malformed tokens trigger sign-out.

## Feedback

- **Correctness**: The three-way split is well-reasoned. When `getToken` throws, the request rejects with the original 401 and the session stays alive — next API call gets a fresh retry. When it returns null or a bad token, sign-out is appropriate. Logic is sound.

- **Observability**: Good call using a distinct `TOKEN_REFRESH_FAILED` error code for the transient path vs `SESSION_EXPIRED` for the authoritative one. Makes monitoring/alerting easy.

- **Explicit `else` branch**: Nice improvement — the no-`tokenProvider` path is now explicit with its own log rather than silently falling through.

- **Tests (non-blocking)**: The existing test file (`mobile/src/lib/__tests__/api.test.ts`) doesn't exercise the response interceptor at all — it mocks axios at a higher level. Given this is a behaviour change in auth handling, unit tests covering the three branches (throw → no sign-out, null → sign-out, valid → retry) would be valuable. Not blocking since the interceptor wasn't tested before either, but worth a follow-up.

## Verdict

**Approved.** Clean, focused fix for a real UX issue (users getting booted mid-chat on network hiccups). The code is well-commented and the control flow is easy to follow.

slam-paws (approved):
## PR Review — #132 fix(mobile): don't sign out on transient Clerk token-refresh errors

**Verdict: Approve** ✅

### What it does

Previously, *any* failure in the token-refresh path — including transient network errors or Clerk API 5xx responses — resulted in `handleAuthFailure()` being called, which signs the user out mid-chat. This PR restructures the 401 interceptor to distinguish three cases:

1. **`getToken` throws** (transient) → reject the current request with the original 401 but **keep the session alive**.
2. **`getToken` returns a malformed token** → sign out (unrecoverable).
3. **`getToken` returns `null`** → sign out (Clerk says "no session").

This is a correct and important fix — silently booting users on transient Clerk hiccups is a bad UX.

### Code quality

- Logic is well-structured and the three-way branching is clear.
- Good inline comments documenting the intent of each branch.
- `trackError` event names are appropriate (`TOKEN_REFRESH_FAILED` vs `SESSION_EXPIRED`).
- The `else` branch for no `tokenProvider` is a nice cleanup — previously this case fell through to the unconditional sign-out block, which worked, but was accidental rather than intentional.

1. **No test coverage for the interceptor** — The existing `api.test.ts` doesn't test the 401 retry logic at all (pre-existing gap, not introduced by this PR). Worth a follow-up to add tests for the three branches, especially the "transient throw" path that is the whole motivation for this change.

2. **Silent rejection on transient failure** — When `refreshThrew` is true, the request is rejected with the original 401 `ApiClientError`. Callers that show a generic "something went wrong" toast will handle this fine, but it's worth confirming that no caller interprets a 401 rejection as "definitely signed out" and takes its own sign-out action. From a quick read of the interceptor's downstream (`ApiClientError` with status 401), this seems safe since auth state is managed centrally, but worth a sanity check.

3. **`let refreshThrew` + `let freshToken` pattern** — Minor style note: an alternative would be a try/catch that returns early on success and falls through on throw, avoiding the boolean flag. But the current approach reads fine and matches the existing code style, so no action needed.

Looks good to merge. Nice fix. 🚀

---
*🤖 Reviewed by Slam Paws via [Claude Code](https://claude.com/claude-code)*

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Prevents the mobile app from signing users out when Clerk's `getToken({ forceRefresh: true })` fails transiently (network hiccup, Clerk 5xx). Previously, any failure in the catch block fell through to `handleAuthFailure()` → sign-out. Now the interceptor distinguishes three outcomes:

1. **getToken throws** (transient) → reject the request with the original 401 but keep the session alive
2. **getToken returns malformed token** → sign out (unrecoverable)
3. **getToken returns null** → sign out (Clerk says "no session")

## Feedback

- **[Correctness]**: The three-way branching logic is sound. The `refreshThrew` guard correctly prevents sign-out on transient errors while still allowing the 401 to propagate so callers can surface an error toast. The new `else` branch for missing `tokenProvider` is a good addition — the old code silently fell through to sign-out without logging.

- **[Tests — nice-to-have]**: The existing `api.test.ts` mocks axios at a level that bypasses interceptors entirely, so neither the old nor new 401 retry logic has unit test coverage. Not a blocker since this is pre-existing, but the three-way branch is exactly the kind of thing where a regression test would pay for itself. Consider adding interceptor-level tests in a follow-up.

- **[Tracking]**: Good use of distinct tracking labels — `TOKEN_REFRESH_FAILED` for the transient path vs `SESSION_EXPIRED` for the sign-out paths. This will make it easy to monitor the split in Mixpanel.

## Verdict

**Approve** — this is a targeted, well-commented fix for a real UX problem (users getting booted mid-chat on transient Clerk errors). The logic is clear and the comments explain the rationale for each branch.

slam-paws (approved):
## Review: PR #132 — fix(mobile): don't sign out on transient Clerk token-refresh errors

**Verdict: Approve**

### What it does

Fixes a real UX pain-point: the old interceptor treated *every* token-refresh failure identically — including transient network blips or Clerk 5xx errors — and called `handleAuthFailure()` → `signOut()`. Users were getting kicked out mid-conversation for reasons that would self-heal on the next request.

The new code cleanly separates three outcomes:

| Outcome | Behaviour |
|---|---|
| `getToken` **throws** (network error, Clerk outage) | Log + `trackError`, reject this request with the original 401, **do not sign out** |
| `getToken` returns a **malformed** token | Mark `INVALID_TOKEN_KEY`, sign out |
| `getToken` returns **null** (Clerk says "no session") | Sign out |

This is the correct distinction — transient failures shouldn't destroy the session.

### Code quality

- Logic is well-structured: `refreshThrew` flag keeps the control flow readable.
- `trackError('auth', 'TOKEN_REFRESH_FAILED', ...)` gives a new, distinct error code so dashboards can distinguish transient refresh failures from true session expirations — good observability.
- The `else` branch (no `tokenProvider`) now correctly calls `handleAuthFailure()` rather than silently falling through. This was a minor pre-existing gap that's now fixed.
- Comments are clear and explain the "why" (three-outcome reasoning) not just the "what".

1. **No test coverage for the interceptor** — `api.test.ts` doesn't exercise the 401-retry path at all (pre-existing gap, not introduced by this PR). A follow-up test that mocks `tokenProvider.getToken` to throw and asserts `signOut` is *not* called would lock this behaviour in.

2. **Repeated `trackError` calls for `SESSION_EXPIRED`** — both the `freshToken === null` path and the `!tokenProvider` path track `SESSION_EXPIRED`. Consider whether they should have distinct codes (e.g. `NO_TOKEN_PROVIDER`) so you can tell them apart in Mixpanel, but this is a nit.

LGTM — ship it.

---
*🤖 Reviewed by Slam Paws via [Claude Code](https://claude.com/claude-code)*

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Correctly distinguishes transient Clerk token-refresh errors (network hiccup, 5xx) from authoritative "no session" responses, preventing unnecessary sign-outs during momentary connectivity issues.

## Feedback
- **Correctness**: The three-way branch (throw → transient, null → expired, malformed → unrecoverable) is sound. When `refreshThrew` is true the request falls through to the standard `ApiClientError` rejection with the original 401, which lets callers surface a retriable error — good.
- **Error tracking**: `TOKEN_REFRESH_FAILED` vs `SESSION_EXPIRED` gives clean observability into which path fired. The 401 skip at line 252 (`if (status !== 401)`) prevents double-tracking — nice.
- **`else` for missing `tokenProvider`**: The old code silently fell through to sign-out even without a provider. Making this explicit is a clarity win.
- **Tests (non-blocking)**: The existing test file (`mobile/src/lib/__tests__/api.test.ts`) doesn't exercise the 401 response interceptor at all — it only covers the HTTP wrappers and `ApiClientError`. This is pre-existing, but given this PR adds meaningful branching logic in the interceptor, it would be worth adding tests for the three outcomes (token refreshed successfully, transient throw, null/malformed token) in a follow-up.

## Verdict
Approve — the logic is correct, well-commented, and the change is scoped tightly to the problem. The missing interceptor test coverage is pre-existing and non-blocking.

PR #141, opened 2026-04-20
fix(backend): stop redundant share suggestions + placeholder AI messages

shantamg:
## Summary
- Stops the redundant "share more about X" suggestions that kept regenerating every time a partner refined their empathy statement (reproduced live in a Stage-2 session with @piet).
- Stops the `[AI processing — please continue the conversation]` placeholder from ever reaching the chat; treats empty-response-after-dispatch as a retryable error instead.
- Fixes `ReconcilerShareOffer.deliveryStatus` staying `PENDING` forever when offers are accepted via the refinement path.
- De-dups the "Let's move forward — {partner} is also reflecting on your perspective" chat line that fired verbatim on every refinement cycle.

## Surfaced by
Real prod session `cmo6ki8bf000gon1xonjuwwpt`. Both partners ended up in the refinement loop; the "sharing suggestion" panel kept regenerating a substantively identical "I feel guilty about not responding…" draft three times because the `GET /reconciler/share-offer` fallback (`generateShareOffer`) didn't run the `hasContextAlreadyBeenShared` check that the asymmetric reconciler path does.

## What's in here
1. `micro-tag-parser.ts` — no more placeholder fallback (+ 3 regression tests).
2. `controllers/messages.ts` — empty-response guard moved after dispatch handling; routes through `streamError` cleanup so the client gets a proper retry prompt.
3. `controllers/reconciler.ts` (`refinementFinalizeHandler`) — sets `deliveryStatus=DELIVERED` + `deliveredAt` on accept, mirroring `respondToShareSuggestion`.
4. `reconciler/sharing.ts` (`generateShareOffer`) — runs `hasContextAlreadyBeenShared` before synthesizing a new suggestion.
5. `reconciler/circuit-breaker.ts` — new `markResultHandledAlreadyShared` helper that attaches a `SKIPPED` sentinel share-offer to the reconciler result whenever the shortcut fires.
6. `reconciler/state.ts` + `controllers/stage2.ts` (both symmetric branches) — use the helper so every downstream reader, not just the patched fallback, sees "this result was handled, don't re-offer."
7. `reconciler/state.ts` (`markEmpathyReady`) — de-dups the alignment chat message against the most recent AI message to the guesser.

Also ran a one-off SQL on prod to mark the stuck `OFFERED` share-offer in that session as `SKIPPED` so the UI cleared immediately.

## Not in this PR (next design round)
- Impasse off-ramp for cases where the two partners fundamentally can't engage in good faith (e.g. factual contradictions, one side performative). The reconciler currently treats each partner's Stage-1 content as ground truth with no epistemic check, and the dispatch-handler has no `IMPASSE` handler — the AI freelanced `HANDLE_DENIAL_IMPASSE` and `HANDLE_DIRECT_CONTACT_REQUEST` in the session, both of which landed as `Unknown tag ignored` warnings.

## Test plan
- [x] `npm run check` clean across all workspaces
- [x] `npx jest` — 61 suites, 1162 tests passing
- [ ] Manual: verify prod session no longer regenerates "I feel guilty about messages piling up" share suggestion after piet refines
- [ ] Manual: trigger an unknown dispatch tag (or force an empty-response case) and confirm client sees a retry prompt instead of `[AI processing…]`
- [ ] Manual: verify the "Let's move forward" message posts at most once per session state

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -1715,20 +1715,6 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
       // Clean accumulated text (strip <draft> and <dispatch> tags if they leaked through)
       accumulatedText = parsed.response;
 
-      // Guard: prevent empty AI responses from being saved to DB
-      if (!accumulatedText.trim() && !isDispatchMessage) {
-        logger.error(`[sendMessageStream:${requestId}] Empty AI response after tag stripping — skipping DB save`);
-        // Don't save empty message, but send error event so frontend can retry
-        if (!clientDisconnected) {
-          sendSSE(res, {
-            event: 'error',
-            data: { message: 'Empty AI response — please try again.', retryable: true },
-          });
-        }
-        res.end();
-        return;
-      }
-
       // =========================================================================
       // DISPATCH HANDLING: If dispatch tag detected, get and stream dispatched response
       // Dispatch messages are system responses - skip classifier/embeddings
@@ -1768,6 +1754,18 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
           logger.info(`[sendMessageStream:${requestId}] Unknown dispatch tag "${dispatchTag}" — using original AI response`);
           isDispatchMessage = false;
         }
+      }
+
+      // Guard: empty AI response after parsing + dispatch means the model emitted
+      // content entirely inside tags we couldn't route (e.g. an unknown dispatch
+      // tag, or a stray <draft> with no chat text). Treat as a stream error so
+      // the user message is cleaned up and the client gets a retry prompt —
+      // better than persisting a confusing placeholder or a blank AI message.
+      if (!accumulatedText.trim()) {
+        logger.error(`[sendMessageStream:${requestId}] Empty AI response after tag stripping and dispatch`, {
+          dispatchTag: dispatchTag ?? null,
+        });
+        throw new Error('Empty AI response after tag stripping and dispatch');
       }
 
       finalizeTurnMetrics(turnId);
--- a/backend/src/controllers/reconciler.ts
+++ b/backend/src/controllers/reconciler.ts
@@ -948,6 +948,8 @@ export async function refinementFinalizeHandler(
           sharedContent: content,
           refinedContent: content,
           sharedAt: now,
+          deliveryStatus: 'DELIVERED',
+          deliveredAt: now,
         },
       });
 
--- a/backend/src/controllers/stage2.ts
+++ b/backend/src/controllers/stage2.ts
@@ -34,6 +34,7 @@ import {
   getSharedContextForGuesser,
   generateShareSuggestionForDirection,
   hasContextAlreadyBeenShared,
+  markResultHandledAlreadyShared,
   incrementAttempts,
 } from '../services/reconciler';
 import { isSessionCreator } from '../utils/session';
@@ -133,6 +134,7 @@ async function triggerReconcilerAndUpdateStatuses(sessionId: string): Promise<vo
           `[triggerReconcilerAndUpdateStatuses] Context already shared B→A, skipping AWAITING_SHARING for User A`
         );
         statusA = EmpathyStatus.READY;
+        await markResultHandledAlreadyShared(sessionId, userAId, userBId);
       } else {
         statusA = hasSignificantGapsA ? EmpathyStatus.AWAITING_SHARING : EmpathyStatus.READY;
       }
@@ -193,6 +195,7 @@ async function triggerReconcilerAndUpdateStatuses(sessionId: string): Promise<vo
           `[triggerReconcilerAndUpdateStatuses] Context already shared A→B, skipping AWAITING_SHARING for User B`
         );
         statusB = EmpathyStatus.READY;
+        await markResultHandledAlreadyShared(sessionId, userBId, userAId);
       } else {
         statusB = hasSignificantGapsB ? EmpathyStatus.AWAITING_SHARING : EmpathyStatus.READY;
       }
--- a/backend/src/services/reconciler/circuit-breaker.ts
+++ b/backend/src/services/reconciler/circuit-breaker.ts
@@ -100,3 +100,44 @@ export async function hasContextAlreadyBeenShared(
 
   return false;
 }
+
+// ============================================================================
+// Helper: Mark a reconciler result "handled" via a SKIPPED sentinel share offer
+// ============================================================================
+
+/**
+ * When the context-already-shared shortcut fires, the active reconciler result
+ * still carries "significant gap / OFFER_SHARING" from its analysis. Without a
+ * marker, downstream readers (notably the GET /share-offer fallback in
+ * sharing.ts#generateShareOffer) would happily synthesize yet another redundant
+ * share suggestion.
+ *
+ * Attaching a SKIPPED ReconcilerShareOffer to the active result tells every
+ * reader "this one was handled, do not re-offer" without deleting the analysis
+ * itself (which is still useful data for status/summary endpoints).
+ */
+export async function markResultHandledAlreadyShared(
+  sessionId: string,
+  guesserId: string,
+  subjectId: string
+): Promise<void> {
+  const result = await prisma.reconcilerResult.findFirst({
+    where: { sessionId, guesserId, subjectId, supersededAt: null },
+    select: { id: true },
+  });
+  if (!result) return;
+
+  await prisma.reconcilerShareOffer.upsert({
+    where: { resultId: result.id },
+    create: {
+      resultId: result.id,
+      userId: subjectId,
+      status: 'SKIPPED',
+      skippedAt: new Date(),
+    },
+    update: {
+      status: 'SKIPPED',
+      skippedAt: new Date(),
+    },
+  });
+}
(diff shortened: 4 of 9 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes two related UX bugs: (1) the reconciler kept re-offering share suggestions after context was already shared, causing a redundant loop; (2) the micro-tag parser returned a confusing `[AI processing — please continue the conversation]` placeholder when all model output lived inside tags, which then got persisted as a real user-facing message.

## Feedback

- **[Correctness — messages.ts]**: The relocation of the empty-response guard from *before* dispatch handling to *after* is the key fix. Previously the parser's placeholder text masked the problem; with the parser now correctly returning empty, the old guard position would have short-circuited before dispatch could fill in the response. The new throw-based approach also gives better cleanup semantics (user message deleted, Ably error published) via the existing `streamError` handler. Well done.

- **[Defense in depth — reconciler]**: Good layered approach to stopping redundant share offers: `markResultHandledAlreadyShared` (SKIPPED sentinel) + `hasContextAlreadyBeenShared` guard in `generateShareOffer` + dedup in `markEmpathyReady`. Each layer independently breaks the loop, so even if one is bypassed by a race condition the others catch it.

- **[Minor — TOCTOU in dedup check, `state.ts:62-68`]**: The `lastAiMessageForGuesser` query runs *outside* the transaction, so there's a small window where a concurrent call could insert the same alignment message between the check and the transaction's `create`. Worst case is one duplicate (same behavior as before this PR), so not blocking — but moving the `findFirst` inside the `tx` scope would eliminate the race entirely if you want to tighten it up later.

- **[Schema alignment — reconciler.ts]**: Setting `deliveryStatus: 'DELIVERED'` and `deliveredAt` in `refinementFinalizeHandler` is a good catch — previously the share offer was accepted/shared but the delivery tracking fields were left at defaults.

- **[Testing]**: Solid regression tests for the parser change covering thinking-only, dispatch-only, and draft-only cases. The reconciler path changes are integration-heavy and harder to unit test, but the guards are straightforward enough that the risk is low.

## Verdict

**Approve** — clean, well-commented fix with proper defense in depth. The TOCTOU note is a minor hardening opportunity, not a blocker.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes two related UX bugs in the reconciler loop: (1) redundant share suggestions kept re-appearing after context was already shared, and (2) placeholder `[AI processing — please continue the conversation]` messages leaked into chat when AI output was entirely inside semantic tags. Clean, well-reasoned fix across multiple code paths.

## Feedback

**Reconciler de-dup (`state.ts`, `circuit-breaker.ts`, `sharing.ts`, `stage2.ts`)**

- **[Correctness]** The three-pronged approach to closing the redundant-share-suggestion loop is solid:
  1. `markResultHandledAlreadyShared` creates a SKIPPED sentinel on the active result — prevents downstream `generateShareOffer` from re-drafting.
  2. `hasContextAlreadyBeenShared` guard inside `generateShareOffer` catches the GET fallback path.
  3. De-dup guard in `markEmpathyReady` prevents duplicate alignment messages.

  All three call sites (`state.ts:runReconcilerForDirection`, `stage2.ts:triggerReconcilerAndUpdateStatuses` × 2) are covered. Nice.

- **[Correctness — minor]** The `alreadyPosted` check in `markEmpathyReady` reads the last AI message _outside_ the transaction, creating a small TOCTOU window where two concurrent reconciler runs could both see no duplicate and both create the alignment message. Since this is strictly additive (worst case matches pre-PR behavior) and the alternative would be a Serializable transaction, this is fine as-is. Just noting it.

- **[Types]** `markEmpathyReady` return type stays `Promise<void>` while the transaction now returns `Message | null` — the early `if (!savedMessage)` return narrows correctly before `publishMessageAIResponse` accesses `savedMessage.id`. Looks clean.

**Micro-tag parser (`micro-tag-parser.ts`)**

- **[Correctness]** Removing the `[AI processing...]` placeholder and returning empty is the right call. The caller (`sendMessageStream`) is now the single decision-maker for empty responses, which is cleaner separation of concerns. Good regression tests covering thinking-only, dispatch-only, and draft-only cases.

**Empty-response guard (`messages.ts`)**

- **[Correctness]** Moving the guard _after_ dispatch handling is the key fix — this prevents false positives where the model emits valid `<dispatch>` content with no visible text. Throwing an error (instead of the old early-return) routes through the stream-error cleanup path, which deletes the user message and sends a retry prompt. This is actually better UX than before, where the user message was kept in DB but the AI response was silently dropped.

**Delivery status (`reconciler.ts`)**

- **[Correctness]** Adding `deliveryStatus: 'DELIVERED'` and `deliveredAt: now` when a share offer is ACCEPTED is a straightforward data-integrity fix — the content is shared immediately, so marking it delivered at write time is correct.

**Tests**

- Three new micro-tag-parser tests are well-targeted. No new tests for `markResultHandledAlreadyShared` specifically, but the existing `reconciler.test.ts` already covers the "context already shared → READY" shortcut, and the SKIPPED sentinel is validated by the route-level tests. Adequate coverage.

## Verdict

**Approve.** This is a well-scoped fix for a real production issue. The multi-path approach to closing the redundant-suggestion loop is thorough without being over-engineered, and the micro-tag-parser change is a clean separation-of-concerns improvement.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes two related bugs: (1) the micro-tag parser's placeholder fallback (`[AI processing — please continue the conversation]`) was leaking into user-visible messages when the model emitted only tags, and (2) redundant share suggestions kept re-firing after context had already been shared, making the UX feel stuck.

## Feedback

**Good design decisions:**

- **[Correctness]** Moving the empty-response guard in `messages.ts` to *after* dispatch handling is the right call. The old guard ran before dispatch could fill in `accumulatedText`, meaning it couldn't catch the case where dispatch failed/returned null. Now the guard correctly covers all paths — and by throwing into the existing catch block, the error handling is unified (user message cleanup, Ably error, SSE error) rather than duplicated inline.

- **[Architecture]** The multi-layered approach to preventing redundant share suggestions is solid: SKIPPED sentinel via `markResultHandledAlreadyShared`, the `hasContextAlreadyBeenShared` guard in `generateShareOffer`, and the de-dup check in `markEmpathyReady`. Belt-and-suspenders is appropriate here since the reconciler runs iteratively from multiple entry points.

- **[Data integrity]** Setting `deliveryStatus: 'DELIVERED'` + `deliveredAt` in `refinementFinalizeHandler` closes a gap where accepted shares could sit with `PENDING` delivery status.

- **[Tests]** The three new regression tests in `micro-tag-parser.test.ts` clearly document the intended behavior for thinking-only, dispatch-only, and draft-only cases.

**Minor observations (non-blocking):**

- **[Race window — `state.ts:64-65`]** The `alreadyPosted` check reads the last AI message outside the transaction, then the skip decision happens inside. This is a TOCTOU gap — two concurrent `markEmpathyReady` calls could both see `alreadyPosted = false` and both create the message. In practice this is a significant improvement over the current "always create" behavior and the reconciler already tolerates concurrent calls, so this is fine as-is. Just noting for awareness.

- **[Dispatch edge case]** If `handleDispatch` returns an empty string (`""`), the new guard now throws, whereas the old code would have skipped it due to the `!isDispatchMessage` check. This is actually better behavior — an empty dispatch response shouldn't be persisted either.

## Verdict

Approve. The changes are well-reasoned, correctly address both bugs, and the test coverage is good.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes two related bugs in the reconciler loop: (1) redundant share suggestions kept being re-synthesized after context was already shared, and (2) the micro-tag parser's `[AI processing — please continue the conversation]` placeholder was leaking into the DB as a user-visible message.

## Feedback

**Good patterns:**

- **Defense in depth on redundant suggestions** — The SKIPPED sentinel on `ReconcilerShareOffer` catches "this specific result was handled", while the `hasContextAlreadyBeenShared` check in `generateShareOffer` catches "context was shared in any previous cycle, even if the result was superseded/recreated". Belt and suspenders — well done.
- **Moving the empty-response guard after dispatch** (`messages.ts:1756-1770`) — The old position blocked dispatch handlers from filling in content. Moving it after dispatch + throwing into the existing error handler is cleaner and avoids duplicating the user-message cleanup logic.
- **Dedup guard in `markEmpathyReady`** — Checking the last AI message before creating a new one prevents identical alignment messages from stacking up, while still running the status transition unconditionally. The null-message early return correctly skips the Ably publish but still calls `checkAndRevealBothIfReady`.
- **Good regression tests** on the micro-tag parser covering the three empty-response shapes (thinking-only, dispatch-only, draft-only).

**Minor observations (non-blocking):**

- **`markEmpathyReady` dedup read is outside the transaction** (`state.ts:64-67`): Two concurrent calls could both see `alreadyPosted = false` and both create the message. In practice the reconciler flow is sequential per session, so this is unlikely — but worth a note for future readers. If it ever becomes a concern, moving the read inside the transaction would close the window.
- **`sharing.ts:1050` — early return placement**: The new `hasContextAlreadyBeenShared` check is placed after the share-offer status check but *before* `getWitnessingContent` and the AI calls. Good placement — it short-circuits before the expensive operations.

## Verdict

**Approve** — Clean, well-layered fix that addresses the root causes at multiple points in the pipeline. The micro-tag parser change eliminates the placeholder message, the guard relocation in `messages.ts` lets dispatches work correctly, and the redundant-suggestion prevention is thorough without being over-engineered. Tests cover the parser changes. Ship it.

Issue #173, opened 2026-04-25
Investigate 88% session abandonment rate (11% completion)


slam-paws:
## Problem

Over the last 7 days, only 11% of sessions reach completion — 88% are abandoned. With 3 active users and 22 messages sent, this is the most critical product health metric to understand.

## What to investigate

• Where in the session flow are users dropping off? (which stage?)
• Are there error patterns correlated with drop-off (109 network errors in the same window)?
• Is #169 (PERSPECTIVE_STRETCH stage completion failure) a contributing factor?
• Are there UX signals (session length, time-to-abandon) that reveal the cause?

## Priority

Darryl has designated this the *most important thing* — investigate before all other items.

## Provenance

- Channel: #most-important-thing
- Requester: Darryl Finkton Jr (U0ASRE7NZL7)
- Original message: Let's do move that to the top for the next most important thing

---

## Gathered Context

**Category**: Bug (multi-layered; two distinct investigations)

---

### Stage 2 (PERSPECTIVE_STRETCH) — Root cause found, fixed in PR #557

**Feature/screen**: Stage 2 empathy exchange UI (`mobile/src/hooks/useChatUIState.ts`)

**Steps to reproduce**:
1. User A writes an empathy draft — `EmpathyAttempt.status` → `HELD`
2. Partner (User B) shares context — `EmpathyAttempt.status` → `REFINING`
3. Realtime delivers the shared-context message to User A
4. The Revisit / Resubmit panel disappears because `isRefiningEmpathy` was derived from `hasNewSharedContext` only (a transient flag), not from the persisted DB state `EmpathyAttempt.status === 'REFINING'`
5. User A is stuck — no visible UI path to resubmit — even though the backend is waiting for a resubmission

**Expected**: Revisit panel remains visible whenever `EmpathyAttempt.status === 'REFINING'` (persisted)
**Actual**: Panel disappears after the realtime context-delivery event clears `hasNewSharedContext`

**Evidence**: 0% PERSPECTIVE_STRETCH completion across 10+ starts in April, 0% across 50+ starts in May pre-fix. Users re-entered the stage up to 11 times in a single session without completing.

**Fix (PR #557)**: `isRefiningEmpathy` now checks both `hasNewSharedContext` AND `EmpathyAttempt.status === 'REFINING'`. Verified across 4 consecutive Adam/Eve + 1 James/Catherine gold-loop runs with `score=4.0` and `target_reached=True`.

---

[...]

Issue #169, opened 2026-04-24
Investigate PERSPECTIVE_STRETCH stage completion failure


slam-paws:
## Problem

PERSPECTIVE_STRETCH stage appears stuck — user_38 looped through it 4 times across 3 sessions without completing. Only 17% of stage starts completed in the last 24h (1 of 6).

## Context from Shantam

- He already fixed a bug where the button for confirming needs wasn't showing — it's now working
- He got to that stage with Jamal but Jamal hasn't engaged since
- The low completion rate may be explained by the button bug, not a deeper UX or logic issue
- He also wants to try it in a real session to validate

## Investigation scope

1. Check if the button fix landed before or after user_38's failed attempts — if after, the fix likely explains the failures
2. Review PERSPECTIVE_STRETCH stage logic for any remaining dead-ends or unclear CTAs
3. Check completion metrics since the button fix to see if the rate improved

## Provenance

- Channel: #most-important-thing
- Requester: Darryl Finkton Jr (U0ASRE7NZL7)
- Agreed by: Darryl + Shantam (U0AD3TF2U7L)
- Original context: Evening check-in strategy briefing (2026-04-20)

slam-paws:
## Investigation Report

**Root cause**: Partner disengagement — not a code bug. Stage 2 requires mutual empathy exchange (both users write drafts, both reach READY, mutual reveal, both validate). When one partner stops engaging, the other is stuck indefinitely with no timeout or re-engagement mechanism.

**Evidence**:

| Check | Finding |
|-------|---------|
| **DB: user_38 (Darryl) session** | Darryl submitted empathy (status=READY, Apr 19 03:50, version 5 = 4 refinements). Shantam never created an empathy draft in this session. Mutual reveal never fired. `gatesSatisfied = {}`. |
| **DB: Completion rates** | 2 of 6 stage-2 entries ever completed (33%). Both completions from the first session (Shantam/Jamal, Apr 16). Every session after Apr 16 is stuck as IN_PROGRESS. |
| **DB: Post-fix sessions** | Shantam/piet session started Apr 20 (after both fixes) — also stuck in Stage 2. Fixes did not resolve the pattern. |
| **Git: NeedsDrawer button fix (89860bc, Apr 17)** | This fix is for **Stage 3 (NEED_MAPPING)**, not Stage 2. It constrained the NeedsDrawer layout so confirm buttons stayed visible. Does not explain Stage 2 failures. |
[...]

Issue #152, opened 2026-04-20
MEDIUM: Security audit — consolidated medium-severity findings (Apr 2026)


slam-paws:
## Summary

Consolidated medium-severity findings from the April 2026 weekly security audit.

### 1. Missing Rate Limits on LLM-Backed Endpoints
`/sessions/:id/messages/stream` and Stage 2-3 reconciliation endpoints lack rate limiting. The `streamingRateLimit` (10/min) and `empathyRateLimit` (20/min) are defined but not applied to these expensive endpoints. Global rate limiter defined but not applied.

### 2. Public Invitation Endpoint
`GET /invitations/:id` requires no authentication. Returns inviter name, session ID, and relationship info. Risk of enumeration.

### 3. Slack Health Endpoint Information Leak
`GET /api/slack/health` is public and returns whether Slack is configured and if shared secret is required — useful reconnaissance for attackers.

### 4. Secrets Not Validated at Startup
Missing `CLERK_SECRET_KEY` only caught on first authenticated request. Server boots and handles requests before discovering missing secrets.

### 5. Analytics Consent Missing
No Mixpanel opt-in mechanism. `setUserProperties()` allows arbitrary properties to be set — no audit of what's tracked.

### 6. No Incident Response Plan
No formal incident response documentation, breach notification procedures, or escalation runbooks.

### 7. npm Dependency Cleanup
- 55 npm audit findings (42 HIGH) — almost all from `vercel` CLI in docs-site devDeps
- `@types/*`, `typescript`, `tsc-alias` in backend production dependencies (should be devDeps)
- Deprecated `@types/ably` should be removed
- Zod 3→4 migration incomplete (backend vs shared version split)

### 8. Inner Work Sessions Missing Middleware Auth Check
`/inner-thoughts/:id` routes rely on controller-level ownership checks rather than middleware-enforced `requireInnerWorkAccess`.

## Provenance
- **Channel:** #security-audit
- **Requested by:** weekly security audit (automated)
- **Original message:** weekly security audit cron
- **Prompt(s) used:** 7-agent parallel security scan across all domains

slam-paws:
## Investigation Report

**Root cause**: Confirmed 8 medium-severity security findings from the April 2026 audit. All findings verified against current codebase. Severity and fixability assessed below.

**Evidence**: Direct code inspection of route definitions, middleware, package.json files, and controller logic across backend, mobile, and shared workspaces.

---

### Finding-by-Finding Assessment

[...]

Issue #151, opened 2026-04-20
HIGH: GDPR/CCPA compliance gaps — incomplete deletion, no data export, no COPPA


slam-paws:
## Summary

Multiple GDPR/CCPA compliance gaps identified across deletion, data portability, age verification, and analytics consent.

### 1. Incomplete Data Deletion (GDPR Article 17)
Account deletion (`DELETE /auth/me`) removes database records but does NOT delete from:
- Mixpanel (no API call to delete analytics)
- Sentry (no mechanism to purge error logs)
- AssemblyAI (no cleanup of transcripts)
- Ably (no event history cleanup)
- Render database backups (unknown retention)

### 2. No Data Export / Subject Access Request (GDPR Article 15)
No endpoint exists for users to export their data in a portable format. Required for regulatory compliance.

### 3. No COPPA Age Verification
No age verification at signup. The app handles sensitive emotional/conflict content that is inappropriate for minors without safeguards. Crisis detector identifies child-related patterns but takes no protective action.

### 4. No Analytics Consent
Mixpanel tracking begins without explicit user opt-in. No consent mechanism for analytics data collection.

## Remediation

1. **Data deletion**: Add API calls to Mixpanel, Sentry deletion APIs on account deletion
2. **Data export**: Implement `GET /api/v1/auth/me/export` returning all user data as JSON
3. **Age verification**: Add age gate at signup (13+ minimum) or parental consent flow
4. **Analytics consent**: Add Mixpanel opt-in during onboarding; respect opt-out

## Provenance
- **Channel:** #security-audit
- **Requested by:** weekly security audit (automated)
- **Original message:** weekly security audit cron
- **Prompt(s) used:** Compliance + Data Security agent scans

slam-paws:
## Investigation Report

**Root cause**: All four GDPR/CCPA compliance gaps confirmed with code-level evidence. The gaps are real, not false positives.

**Evidence**:

### 1. Incomplete Data Deletion — CONFIRMED
`DELETE /auth/me` handled by `account-deletion.ts` correctly deletes DB records (cascade + manual cleanup of `PreSessionMessage`, anonymizes `GlobalLibraryItem.contributedBy` and `ReconcilerResult` names) and removes the Clerk user. However, **no external service cleanup exists**:
- **Mixpanel**: No `people.delete()` call — user events/profiles persist forever
- **Sentry**: No user data deletion — error reports with user context remain
[...]

PR #179, opened 2026-04-26
fix(slam-bot): stop needs-info re-dispatch loop

shantamg:
## Summary

- Five `bot:needs-info` issues (#150, #151, #152, #169, #173) were burning Claude Code credits every 30 min — ~936 Sonnet runs over 6 days, with daily volume climbing 76→231 as more issues fell into the trap.
- Root cause: `stages/01-create-issue/CONTEXT.md` told the LLM to *"skip to stage 02 on the next dispatcher tick"* on re-entry. But the dispatcher always re-enters at stage 01 (the registry has no per-issue stage state), so stage 02 (where the `waiting-human-${N}.txt` marker that gates re-dispatch is written) was never reached. Result: every 30-min cooldown expiry → another Sonnet fire → repeat forever.
- Fix: route on re-entry from `needs-info/CLAUDE.md` based on the bot-comment meta tag (same pattern `spec-builder/CLAUDE.md` and `research/CLAUDE.md` already use). Stage 01 inlines stage 02's logic on re-entry instead of deferring, and writes the marker as a failsafe on every exit path.
- Already shipped manually before this PR: wrote `waiting-human-{150,151,152,169,173}.txt` on EC2 to halt the live loop. The dispatcher saw the markers on the next tick and stopped firing on those issues.

## Why the bug bit only `needs-info`

`bot:spec-builder` and `bot:research` both have L1 `CLAUDE.md` instructions to determine the current stage from the meta tag at the top of each invocation. `bot:needs-info` was missing that pattern, so it relied on stage 01 to detect re-entry and route — but stage 01's instruction said "exit and let the next tick do it," which never worked because the next tick was identical.

## Damage

| Issue | Cycles | Days looping |
|---|---|---|
| #150 | 262 | 6 |
| #151 | 263 | 6 |
| #152 | 263 | 6 |
| #169 | 98 | 2 |
| #173 | 50 | 1 |

## Out of scope (followups)

- **Marker auto-clearing on human reply.** `pipeline-monitor.sh`'s Check 7 (the only thing that clears markers when a non-bot user comments) only runs while a `bot:milestone-builder` issue is open. With this PR's marker reliably written, the bot won't re-engage on a non-`@MwfBot` reply unless the human removes/re-adds the label. The cleanest fix is probably extending `clear-stale-locks.sh` to scan markers, compare to the state file's `updatedAt`, and fetch comments only when the issue has changed since the marker was written.
- **Empirical audit of `bot:spec-builder`** — its L1 routing should keep it out of this trap, but no harm in a quick check.

## Test plan

- [ ] Verify the 5 stuck issues stay quiet after this lands (no new `bot:in-progress` label cycling)
- [ ] Trigger a fresh `bot:needs-info` flow end-to-end: confirm marker is written on first run, dispatcher skips on subsequent ticks
- [ ] Reply on a `bot:needs-info` issue with `@MwfBot` and confirm `check-github.sh` still routes correctly (independent of this change)

### DIFF
--- a/bot-workspaces/needs-info/CLAUDE.md
+++ b/bot-workspaces/needs-info/CLAUDE.md
@@ -2,6 +2,15 @@
 
 Structured interview loop for vague or incomplete requests. Asks clarifying questions, polls for responses, and graduates the issue when enough context is gathered.
 
+## Modes
+
+The dispatcher always invokes this workspace at `entry_stage: 01-create-issue` because there is no per-issue stage state in the registry. Determine the actual current stage from the issue's bot comments before loading a stage CONTEXT.md:
+
+- **No bot comment with `<!-- bot:needs-info-meta` metadata exists** → fresh issue, load `stages/01-create-issue/CONTEXT.md`.
+- **A bot comment with `<!-- bot:needs-info-meta` metadata exists** → re-entry, load `stages/02-interview/CONTEXT.md` directly. Do NOT exit early expecting a "next tick" to handle stage 02 — the next tick will re-enter at stage 01 again, producing an infinite re-dispatch loop (every 30 min, while `bot:needs-info` carries `keep_label: true`).
+
+The `waiting-human-${ISSUE_NUMBER}.txt` marker file in the claims dir is the only signal that gates re-dispatch. Whichever stage runs MUST leave that marker present on every exit path that does not graduate or close the issue.
+
 ## What to Load
 
 | Resource | When | Why |
--- a/bot-workspaces/needs-info/stages/01-create-issue/CONTEXT.md
+++ b/bot-workspaces/needs-info/stages/01-create-issue/CONTEXT.md
@@ -31,7 +31,9 @@ Any other `gh` read call indicates a bug.
 ## Process
 
 1. **Confirm the label** via the state file helper `github_state_issue_has_label`. Then **read the issue body** via `gh issue view <number>`.
-2. **Check for existing bot comments** on the issue. If the bot has already posted an initial comment, skip to stage 02 (this is a re-entry).
+2. **Detect re-entry**: check for existing bot comments on the issue.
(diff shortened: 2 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes an infinite re-dispatch loop in the `needs-info` workspace. The root cause: the dispatcher always enters at `01-create-issue` (no per-issue stage state), but stage 01 previously told the LLM to "skip to stage 02 on the next tick" — which would just re-enter stage 01 again, every 30 minutes. The fix makes stage 01 inline stage 02's logic on re-entry and writes the `waiting-human` marker on every exit path.

## Feedback

- **Correctness**: The fix is well-designed. Re-entry detection via `<!-- bot:needs-info-meta` metadata is robust, inlining stage 02 avoids the loop, and the new step 6 marker write prevents re-dispatch after first-run. The failsafe is good defensive design.

- **Minor inconsistency** (non-blocking): The **Orchestrator Rules** section in `bot-workspaces/needs-info/CLAUDE.md:33` still says _"Re-entry: stage 02 re-enters on each dispatcher tick until graduation or stale"_ — but the new behavior is that stage 01 re-enters and inlines stage 02. The new **Modes** section (which the LLM reads first) is clear and authoritative, so this is unlikely to cause real confusion, but updating the Orchestrator Rules to match would remove the contradiction. Something like:
  > Re-entry: stage 01 re-enters on each dispatcher tick and inlines stage 02 when metadata exists; the waiting-human marker gates re-dispatch until graduation or stale

- **Good patterns**: Clear explanation of *why* the loop happens in the Modes section, explicit "Do NOT exit early" instruction, and the failsafe catch-all — all reduce the chance of LLM misrouting.

## Verdict

**Approve.** Clean fix that addresses the root cause with good defensive layers. The Orchestrator Rules nit is cosmetic — ship it.

PR #241, opened 2026-05-01
fix(mobile): wire web Mixpanel to real mixpanel-browser SDK

shantamg:
## Summary
- `mobile/src/services/mixpanel.web.ts` was a console-only no-op stub. Every `track`/`identify`/`alias` call from the Expo web build silently dropped on the floor, so users who used the browser version of the app (e.g. Stephanie) never appeared in Mixpanel.
- Replaced with a real implementation backed by `mixpanel-browser`, mirroring the surface of `./mixpanel` (native) so call sites in `analytics.ts` and elsewhere remain platform-agnostic.
- Falls back to the previous console-logging no-op when `EXPO_PUBLIC_MIXPANEL_TOKEN` is absent — matches native behavior.
- Preserves the existing `identify()` dedupe (no-op when the same `userId` re-identifies).
- Adds `mixpanel-browser ^2.78.0` to `mobile/package.json` (already present in the website workspace).

## Verification
- `npm run check` passes across all workspaces.

## Deploy note
`EXPO_PUBLIC_MIXPANEL_TOKEN` must be set at web build time wherever the web app is built/hosted. Expo inlines `EXPO_PUBLIC_*` vars at build time, so if native already has the token, the same `.env` carries through — but if web is built in a separate CI lane, double-check the env there. Without the token, the new code logs `[Mixpanel Web] Initialized (No-Op Mode — no EXPO_PUBLIC_MIXPANEL_TOKEN)` and silently no-ops.

## Test plan
- [ ] Confirm `EXPO_PUBLIC_MIXPANEL_TOKEN` is plumbed into the web build env
- [ ] Load the web build, sign in, take a session action
- [ ] Confirm events appear in Mixpanel under the signed-in `user_*` distinct_id
- [ ] Confirm anonymous-before-signin events get aliased to the user via the existing `alias()` flow

### DIFF
--- a/mobile/package.json
+++ b/mobile/package.json
@@ -56,6 +56,7 @@
     "fbjs": "^3.0.5",
     "lucide-react-native": "^0.562.0",
     "metro-react-native-babel-transformer": "^0.77.0",
+    "mixpanel-browser": "^2.78.0",
     "mixpanel-react-native": "^3.1.3",
     "react": "19.1.0",
     "react-dom": "19.1.0",
--- a/mobile/src/components/MixpanelInitializer.tsx
+++ b/mobile/src/components/MixpanelInitializer.tsx
@@ -126,7 +126,7 @@ export function MixpanelInitializer() {
         // Set user properties
         setUserPropertiesOnce({
           first_seen_at: new Date().toISOString(),
-          signup_source: 'mobile',
+          signup_source: Platform.OS === 'web' ? 'web' : 'mobile',
         });
 
         setUserProperties({
--- a/mobile/src/services/mixpanel.web.ts
+++ b/mobile/src/services/mixpanel.web.ts
@@ -77,6 +77,7 @@ export const alias = (aliasId: string): void => {
   }
   try {
     mixpanel.alias(aliasId);
+    console.log('[Mixpanel Web] alias() completed for user:', aliasId);
   } catch (error) {
     console.error('[Mixpanel Web] alias() failed:', error);
   }
--- a/mobile/src/services/mixpanel.web.ts
+++ b/mobile/src/services/mixpanel.web.ts
@@ -1,13 +1,19 @@
 /**
- * Web stub for Mixpanel analytics.
- * Uses console logging in dev; no-op in production.
- * Can be replaced with mixpanel-browser SDK later for real web analytics.
+ * Web implementation of the Mixpanel analytics service.
+ *
+ * Mirrors the surface of `./mixpanel` (native) so call sites in `analytics.ts`
+ * and elsewhere remain platform-agnostic. Metro picks this file for web targets.
+ *
+ * No token → falls back to a console-logging no-op (matches native behavior).
  */
 
+import mixpanel from 'mixpanel-browser';
+
 let didInit = false;
+let hasToken = false;
 let identifiedUserId: string | null = null;
 
-const log = (message: string, ...args: unknown[]) => {
+const log = (message: string, ...args: unknown[]): void => {
   if (__DEV__) {
     console.log(`[Mixpanel Web] ${message}`, ...args);
   }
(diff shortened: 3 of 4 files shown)


slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary

Replaces the console-logging web stub (`mixpanel.web.ts`) with a real `mixpanel-browser` SDK integration, while preserving the no-op fallback when `EXPO_PUBLIC_MIXPANEL_TOKEN` is absent.

## Feedback

- **Correctness**: The API surface matches the native `mixpanel.ts` exactly — all exported functions have the same signatures and semantics. The `hasToken` guard cleanly separates real-SDK vs no-op paths.
- **Consistency**: The try/catch pattern mirrors native precisely (`identify`, `alias`, `setUserProperties`, `setUserPropertiesOnce` wrapped; `track`, `registerSuperProperties`, `reset` unwrapped). Good.
- **Config**: `track_pageview: false` and `persistence: 'localStorage'` are the right defaults for this app's architecture.
- **Minor nit (non-blocking)**: Native `alias()` logs on success (`console.log('[Mixpanel] alias() completed for user:', aliasId)`) — the web version doesn't. Not a problem, just a small behavioral difference if anyone is ever debugging alias flows on web.

## Verdict

**Approve** — Clean, well-scoped change that properly wires web analytics with the same graceful-fallback pattern as native. No issues found.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Replaces the web Mixpanel no-op stub with the real `mixpanel-browser` SDK, gracefully falling back to console-logging when `EXPO_PUBLIC_MIXPANEL_TOKEN` is absent. Also correctly sets `signup_source` to `'web'` on web platforms.

## Feedback
- **Good pattern**: The token-present/absent branching mirrors the native service cleanly. Metro platform resolution keeps call sites platform-agnostic — nice.
- **Minor (non-blocking) — `identify()` retry behavior**: In `mixpanel.web.ts`, `identifiedUserId` is set *before* the `try/catch` around `mixpanel.identify()`. If the SDK call throws, the local guard prevents retrying on the next call. The native version (`mixpanel.ts:110`) sets `identifiedUserId` inside the `try` block so a failed call is retried. Consider moving the assignment after the `mixpanel.identify()` call for consistency:
  ```ts
  try {
    mixpanel.identify(userId);
    identifiedUserId = userId;  // ← move here
  } catch (error) { ... }
  ```
  (The no-op branch above can keep it where it is since it can't fail.)

## Verdict
**Approved** — clean, well-scoped change. The `identify` ordering is a minor robustness nit, not a blocker.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Replaces the web Mixpanel no-op stub with a real `mixpanel-browser` SDK integration, while preserving the graceful no-token fallback. Also makes `signup_source` platform-aware in `MixpanelInitializer`.

## Feedback

- **Good**: Clean implementation that mirrors the native service's export surface exactly. The `hasToken` guard pattern keeps the no-op fallback working identically to before when no token is set.
- **Good**: `track_pageview: false` and `persistence: 'localStorage'` are the right choices for an Expo web SPA.
- **Good**: `Platform.OS` check for `signup_source` is the right fix to get accurate attribution.
- **Minor (non-blocking) — `identify` ordering** (`mixpanel.web.ts`): `identifiedUserId` is set *before* the `try/catch` SDK call. If `mixpanel.identify()` throws, the userId is cached anyway, so subsequent calls with the same userId silently skip without ever actually identifying. The native implementation (`mixpanel.ts:108-113`) sets `identifiedUserId` only *after* the SDK call succeeds. Suggest moving `identifiedUserId = userId` inside the `try` block after `mixpanel.identify(userId)` for consistency:
  ```ts
  try {
    mixpanel.identify(userId);
    identifiedUserId = userId;
  } catch (error) {
    console.error('[Mixpanel Web] identify() failed:', error);
  }
  ```

## Verdict
**Approve** — solid, well-structured change. The identify ordering nit is minor and can be fixed in a follow-up if desired.

PR #243, opened 2026-05-01
[codex] Cap bot cleanup GitHub calls

shantamg:
## Summary

- Apply the same guard pattern as Lovely PR LvlyAI/lovely#2420 to Slam bot cleanup.
- Reserve GraphQL budget before stale worktree/branch cleanup asks GitHub for PR state.
- Prevent overlapping GitHub-backed cleanup runs and cap PR-state checks per run.

## Why

Lovely bot hit GitHub's GraphQL rate limit because `clear-stale-locks.sh` was doing many `gh pr list --json state` calls while sweeping old worktrees/branches. Meet Without Fear has the same cleanup pattern, and those calls share the budget used by `github-state-scanner` and review-request handling.

This keeps stale cleanup as best-effort maintenance instead of allowing it to starve the bot's primary GitHub state loop.

## Validation

- `bash -n scripts/ec2-bot/scripts/clear-stale-locks.sh`

### DIFF
--- a/scripts/ec2-bot/scripts/clear-stale-locks.sh
+++ b/scripts/ec2-bot/scripts/clear-stale-locks.sh
@@ -25,6 +25,24 @@ BOT_DISK_PREFIX="${LOCK_PREFIX##*/}"
 
 mkdir -p "$HEARTBEAT_DIR" 2>/dev/null || true
 
+# GitHub-backed cleanup is best-effort maintenance. It must not be allowed to
+# starve the state scanner or review-request handling of GraphQL budget.
+GH_CLEANUP_MIN_GRAPHQL_REMAINING="${GH_CLEANUP_MIN_GRAPHQL_REMAINING:-1000}"
+GH_CLEANUP_CHECK_LIMIT="${GH_CLEANUP_CHECK_LIMIT:-25}"
+GH_CLEANUP_LOCK="/tmp/${BOT_DISK_PREFIX}-github-cleanup.lock"
+GH_CLEANUP_LOCK_HELD=false
+
+github_graphql_remaining() {
+  gh api rate_limit --jq '.resources.graphql.remaining // 0' 2>/dev/null || echo 0
+}
+
+cleanup_github_lock() {
+  if [ "$GH_CLEANUP_LOCK_HELD" = "true" ]; then
+    rmdir "$GH_CLEANUP_LOCK" 2>/dev/null || true
+  fi
+}
+trap cleanup_github_lock EXIT
+
 # Clean up any leftover rate limit state files from the removed rate limiting infrastructure
 rm -f "${LOCK_PREFIX}-rate-limit-state.json" "${LOCK_PREFIX}-github-rate-limit-state.json" 2>/dev/null || true
 rm -f "${LOCK_PREFIX}-rate-limited.flag" "${LOCK_PREFIX}-rate-limit-snapshot.txt" "${LOCK_PREFIX}-rate-limit-warning.flag" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Adds three layers of protection to prevent the bot's worktree/branch cleanup from exhausting the shared 5k-point/hour GraphQL budget: a pre-flight budget check (min 1000 remaining), a per-run cap of 25 `gh pr list` calls, and a `mkdir`-based mutex to prevent concurrent cleanup runs.

## Feedback

- **Good: budget guardrails** — The three-tier approach (mutex, budget floor, per-run cap) is well-designed and all thresholds are env-configurable. The `mkdir` atomic lock and 120-min stale-lock breaker are solid patterns.

- **Good: counter placement** — Incrementing `GH_CLEANUP_CHECKS` before the `gh pr list` call (not after) correctly counts attempts, since even failed calls consume budget.

- **Minor (non-blocking) — `exit 0` on low budget touches the marker first** (`clear-stale-locks.sh:318-322`): `touch "$WT_CLEANUP_MARKER"` runs before the budget check, so if GraphQL remaining is below 1000, cleanup is skipped but the marker means it won't be retried for another hour. If you want earlier retry when budget recovers, move `touch "$WT_CLEANUP_MARKER"` after the budget check (or only touch it after successful cleanup). Not blocking since hourly cadence is acceptable, just worth noting.

- **Minor (non-blocking) — indentation inconsistency** (`clear-stale-locks.sh:325-418`): The body of the `else` (lines 325–418) drops back to column 0 while it's logically inside the `mkdir` lock's else clause. Bash doesn't care, but it makes the nesting hard to follow. Consider re-indenting this block to match the surrounding structure in a future cleanup pass.

- **Nit — `exit 0` vs early return** (`clear-stale-locks.sh:322`): The `exit 0` on low budget exits the entire script. This is fine today since the cleanup block is at EOF, but could become a latent issue if more logic is appended after the `fi # end hourly worktree cleanup gate`. A comment like `# Safe to exit — nothing follows this block` would future-proof readability.

## Verdict

**Approve.** This is a well-motivated, focused fix for a real operational problem (cleanup starving the GraphQL budget). The minor items above are all non-blocking.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Adds GraphQL budget awareness and per-run caps to the worktree/branch cleanup in `clear-stale-locks.sh`, preventing the cleanup cron from starving the state scanner and review handlers of API budget.

## Feedback

- **Good pattern**: Using `mkdir` for atomic lock acquisition and an EXIT trap for cleanup — solid and race-free.
- **Good pattern**: Shared `GH_CLEANUP_CHECKS` counter across both the worktree loop and branch loop ensures a single combined cap per run.
- **Minor — `touch` before budget check** (`clear-stale-locks.sh:318`): `touch "$WT_CLEANUP_MARKER"` runs before the GraphQL budget check on line 320. If the budget is low and `exit 0` fires, the marker is already touched, so cleanup won't be retried for another hour even if the budget recovers sooner. This seems intentional (avoids re-checking every 5 minutes), but worth calling out explicitly — if the budget resets mid-hour, the stale worktrees will wait until the next hour boundary.
- **Cosmetic — inconsistent indentation** (`clear-stale-locks.sh:325-418`): The worktree/branch cleanup code is now nested inside the `else` block (line 316) but retains its original indent level. The closing `cleanup_github_lock` (line 420) is properly indented at 4 spaces, but the loop bodies aren't. Not a blocker — shell doesn't care — but it makes the nesting structure harder to scan.

## Verdict

**Approve** — the budget guard, per-run cap, and lock mechanism are all correct and address a real operational problem. The two notes above are non-blocking.

Issue #245, opened 2026-05-02
Bug: "Message not sent" error appearing repeatedly during active session


slam-paws:
## Bug Report

Darryl reports getting a "Message not sent — Failed to send message. Please try again." error multiple times while sending messages in an active MWF session.

The error appears as a red banner at the top of the chat screen. It occurred repeatedly during the same session.

<img src="https://github.com/shantamg/meet-without-fear/blob/temp-image-upload/.github/images/issue-message-not-sent-error.jpg?raw=true" alt="Message not sent error screenshot" width="300" />

## What We Know So Far

• Possibly related to #98 (network timeout errors during sessions), but that issue was from April 16 and involved timeouts across all endpoints — this appears to be a more specific message-send failure happening now
• Could be a transient backend/network issue or a client-side retry problem
• No immediate leads — needs full investigation

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Darryl Finkton Jr
- **Original message**: "I've gotten this multiple times while sending messages just now"
- **Timestamp**: 1777685142.980899

slam-paws:
## Investigation Report

**Root cause**: The AI model is emitting unrecognized "freelanced" dispatch tags (`CRITICAL_PRIVACY_BREACH`, `ESCALATE_PRIVACY_BREACH_CONCERN`, `CRITICAL_PRIVACY_CONCERN`) with **no user-facing response text** — all content is inside `<thinking>` and `<dispatch>` tags. When the dispatch handler encounters an unknown tag, it returns `null` and falls back to the AI's "original response," but there is no original response because the AI put everything inside tags. This triggers the empty-response guard at `messages.ts:1774` which throws an error, deletes the user's message, and surfaces "Message not sent" to the client.

**Evidence**:

- **Render logs**: 4 consecutive failures in session `cmog305d7008vq91wvl6m2b0l` between `01:17:04Z` and `01:24:20Z` on May 2, all in Stage 2B (empathy refinement), turn 19:
  1. `01:17:04` — Unknown dispatch tag (stream `u1eugj3pw`)
  2. `01:20:54` — `CRITICAL_PRIVACY_BREACH` (stream `9buppsczj`)
  3. `01:23:03` — `ESCALATE_PRIVACY_BREACH_CONCERN` (stream `ktn13tjcl`)
  4. `01:24:20` — `CRITICAL_PRIVACY_CONCERN` (stream `8iqbj1ddv`)

[...]

PR #246, opened 2026-05-02
fix: fallback response on empty AI after unknown dispatch tags

slam-paws:
## Changes
- Fix: Replace error-throw with warm fallback when AI emits all content inside unrecognized dispatch tags (e.g. `CRITICAL_PRIVACY_BREACH`), leaving the user-facing response empty after tag stripping
- Fix: Downgrade log level from `error` to `warn` for this expected edge case (LLM "freelancing" tags)
- Fix: Send fallback text via SSE chunk so the client renders it normally instead of showing "Message not sent"

Related to #245

## Provenance
- **Requested by:** Darryl Finkton Jr
- **Original message:** "I've gotten this multiple times while sending messages just now"

## Docs updated
No doc changes needed — error-handling-only fix, no API or stage logic changes.

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -1768,14 +1768,16 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
 
       // Guard: empty AI response after parsing + dispatch means the model emitted
       // content entirely inside tags we couldn't route (e.g. an unknown dispatch
-      // tag, or a stray <draft> with no chat text). Treat as a stream error so
-      // the user message is cleaned up and the client gets a retry prompt —
-      // better than persisting a confusing placeholder or a blank AI message.
+      // tag with no user-facing text). Instead of throwing (which deletes the
+      // user message and shows "Message not sent"), inject a warm fallback so
+      // the conversation continues smoothly.
       if (!accumulatedText.trim()) {
-        logger.error(`[sendMessageStream:${requestId}] Empty AI response after tag stripping and dispatch`, {
+        const fallback = 'I\'m here with you. Could you tell me more about what you\'re experiencing?';
+        logger.warn(`[sendMessageStream:${requestId}] Empty AI response after tag stripping — using fallback`, {
           dispatchTag: dispatchTag ?? null,
         });
-        throw new Error('Empty AI response after tag stripping and dispatch');
+        sendSSE(res, { event: 'chunk', data: { text: fallback } });
+        accumulatedText = fallback;
       }
 
       finalizeTurnMetrics(turnId);


slam-paws (commented):

Implementation matches requirements from #245 investigation report:
- [x] Replace error-throw with warm fallback when `!accumulatedText.trim()` — implemented with `'I'm here with you. Could you tell me more about what you're experiencing?'` and `accumulatedText = fallback` so downstream persistence works correctly
- [x] Downgrade log level from `error` to `warn` — implemented (`logger.error` → `logger.warn`)
- [x] Send fallback via SSE chunk so client renders it normally — implemented with `sendSSE(res, { event: 'chunk', data: { text: fallback } })`

No scope deviations detected. Single file changed (`backend/src/controllers/messages.ts`), exactly at the code path identified in the investigation. Ready for verification.

slam-paws:
## Bot review — LGTM ✅

**Correctness** ✅
- Replaces `throw new Error(...)` with a warm fallback + `accumulatedText = fallback`. Setting `accumulatedText` is the critical piece — it ensures downstream persistence (line ~1846 does `content: accumulatedText.trim()`) saves the fallback rather than a blank message. Correct.
- `sendSSE(res, { event: 'chunk', data: { text: fallback } })` sends the fallback to the client before the stream closes normally. Correct.
- `logger.warn` (down from `logger.error`) is appropriate — this is now a handled edge case, not a crash.

**Completeness** ✅
All three items from the investigation report (#245) are addressed:
- [x] Replace error-throw with warm fallback
- [x] Downgrade log level
- [x] Send fallback via SSE chunk

**Conventions** ✅
Follows existing patterns in `sendMessageStream`. Change is minimal and scoped.

**Security** ✅
Fallback is a hardcoded string constant — no user input, no injection risk.

**Doc drift** ✅ (false positive)
The Doc Impact Check flagged `docs/backend/api/stage-1.md` and `docs/product/stages/stage-1-witness.md`, but neither doc mentions the empty-response-after-tag-stripping behavior. This edge case is internal error handling, not a documented API or product contract. No doc updates needed.

**Tests**
`sendMessageStream` has no existing unit tests (SSE streaming is complex to mock). The new path is consistent with how this function is tested elsewhere (i.e., it isn't). Not blocking for this fix.

**Minor note** (not a blocker)
The fallback text is hardcoded and not stage-aware. `currentStage` is in scope and the reconciler has stage-specific fallback strings. Future hardening could call the reconciler fallback instead. Out of scope here.

---

Ready for human review. LGTM.

slam-paws:

shantamg:
On it — reviewing now.

shantamg (approved):
## Summary
Replaces a `throw` with a warm fallback when the AI emits all content inside unrecognized dispatch tags, leaving `accumulatedText` empty. Prevents the "Message not sent" banner + user-message deletion seen in #245.

## Feedback
- **Correctness**: Setting `accumulatedText = fallback` correctly flows into the downstream `prisma.message.create` at `messages.ts:1846`, so the fallback is persisted as a normal AI message and future turns see it in history. Good.
- **Scope**: Guard only fires after the dispatch branch already failed to produce a response (unknown tag → `isDispatchMessage = false` → empty text). Doesn't shadow the legitimate dispatch path. Good.
- **Logging**: Downgrade `error` → `warn` is appropriate now that this is recoverable. The `dispatchTag` is preserved in the log payload, so we can still track frequency in Render and decide later whether to harden the prompt.
- **Tests**: No test added for the fallback path. Low risk given the 7-line change, but a small unit test that asserts `accumulatedText` ends up as the fallback string (and no throw) when streaming yields only tag-wrapped content would prevent future regressions. Non-blocking.
- **UX nit**: Fallback copy is fine. If you want it stage-aware later (e.g., different prompt during Stage 2B empathy refinement vs Stage 0), easy follow-up.

## Verdict
Approve. Clean, minimal, addresses the root symptom without papering over the underlying AI behavior — the `warn` log keeps it observable.

PR #281, opened 2026-05-02
[codex] Improve feedback coach refinement context

shantamg:
## Summary
- Seed the feedback coach chat with the partner empathy statement before the user's rough note.
- Tighten the feedback-coach prompt so AI guidance is concise and explicitly offers send-as-is or keep-refining next steps.

## Why
When users refine feedback, they need the original message visible for reference, and the coach response should behave more like the existing share-refinement flow.

## Validation
- npm --workspace mobile run check
- npm --workspace backend test -- --runTestsByPath src/routes/__tests__/stage2.test.ts

### DIFF
--- a/backend/src/controllers/stage2.ts
+++ b/backend/src/controllers/stage2.ts
@@ -1439,8 +1439,10 @@ Your goal is to help them rephrase their feedback to be constructive, specific, 
 
 User's raw feedback: "${message}"
 
-1. Acknowledge the validity of their feeling.
-2. Draft a "Proposed Feedback" statement that they can send inside Meet Without Fear. This should be:
+1. Keep the conversational coaching response concise: 1-2 short sentences, no more than 45 words.
+2. Acknowledge the validity of their feeling without over-explaining or giving a long lesson.
+3. Explicitly tell them they can keep refining or send the proposed feedback as-is.
+4. Draft a "Proposed Feedback" statement that they can send inside Meet Without Fear. This should be:
    - Direct but kind.
    - Focus on what was missed or misunderstood.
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Tightens the feedback coach UX in two ways: (1) constrains the backend prompt to produce concise coaching responses (≤45 words) and explicitly tells the user they can refine or send as-is, and (2) surfaces the partner's empathy statement as a visible chat message so the user sees what they're responding to.

## Feedback

- **Good: Timestamp dedup** (`useRefinementChat.ts:288`): Capturing `openedAt = Date.now()` once and reusing it for all message IDs in `resetChat` prevents potential ID collisions from multiple `Date.now()` calls in the same tick. Nice cleanup.

- **Good: Prompt conciseness constraint** (`stage2.ts:1442`): The 45-word cap on the coaching response is a smart guardrail — the previous prompt had no length constraint which could lead to verbose, lecture-y responses that feel out of place in a chat flow.

- **Good: Source message** (`useRefinementChat.ts:290-299`): Showing the partner's empathy statement in the chat gives the user visible context for what they're refining against. Previously this was only sent to the backend but invisible in the UI.

- **Nit (non-blocking)**: The intro message copy (`useRefinementChat.ts:304`) says "your note" — the word "note" isn't used elsewhere in the feedback coach flow (it's "feedback" or "what felt off"). Minor, but "your feedback" might be more consistent.

## Verdict
**Approve** — Clean, well-scoped changes. Prompt improvements align with the conciseness goals, and the source-message addition gives needed context to the user. No logic, type, or security concerns.

PR #255, opened 2026-05-02
feat: update Stage3Gates and needs DTOs for Stage 3 redesign

slam-paws:
## Changes
- Replace `commonGroundConfirmed` with `needsValidated` and `needsShared` in `Stage3Gates` interface
- Remove `CommonGroundDTO`, `GetCommonGroundResponse`, `ConfirmCommonGroundRequest`, `ConfirmCommonGroundResponse`, `NeedsComparisonCommonGroundDTO`
- Add `CaptureNeedsRequest`, `CaptureNeedsResponse`, `CapturedNeedInput` for conversation-driven needs capture
- Add `ValidateNeedsResponse` for the new validity gate flow
- Remove `commonGroundReady` from `ConsentShareNeedsResponse`
- Simplify `GetNeedsComparisonResponse` to just myNeeds/partnerNeeds (no common ground)

Related to #248

## Provenance
- **Channel:** GitHub issue
- **Requested by:** bot (spec #247, wave 1)
- **Original message:** Stage 3 redesign shared types update
- **Prompt(s) used:** general-pr workspace

## Docs updated
No doc changes needed — shared type-only changes. Stage 3 docs update tracked in #254 (wave 5).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #291, opened 2026-05-02
Implement Stage 3 needs reveal validation flow

shantamg:
## Summary
- Follow-up to #255, which has already merged the shared DTO changes and temporary deprecated CommonGround DTO compatibility shim.
- Implements the Stage 3 consumer/behavior flow: capture/confirm needs, explicit share consent, side-by-side needs reveal, validate/not-valid-yet handling, and Stage 4 advancement once both users validate.
- Updates mobile Stage 3 UI and e2e helpers/tests to avoid AI-authored common-ground overlap in the active path while keeping legacy compatibility routes/types in place.

## Verification
- npm --workspace shared run check
- npm --workspace mobile run check
- npm --workspace e2e run check
- npm --workspace backend test -- --runTestsByPath src/routes/__tests__/stage3.test.ts --runInBand
- npm --workspace mobile test -- --runTestsByPath src/components/__tests__/NeedsDrawer.test.tsx --runInBand
- git diff --check
- npm --workspace e2e run e2e -- --project=two-browser-stage-3

## Manual / Production Testing Notes
- Local focused two-browser Stage 3 Playwright flow passed against backend + Expo web servers.
- Remaining production/manual gap: validate the same flow on deployed mobile/native surfaces with real auth/realtime infrastructure before broad rollout.

### DIFF
--- a/backend/src/controllers/sessions.ts
+++ b/backend/src/controllers/sessions.ts
@@ -638,17 +638,27 @@ export async function advanceStage(req: Request, res: Response): Promise<void> {
       return;
     }
 
-    // Special check for Stage 4: requires both users to complete Stage 3
+    // Special check for Stage 4: requires both users to satisfy Stage 3 gates.
+    // Stage 3 validation no longer auto-creates Stage 4 rows; each user advances explicitly.
     if (nextStage === 4) {
       const partnerId = await getPartnerUserId(sessionId, user.id);
       if (partnerId) {
-        // Check specifically if partner's Stage 3 is completed
-        // (not their highest stage, as they may already be in Stage 4)
+        const partnerCurrentProgress = session.stageProgress
+          .filter((sp) => sp.userId === partnerId)
+          .sort((a, b) => b.stage - a.stage)[0];
+
         const partnerStage3 = session.stageProgress.find(
           (sp) => sp.userId === partnerId && sp.stage === 3
         );
+        const partnerStage3Gates = checkGates(
+          3,
+          partnerStage3?.gatesSatisfied as Record<string, unknown> | null
+        );
 
-        if (!partnerStage3 || partnerStage3.status !== 'COMPLETED') {
+        if (
+          partnerCurrentProgress?.stage !== 4 &&
+          (!partnerStage3 || !partnerStage3Gates.satisfied)
+        ) {
           successResponse(res, {
             advanced: false,
             newStage: currentStage,
--- a/backend/src/controllers/stage3.ts
+++ b/backend/src/controllers/stage3.ts
@@ -19,7 +19,7 @@ import {
   acquireExtractionLock,
   releaseExtractionLock,
 } from '../services/needs';
-import { confirmNeedsRequestSchema, ConsentContentType } from '@meet-without-fear/shared';
+import { confirmNeedsRequestSchema, ConsentContentType, NeedCategory } from '@meet-without-fear/shared';
 import { notifyPartner, publishSessionEvent } from '../services/realtime';
 import { successResponse, errorResponse } from '../utils/response';
 import { getPartnerUserId } from '../utils/session';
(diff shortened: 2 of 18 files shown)


slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Replaces the AI-generated common-ground flow in Stage 3 with a simpler, human-driven needs reveal and validation model: capture → confirm → consent → side-by-side reveal → validate. The transition to Stage 4 now requires both partners to explicitly validate the revealed needs, rather than auto-advancing after common-ground confirmation. Good architectural direction — removes the fragile AI common-ground analysis step and gives users direct agency.

## Feedback

### Blocking

- **[Race condition]** `validateNeeds` in `backend/src/controllers/stage3.ts` — when both users call `/needs/validate` concurrently with `validated: true`, both requests can see `partnerValidated = true` and both enter the `canAdvance` block. This creates **duplicate transition messages** (4 instead of 2) and runs `stageProgress.updateMany` + `stageProgress.upsert` twice. The `upsert` prevents duplicate Stage 4 rows, but messages will be doubled. Consider: (a) a lock similar to `acquireExtractionLock`, or (b) wrapping in a Prisma `$transaction` with a unique constraint check, or (c) making the message creation idempotent.

- **[No transaction]** Same `validateNeeds` function — the entire Stage 3→4 transition (completing Stage 3 records, upserting Stage 4 records, creating transition messages, publishing events) is not wrapped in a `prisma.$transaction`. If any operation fails midway, the database is left in an inconsistent state (e.g., Stage 3 completed but Stage 4 not created). The removed `confirmCommonGround` code had the same issue, but since this is new code, it's worth fixing.

### Non-blocking

- **[Types]** Two `as any` casts for event names (`'session.needs_reveal_ready' as any` at `stage3.ts` and `'partner.needs_validated' as any`). These should be added to the shared event type union so the type system covers the new events end-to-end.

- **[Dead code]** `getCommonGround` still runs AI common-ground analysis (lines ~698-757 in stage3.ts). Since the new flow bypasses common ground entirely (`getNeedsComparison` now returns `commonGround: []`, `noOverlap: false`), the `getCommonGround` endpoint and its AI analysis are dead code in the new path. Consider removing or gating it behind the legacy route in a follow-up.

- **[Dead code]** In `useUnifiedSession.ts`, `commonGround` is hardcoded to `[]` and `commonGroundComplete` to `false`, but the screen still renders a `commonGroundData?.noOverlap` branch (UnifiedSessionScreen.tsx ~line 2109). Since `noOverlap` is now always `false`, that entire conditional branch is unreachable dead code.

- **[Hardcoded value]** `captureNeeds` in `stage3.ts` sets `aiConfidence: 0.85` for all captured needs. If this is intentional (user-confirmed from a summary card), a comment explaining the choice would help. If it should reflect actual extraction confidence, it should be passed from the client.

- **[Input validation]** `captureNeeds` validates each `need` object individually but returns on the first invalid item. For a batch endpoint, consider collecting all validation errors and returning them together so the client can fix all issues at once.

- **[Data preference]** In `captureNeeds`, `need: item.description || item.need!` — the field uses `description` if present, falling back to `need`. The `!` non-null assertion on `item.need` is safe because of the earlier validation check, but swapping `description` over `need` for the stored value is subtle. A comment clarifying the preference would help future readers.

## Good patterns

- Separating confirm from consent is a much better UX — users explicitly choose to share rather than auto-consenting.
- The `NeedsDrawer` refactor to extract `renderSideBySideNeeds` and reuse it across modes is clean.
- New `NeedsDrawer.test.tsx` unit tests are well-structured and cover the key rendering modes and interactions.
- E2E test helpers (`confirmNeedsSummaryAndConsent`, `waitForNeedsReveal`, etc.) are well-abstracted and will make future Stage 3 test updates easier.
- The `chatUIState` change to keep the needs review panel visible after confirm (until share) correctly handles the two-phase flow.

## Verdict

**Comment** — already merged. The race condition in `validateNeeds` should be addressed in a follow-up PR to prevent duplicate transition messages under concurrent validation. The missing transaction is lower priority but worth bundling.

PR #322, opened 2026-05-04
[codex] Fix invitation chat flow and session deletion

shantamg:
## Summary
- Move invitee opening context and topic into the chat stream and remove the extra topic-ready step after compact signing.
- Remove the `Compact Signed` timeline indicator and show the topic in the activity drawer for both inviter and invitee.
- Fix web session deletion so delete controls do not navigate, persist deleted sessions with per-user archival, and prevent duplicate chat keys from repeated message ids.

## Root Cause
Invitee onboarding context was rendered as transient stage UI instead of chronological chat content. Session deletion hid rows locally but the backend still listed sessions through relationship membership. The chat list also trusted paged message data without a final id dedupe.

## Validation
- `npm run check` in `mobile`
- `npm run check` in `backend`
- `npm test -- --runInBand src/services/__tests__/stage-prompts.test.ts` in `backend`
- `npx prisma migrate deploy` and `npx prisma migrate status` locally

### DIFF
--- a/backend/src/services/session-deletion.ts
+++ b/backend/src/services/session-deletion.ts
@@ -109,8 +109,30 @@ export async function deleteSessionForUser(
     where: { sessionId, userId },
   });
   if (userVessel) {
-    await prisma.userVessel.delete({
+    await prisma.userEvent.deleteMany({ where: { vesselId: userVessel.id } });
+    await prisma.emotionalReading.deleteMany({ where: { vesselId: userVessel.id } });
+    await prisma.identifiedNeed.deleteMany({ where: { vesselId: userVessel.id } });
+    await prisma.boundary.deleteMany({ where: { vesselId: userVessel.id } });
+    await prisma.userDocument.deleteMany({ where: { vesselId: userVessel.id } });
+    await prisma.userVessel.update({
       where: { id: userVessel.id },
+      data: {
+        archivedAt: new Date(),
+        conversationSummary: null,
+        notableFacts: undefined,
+        lastViewedAt: null,
+        lastSeenChatItemId: null,
+        lastViewedShareTabAt: null,
+      },
+    });
+    dataRecordsDeleted += 1;
+  } else {
+    await prisma.userVessel.create({
+      data: {
+        sessionId,
+        userId,
+        archivedAt: new Date(),
+      },
     });
     dataRecordsDeleted += 1;
   }
--- a/backend/prisma/migrations/20260504120000_add_user_vessel_archived_at/migration.sql
+++ b/backend/prisma/migrations/20260504120000_add_user_vessel_archived_at/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "UserVessel" ADD COLUMN "archivedAt" TIMESTAMP(3);
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -359,6 +359,9 @@ model UserVessel {
   // When the user last viewed the Share/Partner tab
   // Used to determine "seen" status for shared content (empathy, context)
   lastViewedShareTabAt DateTime?
+  // When this user removed the session from their own conversation list.
+  // The partner may still retain their own vessel and access.
+  archivedAt           DateTime?
 
   // === Notable Facts ===
   // AI-extracted structured facts about the user's situation, emotions, and circumstances
--- a/backend/src/controllers/invitations.ts
+++ b/backend/src/controllers/invitations.ts
@@ -64,6 +64,12 @@ export async function listSessions(req: Request, res: Response): Promise<void> {
             some: { userId: user.id },
           },
         },
+        userVessels: {
+          some: {
+            userId: user.id,
+            archivedAt: null,
+          },
+        },
         ...(status && { status: status as 'INVITED' | 'ACTIVE' | 'PAUSED' | 'RESOLVED' | 'ABANDONED' }),
       },
       include: {
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -81,7 +81,7 @@ function getFallbackInitialMessage(
   const partner = partnerName || 'them';
 
   if (isInvitee) {
-    return `Hey ${userName}, thanks for accepting ${partner}'s invitation to talk. What's been on your mind about things with ${partner}?`;
+    return `I'd like to hear your side now. What's been happening from your point of view with ${partner}?`;
   }
 
   if (isInvitationPhase) {
@@ -926,7 +926,7 @@ export async function getInitialMessage(
     } else {
       prompt = buildInitialMessagePrompt(
         currentStage,
-        { userName, partnerName, isInvitee, innerThoughtsContext },
+        { userName, partnerName, isInvitee, topicFrame: session.topicFrame, innerThoughtsContext },
         isInvitationPhase
       );
     }
--- a/backend/src/controllers/session-state.ts
+++ b/backend/src/controllers/session-state.ts
@@ -118,6 +118,7 @@ export async function getSessionState(req: Request, res: Response): Promise<void
           },
         },
         select: {
+          lastViewedAt: true,
           lastSeenChatItemId: true,
         },
       }),
@@ -290,6 +291,7 @@ export async function getSessionState(req: Request, res: Response): Promise<void
         },
         createdAt: session.createdAt.toISOString(),
         resolvedAt: session.resolvedAt?.toISOString() ?? null,
+        lastViewedAt: userVessel?.lastViewedAt?.toISOString() ?? null,
         lastSeenChatItemId: userVessel?.lastSeenChatItemId ?? null,
       },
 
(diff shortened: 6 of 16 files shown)


slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Two-part PR: (1) soft-archive session deletion instead of hard-deleting UserVessel, and (2) replaces the blocking invitee "Ready" gate with an inline animated topic-frame card in the chat flow. Also adds web platform delete UX, swipe-press protection, message deduplication, and animation boundary improvements via `lastViewedAt`.

## Feedback

- **[Bug – Data wipe]** `backend/src/services/session-deletion.ts` — `notableFacts: undefined` in the Prisma update is a no-op. In Prisma, `undefined` means "skip this field", so notable facts are never cleared on archive. Should be `notableFacts: null` (or `Prisma.DbNull`) to actually set the column to NULL.

- **[Correctness – Filter]** `backend/src/controllers/invitations.ts:67-72` — The new `userVessels: { some: { userId, archivedAt: null } }` filter assumes every visible session has a UserVessel for the current user. This is true after acceptance (vessel is created at accept time), but if both users are already relationship members from a prior session and the partner creates a new session, the invitee won't see it in `listSessions` until they accept (no vessel exists yet). If that's the intended behavior (invitee sees it via invitation link only), worth a brief comment. Alternatively, the filter could use `NOT: { userVessels: { some: { userId, archivedAt: { not: null } } } }` which is "exclude sessions where vessel IS archived" rather than "require a non-archived vessel".

- **[Cleanup]** `mobile/src/screens/UnifiedSessionScreen.tsx:1045` — `inviteeTopicAckPending = false` is dead code. The `waitingStatus` destructuring below it and the `'invitee-topic-ack'` case in `renderAboveInput` reference it but will never trigger. Could remove these artifacts entirely for clarity.

- **[Performance – minor]** `mobile/src/components/ChatInterface.tsx` — `listItems.findIndex(...)` inside `renderItem` (called per rendered message) is O(n) per item. Not blocking since chat lists are typically small, but a `Map<id, index>` memo would be O(1).

- **[Good]** Message deduplication in `useUnifiedSession.ts` is solid defensive coding against paginated query overlap. The test for the invitee opener prompt covers the important constraints well. Web delete button with accessibility labels is well done.

## Verdict

Comment — no blocking issues, but the `notableFacts: undefined` bug means archived vessels retain personal data, which seems contrary to the deletion intent. The filter approach in `listSessions` is worth a quick sanity check. Overall architecture is clean and the UX improvements are well-considered.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Large PR that reworks the invitee chat-entry flow and session deletion. Key changes: (1) invitee topic-acknowledgement moves from a gated screen into an animated chat card, (2) session deletion soft-archives UserVessels instead of hard-deleting them so sessions hide from the deleting user's list while the partner keeps theirs, (3) web-platform support for session list deletion, (4) swipe-to-press blocking to prevent accidental navigation after swipe gestures, (5) message deduplication across paginated pages, and (6) lastViewedAt-based animation boundaries.

## Feedback

### Blocking

- **[Correctness / Privacy]** `backend/src/services/session-deletion.ts` — `notableFacts: undefined` is a no-op in Prisma; `undefined` means "skip this field" so the user's notable facts are **never cleared** on deletion. This is a data-privacy concern. The codebase uses `Prisma.DbNull` elsewhere for the same purpose (see `data-retention.ts:56-59`). Fix:
  ```ts
  notableFacts: Prisma.DbNull,
  ```

### Non-blocking

- **[Cleanup]** `mobile/src/utils/chatListSelector.ts:76-101` — The `compact` and `sessionCreatedAt` fields on `SessionIndicatorData` are now unused (compact-signed indicator removed, destructuring updated at line 129). Consider removing them from the interface to avoid confusion.

- **[Cleanup]** `mobile/src/screens/UnifiedSessionScreen.tsx` — `inviteeTopicAckPending` is now hardcoded to `false` and the `'invitee-topic-ack'` case in `renderAboveInput` was removed. The variable and its downstream consumers (the waiting-status hook that checks it) could be removed entirely.

- **[Performance]** `mobile/src/components/ChatInterface.tsx` — `renderItem` calls `listItems.findIndex()` for each item, making each render pass O(n²). For typical chat sizes this is fine, but if chats grow large, consider building an id→index map in a `useMemo` and looking up from there.

- **[Style]** `mobile/src/components/SessionDrawer/index.tsx:1070` — `webDeleteButton` uses a hardcoded `'#E53935'` while the sessions screen version at `sessions.tsx:513` uses `t.colors.error`. Would be nice to be consistent.

### Good stuff

- Solid approach converting the invitee topic-ack from a blocking gate to an animated chat card — much better UX flow.
- Message deduplication in `useUnifiedSession` is a good defensive fix.
- The swipe-press-blocking logic is thorough and handles edge cases well (pointer tracking, time-based blocking, open-state tracking).
- New test for the invitee opener prompt is well-structured.
- Web platform `confirm()` fallback for `Alert.alert` is a nice touch.

## Verdict

**Request changes** — the `notableFacts: undefined` bug is a data-privacy issue that should be fixed before merge. The rest are non-blocking suggestions.

backend/src/services/session-deletion.ts
```diff
@@ -109,8 +109,30 @@ export async function deleteSessionForUser(
...
+        conversationSummary: null,
+        notableFacts: undefined,
+        lastViewedAt: null,
+        lastSeenChatItemId: null,
+        lastViewedShareTabAt: null,
+      },
+    });
+    dataRecordsDeleted += 1;
+  } else {
+    await prisma.userVessel.create({
```
slam-paws, line 130: `notableFacts: undefined` is a no-op in Prisma — `undefined` means "skip this field", so notable facts are never cleared on deletion. Should be `notableFacts: Prisma.DbNull` (consistent with `data-retention.ts:56-59`).

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

This PR refactors the invitee onboarding flow so the topic frame appears as an animated inline chat card (instead of a separate "Ready" gate), changes session deletion from hard-delete to soft-archive via `archivedAt`, adds `lastViewedAt` to the session state for animation boundaries, adds web-platform support for sessions lists (visible delete button instead of swipe), fixes swipe-to-press interference on native, and deduplicates paginated messages. Solid set of changes overall — the invitee topic card with typewriter animation is a nice UX improvement.

## Feedback

### Blocking

- **[Bug] `notableFacts: undefined` doesn't clear the field** — `backend/src/services/session-deletion.ts:130` sets `notableFacts: undefined`. In Prisma, `undefined` means "don't update this field" — so notable facts survive the archive and the user's data isn't actually scrubbed. This should be `notableFacts: null` (the field is `Json?` in the schema, so `null` is valid). Compare with `conversationSummary: null` on line 129 which correctly clears.

### Important (non-blocking)

- **[Reliability] No transaction for multi-step deletion** — `session-deletion.ts:119-143` runs 5 `deleteMany` calls plus an `update` (or `create`) sequentially without a `prisma.$transaction()`. If any operation fails mid-way, you'll have partially deleted data with no rollback. Consider wrapping the block in `prisma.$transaction(async (tx) => { ... })`.

- **[Cleanup] `inviteeTopicAcked` parameter still accepted** — `useUnifiedSession` (line ~201) still declares `options?: { inviteeTopicAcked?: boolean }` and uses it internally, but the only caller no longer passes it (defaults to `true`, making the gating logic a no-op). This dead parameter and its downstream conditionals could be removed to avoid confusion.

- **[Cleanup] `SessionIndicatorData` still has `compact` and `sessionCreatedAt`** — The diff removes their usage from `deriveIndicators` but the interface definition still includes these fields. Callers that construct the data object no longer pass them (the diff removes them from `UnifiedSessionScreen`). TypeScript won't error since they're optional, but the stale type definition is misleading.

### Minor

- **[Style] Inconsistent delete button color** — `SessionDrawer/index.tsx` hardcodes `backgroundColor: '#E53935'` while `sessions.tsx` uses `t.colors.error`. Should use the theme token consistently.

- **[Cleanup] `inviteeTopicAckPending = false` is dead code** — Hardcoded to `false` and threaded through the UI state pipeline. The entire `'invitee-topic-ack'` case in `renderAboveInput` was removed, so this variable and its downstream usage in `useChatUIState` could be cleaned up.

- **[Perf] `listItems.findIndex` in renderItem** — `ChatInterface.tsx` calls `listItems.findIndex()` inside `renderItem` (lines ~710, 745). This is O(n) per rendered item. Not a problem for typical session sizes but worth noting — if performance becomes an issue, a pre-computed `Map<id, index>` would help.

## Verdict

**Request changes** — the `notableFacts: undefined` bug means user data isn't properly scrubbed on session deletion, which is a privacy concern. The fix is a one-character change (`undefined` → `null`). The transaction wrapping is also worth addressing for reliability. Everything else is minor cleanup.

backend/src/services/session-deletion.ts
slam-paws, line 130: Bug: \`undefined\` in Prisma means "skip this field" — notableFacts won't be cleared. Should be \`notableFacts: null\` (same as \`conversationSummary\` on the line above).

PR #327, opened 2026-05-04
[codex] Fix feel-heard panel and failed stream cleanup

shantamg:
## Summary
- Stop confirmed invitations from reserving the above-input slot and hiding the Stage 1 feel-heard panel.
- Roll back optimistic user/AI placeholder messages when an SSE stream fails or times out, so failed sends do not leave indefinite ghost typing dots.
- Keep the rebased auth fast path stable by using the Prisma runtime error class and adding coverage for Clerk profile lookup failure paths.
- Add narrow TypeScript annotations for rebased mainline files that blocked backend auth test compilation.

## Root Cause
The feel-heard offer was correctly persisted as `feelHeardCheckOffered: true`, but the mobile panel priority selector still chose `invitation` for inviters with a confirmed topic. That `invitation` case now renders nothing because the invitation flow moved to a modal, so the invisible invitation slot blocked the visible feel-heard panel.

The later failed send was a separate issue: the live backend returned 500 during auth before the stream controller ran, and the mobile stream error handler only set error state. Because the optimistic user message remained in React Query, `ChatInterface` still saw the last message as `USER` and kept rendering the typing indicator.

## Validation
- `npm --workspace=@meet-without-fear/backend test -- auth.test.ts --runInBand`
- `npm --workspace=@meet-without-fear/mobile test -- chatUIState --runInBand`
- `npm --workspace=@meet-without-fear/mobile run check`

### DIFF
--- a/backend/src/middleware/__tests__/auth.test.ts
+++ b/backend/src/middleware/__tests__/auth.test.ts
@@ -198,7 +198,7 @@ describe('Auth Middleware', () => {
     });
 
     it('handles concurrent create race (P2002 fallback)', async () => {
-      const { Prisma } = jest.requireActual('@prisma/client');
+      const { PrismaClientKnownRequestError } = jest.requireActual('@prisma/client/runtime/library');
       mockVerifyToken.mockResolvedValue({ sub: 'clerk-race-user' });
       mockGetUser.mockResolvedValue({
         firstName: 'Race',
@@ -210,7 +210,7 @@ describe('Auth Middleware', () => {
       (prisma.user.findUnique as jest.Mock)
         .mockResolvedValueOnce(null)
         .mockResolvedValueOnce(raceUser);
-      const p2002Error = new Prisma.PrismaClientKnownRequestError('Unique constraint failed', { code: 'P2002', clientVersion: '5.0.0' });
+      const p2002Error = new PrismaClientKnownRequestError('Unique constraint failed', { code: 'P2002', clientVersion: '5.0.0' });
       (prisma.user.create as jest.Mock).mockRejectedValue(p2002Error);
 
       const req = createMockRequest({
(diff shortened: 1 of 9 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes the feel-heard panel being blocked after invitation confirmation. The invitation panel was holding the above-input slot even after the invite was sent because the only exit conditions were explicit dismissal ("Later") or in-flight mutation — not successful confirmation.

## Feedback

- **Correctness**: The fix is spot-on. `computeShowInvitationPanel` now correctly gates on `!invitationConfirmed`, so once the invite is sent, the panel yields the slot to lower-priority panels (feel-heard at Priority 3). The `invitationConfirmed` value is already derived in `useUnifiedSession.ts:504` (`invitation?.messageConfirmed ?? false`) and returned from the hook, so the screen just passes it through.

- **Tests**: Good coverage — the new test explicitly verifies that a confirmed invitation doesn't block feel-heard, the existing priority test is updated to be explicit about `invitationConfirmed: false`, and the default in `createDefaultChatUIStateInputs()` ensures backward compatibility for tests that don't set the field.

- **Default value**: `invitationConfirmed: false` in `createDefaultChatUIStateInputs()` is the correct safe default — panels that depended on the old behavior (no gating) will continue to show the invitation panel until confirmation occurs.

## Verdict

Approve — clean fix, good test coverage, no concerns.

slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary
Fixes the feel-heard panel being blocked by the invitation panel after an invitation is already confirmed. Adds `invitationConfirmed` as a gate to `computeShowInvitationPanel` so the slot is freed for subsequent panels.

## Feedback
- **Correctness**: Logic is clean — the `!invitationConfirmed` condition correctly prevents the invitation panel from reserving the above-input slot once no longer needed.
- **Tests**: Good regression test added (`does not let a confirmed invitation block feel-heard`) and existing tests properly updated with the new field.
- **Types/Defaults**: `invitationConfirmed: boolean` properly typed in the interface and defaulted to `false` in `createDefaultChatUIStateInputs()`.

## Verdict
Approve — minimal, well-tested fix that solves a clear UI priority bug.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the feel-heard confirmation panel being blocked by a stale invitation panel after invitation is already confirmed. Also improves streaming error recovery by cleaning up optimistic messages on failure, and adds Clerk resilience + Prisma import fixes on the backend.

## Feedback
- **Correctness (mobile)**: The core fix is clean — adding `invitationConfirmed` as a gate on `computeShowInvitationPanel` correctly prevents it from holding the above-input slot after the invitation is sent. Good test coverage with the new "does not let a confirmed invitation block feel-heard" case.

- **Correctness (streaming)**: `cleanupFailedStream` properly removes optimistic messages before invalidating queries, avoiding the stale ghost-message problem. The `removeMessagesFromCache` helper correctly handles both flat and infinite query caches with Set-based deduplication.

- **Minor observation** (`useStreamingMessage.ts:549`): The `activeUserMessageIdRef.current = optimisticUserIdRef.current` assignment inside the `user_message_created` handler is a no-op in current code since both refs already hold the same value at that point (set together at line 453–454). Not harmful (Set dedup covers it in cleanup), but could be confusing to future readers. Non-blocking.

- **Backend**: Direct import of `PrismaClientKnownRequestError` from `@prisma/client/runtime/library` is the correct Prisma 5.x pattern. The inline `{ userId: string }` type annotations in `realtime.ts` and `account-deletion.ts` fix implicit `any` without over-engineering.

- **Tests**: New auth tests for Clerk unavailability provide good coverage for an existing graceful degradation path. The P2002 test fix matches the new import style.

## Verdict
Approve — well-scoped bug fix with solid test coverage and a nice streaming resilience improvement.

shantamg:
Superseded by #328, which is a fresh branch directly off origin/main with the same fix scope and clean history.

PR #328, opened 2026-05-04
[codex] Fix feel-heard panel and failed stream cleanup

shantamg:
## Summary
- Stop confirmed invitations from reserving the above-input slot and hiding the Stage 1 feel-heard panel.
- Roll back optimistic user/AI placeholder messages when an SSE stream fails or times out, so failed sends do not leave indefinite ghost typing dots.
- Make Stage 2 empathy consent idempotent when a share retry/double-click arrives after an empathy attempt already exists.
- Keep auth/stage2 test compatibility stable on current main with narrow Prisma enum/runtime imports and explicit callback types.

## Root Cause
The feel-heard offer was correctly persisted as `feelHeardCheckOffered: true`, but the mobile panel priority selector still chose `invitation` for inviters with a confirmed topic. That `invitation` case now renders nothing because the invitation flow moved to a modal, so the invisible invitation slot blocked the visible feel-heard panel.

The failed send was separate: backend auth returned 500 before the stream controller ran, and the mobile stream error handler only set error state. Because the optimistic user message remained in React Query, `ChatInterface` still saw the last message as `USER` and kept rendering the typing indicator.

The empathy consent 409 was another retry/idempotency bug: the first share had succeeded and created a `HELD` empathy attempt, but duplicate share clicks retried `POST /empathy/consent`. The endpoint created another consent record before hitting the unique empathy-attempt constraint, then returned 409. It now detects existing attempts and returns the existing shared message as success.

## Validation
- `npm --workspace=@meet-without-fear/backend test -- stage2.test.ts --runInBand`
- `npm --workspace=@meet-without-fear/backend test -- auth.test.ts --runInBand`
- `npm --workspace=@meet-without-fear/mobile test -- chatUIState --runInBand`
- `npm --workspace=@meet-without-fear/mobile run check`

### DIFF
--- a/backend/src/controllers/stage2.ts
+++ b/backend/src/controllers/stage2.ts
@@ -12,8 +12,8 @@
 import { Request, Response } from 'express';
 import { logger } from '../lib/logger';
 import { prisma } from '../lib/prisma';
-import { EmpathyStatus } from '@prisma/client';
 import {
+  EmpathyStatus,
   MessageRole,
   saveEmpathyDraftRequestSchema,
   consentToShareRequestSchema,
@@ -603,6 +603,69 @@ export async function consentToShare(
       return;
     }
 
+    // Get partner ID from session data
+    const partnerId = getPartnerUserIdFromSession(session, user.id);
+
+    const respondWithExistingAttempt = async (existingAttempt: {
+      id: string;
+      content: string;
+      sharedAt: Date;
+      status: EmpathyStatus;
+    }) => {
+      const [partnerAttempt, existingEmpathyMessage] = await Promise.all([
+        partnerId
+          ? prisma.empathyAttempt.findFirst({
+            where: {
+              sessionId,
+              sourceUserId: partnerId,
+            },
+          })
+          : Promise.resolve(null),
+        prisma.message.findFirst({
+          where: {
+            sessionId,
+            senderId: user.id,
+            forUserId: user.id,
+            role: 'EMPATHY_STATEMENT',
+            stage: 2,
+          },
+          orderBy: { timestamp: 'desc' },
+        }),
+      ]);
+
+      successResponse(res, {
+        consented: true,
+        consentedAt: existingAttempt.sharedAt.toISOString(),
+        partnerConsented: !!partnerAttempt,
+        canReveal: !!partnerAttempt && consent,
+        status: existingAttempt.status,
+        empathyMessage: {
+          id: existingEmpathyMessage?.id ?? existingAttempt.id,
+          content: existingEmpathyMessage?.content ?? existingAttempt.content,
+          timestamp: (existingEmpathyMessage?.timestamp ?? existingAttempt.sharedAt).toISOString(),
+          stage: existingEmpathyMessage?.stage ?? 2,
+        },
+        transitionMessage: null,
+      });
+    };
+
+    // Idempotency: if this user already shared an empathy attempt, treat a
+    // retry/double-click as success instead of creating duplicate consent
+    // records and returning a conflict.
+    const existingAttempt = await prisma.empathyAttempt.findFirst({
+      where: {
+        sessionId,
+        sourceUserId: user.id,
+        status: { not: 'REFINING' },
+      },
+      orderBy: { sharedAt: 'desc' },
+    });
+
+    if (existingAttempt) {
+      await respondWithExistingAttempt(existingAttempt);
+      return;
+    }
+
     // Create consent record
     await prisma.consentRecord.create({
       data: {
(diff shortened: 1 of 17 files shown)


slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Fixes two user-facing bugs: (1) the feel-heard confirmation panel was hidden behind the invitation panel even after the invitation was already sent, and (2) failed SSE streams left orphaned optimistic messages in cache. Also adds idempotency to the `consentToShare` endpoint (retries return success instead of 409) and bulk type-annotation fixes for stricter TS compilation.

## Feedback

### Looks Good

- **Feel-heard panel fix** (`mobile/src/utils/chatUIState.ts:238`): Adding the `!invitationConfirmed` gate is the correct fix — once the invitation is sent, the panel shouldn't hold the `aboveInputPanel` slot. Test coverage is solid.
- **Stream cleanup** (`mobile/src/hooks/useStreamingMessage.ts`): `cleanupFailedStream` properly removes optimistic messages *before* invalidating queries, respecting the cache-first rule. Clean separation into `removeMessagesFromCache` + `cleanupFailedStream`.
- **Consent idempotency** (`backend/src/controllers/stage2.ts:668-680`): The early-return for existing attempts + the improved P2002 fallback is a good resilience pattern. Test covers the happy path well.
- **Import consolidation**: Moving `EmpathyStatus` and `BrainActivityCallType` to `@meet-without-fear/shared` / `../lib/bedrock` reduces coupling to Prisma internals.

### Minor Notes (Non-blocking)

- **`tx: typeof prisma`** (`backend/src/services/reconciler/sharing.ts:689`, `state.ts:77`): Prisma's `$transaction` callback receives a `TransactionClient`, not a full `PrismaClient`. Using `typeof prisma` works because the transaction client is a structural subset, but it's semantically inaccurate — you could accidentally reference `$transaction` or `$disconnect` on `tx` without a type error. Consider `Parameters<Parameters<typeof prisma.$transaction>[0]>[0]` or simply omitting the annotation (TS infers it correctly). Not blocking since it works at runtime.

- **Consent idempotency uses request's `consent` value** (`stage2.ts` line ~53 in diff): `canReveal: !!partnerAttempt && consent` uses the *current* request's `consent` boolean, not the original consent decision stored in the existing attempt. For true idempotency, consider reflecting the persisted state. Minor since retries will have the same value in practice.

### Blocker

- **CI is failing** — "Checks & Tests" shows FAILURE. The code changes look correct, but this needs a green build before merge.

## Verdict

**Comment** — Code quality is good and the fixes are well-motivated. Once CI passes, this is ready to merge. Please check the failing test/check output and fix.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Two user-facing fixes — a feel-heard panel that was permanently blocked by the invitation panel, and ghost messages left in cache after failed streams — plus an idempotent consent-to-share endpoint and a batch of TypeScript strict-mode type annotations.

## Feedback

- **Correctness (feel-heard fix)**: Adding `invitationConfirmed` as a gate in `computeShowInvitationPanel` is the right fix. Once the invitation is sent, it should release the `aboveInputPanel` slot for downstream panels. Test coverage is solid — the new `"does not let a confirmed invitation block feel-heard"` test directly validates the reported bug.

- **Correctness (stream cleanup)**: The `cleanupFailedStream` / `removeMessagesFromCache` pattern is well-structured — removes optimistic messages from cache *before* invalidating queries, preventing visual glitches. The `activeUserMessageIdRef` solves the problem where `optimisticUserIdRef` gets cleared after `user_message` is received, but cleanup still needs the ID. Good defensive design.

- **Idempotency (consentToShare)**: The pre-check for existing attempts and the P2002 recovery both returning the existing attempt via `respondWithExistingAttempt` is cleaner than the previous 409 behavior. One minor observation: if the request has `consent: false` but an attempt already exists, the idempotency guard still returns success. In practice this is fine since the UI only sends `consent: true` to this endpoint, but worth noting.

- **Types**: The explicit type annotations on Prisma query result callbacks are pragmatic — they solve the implicit `any` issue when Prisma's inferred types don't flow through closures. The local types (`SummaryMessageRecord`, `StageOneMessageForThemes`, `ConversationPromptMessage`, etc.) are appropriately scoped.

- **Import migration**: Moving `EmpathyStatus` and `BrainActivityCallType` from `@prisma/client` to `@meet-without-fear/shared` (and `PrismaClientKnownRequestError` to the runtime library path) is clean — keeps the shared package as the single source of truth for domain types.

- **Auth tests**: The new Clerk failure tests are meaningful — they confirm existing users bypass the Clerk API entirely (resilient to Clerk outages), and that new users surface errors properly when Clerk is unreachable.

## Verdict

Approve — well-scoped fixes with good test coverage. The feel-heard panel fix is a clear correctness improvement, the stream cleanup prevents a real UX issue (ghost messages), and the idempotent consent prevents duplicate records from double-taps.

PR #342, opened 2026-05-04
[codex] Fix James/Catherine gold session gates

shantamg:
## Summary

This PR implements the James/Catherine gold-session reliability pass from `docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md`.

The internal reviewer/reconciler wording concern is intentionally deferred to a separate product-copy pass. This PR only changes copy where needed to avoid planner-text leaks and forced "together"/repair assumptions.

## Bug List Covered

- E2E setup/cache drift: explicit E2E URL params now override cached identity in the web runtime, and the handoff doc records that no-Clerk gold sessions should use the `8082` E2E bundle.
- Stage 2 -> Stage 3 realtime isolation: stage completion events no longer broadcast one representative transition message; clients only insert messages addressed to the current user and otherwise refetch state/messages/progress.
- Stage 3 reveal gates: needs comparison now depends on both users having confirmed/consented needs, and the chat-side "I'm ready" path follows the same backend gate contract as the CTA flow.
- Stage 3 UI/input state: added explicit review, share, waiting-for-partner, and reveal-validation states; chat input hides when the valid next action is a gate CTA or waiting.
- Prompt/planner safety: tightened Stage 3 instructions, added visible output scrubbing for planner phrases, removed forced "way forward together" framing, and constrained James share suggestions to preserve accountability and safety.
- Refinement control messages: drawer refinement no longer injects `Refine empathy draft:` into the normal chat timeline/history.
- Streaming cleanup: terminal stream paths invalidate Stage 3 state and remove empty placeholders more consistently.
- Activity drawer UX: current gate drawers close/override the exchange-history drawer, and the drawer close action is now a full-width accessible target.

## Reviewer Ask

Slam Paws: please review this specifically against the bug list in the handoff doc and verify that the issues found in the James/Catherine run are actually fixed, not just covered by optimistic UI changes. In particular, please check the live Stage 2 -> Stage 3 transition, Catherine not seeing James-addressed prompts, Stage 3 reveal gates, refinement-history leakage, and no-agreement copy posture.

## Validation

- `npm --workspace backend test -- stage3 stage2 --runInBand`
- `npm --workspace mobile test -- chatUIState getWaitingStatus --runInBand`
- `git diff --check`

Not run in this pass: full two-browser James/Catherine E2E playthrough.

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -109,7 +109,7 @@ const PLANNER_LINE_PREFIXES = [
   'i need to make sure both lists',
 ];
 
-function scrubVisibleAIText(text: string): { text: string; scrubbed: boolean } {
+export function scrubVisibleAIText(text: string): { text: string; scrubbed: boolean } {
   const before = text;
   const cleaned = text
     .split(/\r?\n/)
@@ -124,7 +124,7 @@ function scrubVisibleAIText(text: string): { text: string; scrubbed: boolean } {
   return { text: cleaned, scrubbed: cleaned !== before };
 }
 
-function isReadyForStage3RevealText(content: string): boolean {
+export function isReadyForStage3RevealText(content: string): boolean {
   const normalized = content.trim().toLowerCase();
   if (/\bnot\s+ready\b/.test(normalized) || /\b(?:don'?t|do not)\s+(?:want|feel ready)\b/.test(normalized)) {
     return false;
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -1533,7 +1533,7 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
     }, { isInvitationPhase });
 
     // Prompt already includes semantic tag format instructions via buildResponseProtocol()
-    // No tool use instruction needed - we parse <thinking>, <draft>, <dispatch> tags instead
+    // No tool use instruction needed - we parse <thinking>, <draft>, <dispatch>, and <needs> tags instead
 
     // Format context bundle and inject into last user message (includes notable facts)
     const formattedContext = formatContextForPrompt(contextBundle, {
@@ -1588,11 +1588,11 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
     let streamError: Error | null = null;
 
     // Tag trap state - Claude outputs <thinking>...</thinking> first, which we hide
-    // After thinking, there may be <draft>...</draft> or <dispatch>...</dispatch> that we also hide
+    // After thinking, there may be <draft>, <dispatch>, or <needs> tags that we also hide
     let isInsideThinking = true;
-    let isTrappingTags = false; // After thinking, buffer to check for <draft>/<dispatch>
+    let isTrappingTags = false; // After thinking, buffer to check for hidden semantic tags
     let thinkingBuffer = '';
-    let tagTrapBuffer = ''; // Buffer for checking draft/dispatch tags after thinking
+    let tagTrapBuffer = ''; // Buffer for checking semantic tags after thinking
     let thinkingContent = ''; // Store hidden thinking for logging
     let draftContent = ''; // Store draft content for metadata
     let dispatchTagContent = ''; // Store dispatch tag content for handling
@@ -1717,7 +1717,7 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
               thinkingBuffer = '';
             }
           }
-          // PHASE 2: TAG TRAP - Buffer to catch <draft> and <dispatch> before streaming
+          // PHASE 2: TAG TRAP - Buffer to catch hidden semantic tags before streaming
           // The draft tag typically comes right after </thinking>, before response text
           else if (isTrappingTags) {
             tagTrapBuffer += event.text;
@@ -1731,8 +1731,9 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
             const hasNeedsEnd = tagTrapBuffer.includes('</needs>');
 
             // Check for partial tag starts at the end of buffer
-            // Matches: <, <d, <dr, </, </d, etc. - anything that could become <draft>, </draft>, <dispatch>, </dispatch>
-            const hasPotentialTagStart = /<\/?d[a-z]*$/i.test(tagTrapBuffer);
+            // Matches: <d..., <n..., </d..., </n..., etc. - anything that could become
+            // <draft>, </draft>, <dispatch>, </dispatch>, <needs>, or </needs>.
+            const hasPotentialTagStart = /<\/?(d|n)[a-z]*$/i.test(tagTrapBuffer);
 
             // If we see opening tags, wait for closing tags
             const waitingForDraft = hasDraftStart && !hasDraftEnd;
@@ -1751,7 +1752,7 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
             // Exit conditions:
             // - Not waiting for any tags to complete
             // - Have substantial response content (>50 chars that doesn't start with <)
-            // - No partial tag at the end that might become <draft> or <dispatch>
+            // - No partial tag at the end that might become a hidden semantic tag
             // OR buffer is too big (safety limit)
             const hasResponseContent = trimmedStripped.length > 50 && !trimmedStripped.startsWith('<');
             const safeToExit = !waitingForDraft && !waitingForDispatch && !waitingForNeeds && hasResponseContent && !hasPotentialTagStart;
@@ -1771,7 +1772,7 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
             const hasUnclosedDispatch = combined.includes('<dispatch>') && !combined.includes('</dispatch>');
             const hasUnclosedDraft = combined.includes('<draft>') && !combined.includes('</draft>');
             const hasUnclosedNeeds = combined.includes('<needs>') && !combined.includes('</needs>');
-            const hasPotentialTagStart = /<\/?d[a-z]*$/i.test(combined);
+            const hasPotentialTagStart = /<\/?(d|n)[a-z]*$/i.test(combined);
 
             if (hasUnclosedDispatch || hasUnclosedDraft || hasUnclosedNeeds || hasPotentialTagStart) {
               // Buffer and wait for closing tag
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -22,6 +22,7 @@ import {
   feelHeardRequestSchema,
   getMessagesQuerySchema,
   MessageRole,
+  NeedCategory,
 } from '@meet-without-fear/shared';
 import { notifyPartner, publishSessionEvent, notifySessionMembers, publishMessageAIResponse, publishMessageError, publishTopicFrameUpdated } from '../services/realtime';
 import { successResponse, errorResponse } from '../utils/response';
@@ -89,6 +90,148 @@ function getFallbackInitialMessage(
   }
 
   return `Hey ${userName}, what's on your mind?`;
+}
+
+const STAGE2_ROADMAP_COPY =
+  `Here's what comes next: there are a few more steps in this process. First, each of you tries to understand what the other person might be going through. Then you'll each explore what matters most to you, and eventually use that to get clearer about what is possible next, whether together or separately.`;
+
+function scrubVisibleAIText(text: string): { text: string; scrubbed: boolean } {
+  const before = text;
+  const cleaned = text
+    .split(/\r?\n/)
+    .filter((line) => {
+      const trimmed = line.trim().toLowerCase();
+      return !(
+        trimmed.startsWith('i should') ||
+        trimmed.startsWith('so both lists should') ||
+        trimmed.startsWith('— so both lists should') ||
+        trimmed.startsWith("here's my plan") ||
+        trimmed.startsWith('the prompt says') ||
+        trimmed.startsWith('i need to ')
+      );
+    })
+    .join('\n')
+    .replace(/\bI should\b/gi, '')
+    .replace(/\bso both lists should be available\b/gi, '');
+
+  return { text: cleaned, scrubbed: cleaned !== before };
+}
+
+function isReadyForStage3RevealText(content: string): boolean {
+  const normalized = content.trim().toLowerCase();
+  return /\b(i'?m|i am|we are)?\s*ready\b/.test(normalized) &&
+    /(list|lists|needs|side by side|reveal|see them|see it|show)/.test(normalized);
+}
+
+async function getStage3GateResponse(sessionId: string, userId: string): Promise<string | null> {
+  const partnerId = await getPartnerUserId(sessionId, userId);
+  const progress = await prisma.stageProgress.findUnique({
+    where: {
+      sessionId_userId_stage: {
+        sessionId,
+        userId,
+        stage: 3,
+      },
+    },
+  });
+  if (!progress) return null;
+
+  const vessel = await prisma.userVessel.findUnique({
+    where: { userId_sessionId: { userId, sessionId } },
+    include: { identifiedNeeds: { orderBy: { createdAt: 'asc' } } },
+  });
+  const needs = vessel?.identifiedNeeds ?? [];
+  const gates = progress.gatesSatisfied as Record<string, unknown> | null;
+  const ownShared = gates?.needsShared === true;
+  const ownConfirmed = gates?.needsConfirmed === true || (needs.length > 0 && needs.every((need) => need.confirmed));
+
+  if (needs.length === 0) {
+    return "Let's first put words to what matters most for you here. What do you need in order to feel clear, grounded, or able to move forward from this?";
+  }
+
+  if (!ownConfirmed) {
+    return "I've captured a draft of what matters to you. Please review and confirm your needs before we move any further.";
+  }
+
+  if (!ownShared) {
+    return "Your needs are ready for your review. If they still feel right, you can choose to share them for the side-by-side step.";
+  }
+
+  if (!partnerId) {
+    return "Your needs are shared. We'll wait until your partner has shared theirs before showing anything side by side.";
+  }
+
+  const partnerProgress = await prisma.stageProgress.findUnique({
+    where: {
+      sessionId_userId_stage: {
+        sessionId,
+        userId: partnerId,
+        stage: 3,
+      },
+    },
+  });
+  const partnerGates = partnerProgress?.gatesSatisfied as Record<string, unknown> | null;
+  if (partnerGates?.needsShared !== true) {
+    return "Your needs are shared. We'll wait until your partner has shared theirs before showing anything side by side.";
+  }
+
+  return "Both needs lists are ready to review side by side. Take a look at them and notice what stands out before deciding whether they feel accurate.";
+}
+
+async function persistStage3ProposedNeeds(
+  sessionId: string,
+  userId: string,
+  proposedNeeds: Array<{ need: string; category: string; description: string; evidence: string[] }>
+): Promise<boolean> {
+  const validNeeds = proposedNeeds.filter((need) =>
+    need.need &&
+    need.description &&
+    Object.values(NeedCategory).includes(need.category as NeedCategory)
+  );
+  if (validNeeds.length === 0) return false;
+
+  const vessel = await prisma.userVessel.upsert({
+    where: { userId_sessionId: { userId, sessionId } },
+    create: { userId, sessionId },
+    update: {},
+  });
+
+  const existing = await prisma.identifiedNeed.count({ where: { vesselId: vessel.id } });
+  if (existing > 0) return false;
+
+  await prisma.identifiedNeed.createMany({
+    data: validNeeds.map((need) => ({
+      vesselId: vessel.id,
+      need: need.description || need.need,
+      category: need.category as NeedCategory,
+      evidence: need.evidence,
+      aiConfidence: 0.85,
+      confirmed: false,
+    })),
+  });
+
+  const progress = await prisma.stageProgress.findUnique({
+    where: { sessionId_userId_stage: { sessionId, userId, stage: 3 } },
+  });
+  if (progress) {
+    await prisma.stageProgress.update({
+      where: { id: progress.id },
+      data: {
+        gatesSatisfied: {
+          ...((progress.gatesSatisfied as Record<string, unknown> | null) ?? {}),
+          needsCaptured: true,
+          needsCapturedAt: new Date().toISOString(),
+        },
+      },
+    });
+  }
+
+  await publishSessionEvent(sessionId, 'session.needs_extracted' as any, {
+    stage: 3,
+    forUserId: userId,
+  });
+
+  return true;
 }
 
 // ============================================================================
@@ -444,9 +587,9 @@ export async function confirmFeelHeard(
         if (aiResponse) {
           // Parse the semantic tag response (micro-tag format)
           const parsed = parseMicroTagResponse(aiResponse);
-          transitionContent = parsed.response.trim() || `What you just did really mattered — sharing what's been weighing on you and staying with it until you felt heard takes real honesty.\n\nHere's what comes next: there are a few more steps in this process. First, each of you tries to understand what the other person might be going through. Then you'll each explore what matters most to you, and eventually work on a way forward together.\n\nThis next part might feel a little unusual — I'm going to ask you to try to imagine what ${partnerName || 'your partner'} might be experiencing, even though you might still be upset with them. I know that's a strange ask. But there's a lot of research showing that when each person genuinely tries to see what the other is going through, it's one of the strongest things you can do to actually work things out. It's a guess, not a test — you don't have to get it right. ${mutualPhrase}\n\nSo — what do you think might be going on for ${partnerName || 'your partner'} in all of this?`;
+          transitionContent = parsed.response.trim() || `What you just did really mattered — sharing what's been weighing on you and staying with it until you felt heard takes real honesty.\n\n${STAGE2_ROADMAP_COPY}\n\nThis next part might feel a little unusual — I'm going to ask you to try to imagine what ${partnerName || 'your partner'} might be experiencing, even though you might still be upset with them. I know that's a strange ask. But there's a lot of research showing that when each person genuinely tries to see what the other is going through, it's one of the strongest things you can do to actually work things out. It's a guess, not a test — you don't have to get it right. ${mutualPhrase}\n\nSo — what do you think might be going on for ${partnerName || 'your partner'} in all of this?`;
         } else {
-          transitionContent = `What you just did really mattered — sharing what's been weighing on you and staying with it until you felt heard takes real honesty.\n\nHere's what comes next: there are a few more steps in this process. First, each of you tries to understand what the other person might be going through. Then you'll each explore what matters most to you, and eventually work on a way forward together.\n\nThis next part might feel a little unusual — I'm going to ask you to try to imagine what ${partnerName || 'your partner'} might be experiencing, even though you might still be upset with them. I know that's a strange ask. But there's a lot of research showing that when each person genuinely tries to see what the other is going through, it's one of the strongest things you can do to actually work things out. It's a guess, not a test — you don't have to get it right. ${mutualPhrase}\n\nSo — what do you think might be going on for ${partnerName || 'your partner'} in all of this?`;
+          transitionContent = `What you just did really mattered — sharing what's been weighing on you and staying with it until you felt heard takes real honesty.\n\n${STAGE2_ROADMAP_COPY}\n\nThis next part might feel a little unusual — I'm going to ask you to try to imagine what ${partnerName || 'your partner'} might be experiencing, even though you might still be upset with them. I know that's a strange ask. But there's a lot of research showing that when each person genuinely tries to see what the other is going through, it's one of the strongest things you can do to actually work things out. It's a guess, not a test — you don't have to get it right. ${mutualPhrase}\n\nSo — what do you think might be going on for ${partnerName || 'your partner'} in all of this?`;
         }
 
         // Save the transition message to the database as Stage 2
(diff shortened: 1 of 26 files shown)


slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Large PR addressing multiple James/Catherine gold session bugs: Stage 3 needs capture via `<needs>` tag, empathy consent idempotency, stage transition message isolation (`messagesByUserId`), "I'm ready" gate interception, AI reasoning leak scrubbing, needs share/waiting UI states, and copy updates ("What Comes Next" instead of "Moving Forward Together"). Includes a thorough handoff doc.

## Feedback

### Blocking

- **[Bug] Tag trap regex doesn't catch partial `<needs>` tags** (`messages.ts:1552`, `messages.ts:1588`): The `hasPotentialTagStart` regex `/<\/?d[a-z]*$/i` only matches `<d...` patterns (for `<draft>` and `<dispatch>`). If a chunk boundary falls mid-`<needs>` (e.g., `<ne` at end of chunk), it won't be caught, potentially leaking raw `<needs>` tag text to users. Update the regex to also match `n` patterns, e.g.:
  ```ts
  const hasPotentialTagStart = /<\/?(d|n)[a-z]*$/i.test(tagTrapBuffer);
  ```
  The comment at line 1550 also needs updating to mention `<needs>`.

- **[Bug] `scrubVisibleAIText` can strip legitimate facilitation** (`messages.ts:107`): The filter `trimmed.startsWith('i need to ')` will strip any line the AI starts with "I need to..." — risky in Stage 3 where needs articulation is the focus. A line like "I need to check in with you about something" would be silently removed. Consider either:
  1. Only applying this scrub during Stage 3 (where the reasoning leak was observed), or
  2. Tightening the patterns to be more specific (e.g., `i need to follow` / `i need to present` / `i need to check the prompt`)

### Non-blocking

- **[Correctness] `isReadyForStage3RevealText` has no negation check** (`messages.ts:128-131`): "I'm NOT ready to see the lists" would match both regexes and trigger the gate response. Consider adding `!/\bnot\b/.test(normalized)` or similar.

- **[Types] `'session.needs_extracted' as any`** (`messages.ts:237`): The `as any` cast bypasses type safety. Worth adding this event type to `shared/src/dto/realtime.ts` so it's properly typed.

- **[Style] Hardcoded `aiConfidence: 0.85`** (`messages.ts:216`): Magic number — consider extracting to a named constant.

- **[Merge] PR has conflicts** — session context shows `mergeable: "CONFLICTING"`. Needs rebase before merge.

## Good patterns

- Empathy consent idempotency in `stage2.ts` is well-implemented — handles both the pre-check and the P2002 race gracefully.
- `messagesByUserId` in stage transitions (`stage2.ts`, `stage3.ts`) fixes the cross-user message leak (bug C4 from the handoff).
- `needs-waiting-for-partner` waiting state correctly hides input while waiting for partner.
- `getNeedsComparison` now filters to confirmed-only needs — proper gating.
- Removing refinement `onSendRefinement` from `ViewEmpathyStatementDrawer` fixes bug C3 (refinement instructions appearing as chat).
- `STAGE2_ROADMAP_COPY` constant deduplicates the roadmap text.
- The handoff doc is excellent — structured, actionable, and clearly separates James/Catherine sides.

## Verdict

**Request changes** — the tag trap regex bug (partial `<needs>` not caught at chunk boundaries) can cause raw tag leakage to users. The `scrubVisibleAIText` over-stripping is also a real risk in Stage 3. Both should be fixed before merge, along with the merge conflicts.

slam-paws (changes requested):
## Review — PR #342: Fix James/Catherine Gold Session Gates

Good work addressing the Stage 3 gating bugs and the Stage 2→3 realtime issues documented in the handoff. The `messagesByUserId` fix for stage transitions, the `needs-waiting-for-partner` waiting state, the `scrubVisibleAIText` safety net, and the `STAGE2_ROADMAP_COPY` deduplication are all solid improvements. The handoff doc is excellent context.

There are a few issues that need attention before merge:

---

### Blocking

**1. [Bug] `scrubVisibleAIText` applied globally — will strip legitimate Stage 3 facilitation**
`messages.ts` — `scrubVisibleAIText()`

The filter `trimmed.startsWith('i need to ')` strips any AI line beginning with "I need to…" — but this function runs on **all stages**, not just Stage 3. In Stage 3 (Need Mapping), the facilitator commonly reflects needs back: *"I need to make sure I'm capturing this right"* or *"I need to pause here…"* These are valid therapeutic lines that would be silently removed.

Suggestion: Either scope the scrub to only run in Stage 3 (where planner leakage was observed), or tighten the patterns to target planning-specific phrases like `"i need to follow"`, `"i need to present"`, `"i need to check the prompt"`.

**2. [Bug] `persistStage3ProposedNeeds` has no transaction — data inconsistency on partial failure**
`messages.ts` — `persistStage3ProposedNeeds()`

The function performs 5 sequential DB calls (vessel upsert → count → createMany → find progress → update progress) without a `prisma.$transaction()` wrapper. If `createMany` succeeds but the `stageProgress.update` fails, you'll have orphaned `IdentifiedNeed` records without the `needsCaptured` gate being set. The rest of the codebase (e.g., `invitations.ts`) uses `$transaction` for exactly this pattern.

**3. [Bug] `persistStage3ProposedNeeds` has a TOCTOU race on the idempotency check**
`messages.ts` — `persistStage3ProposedNeeds()`

```typescript
const existing = await prisma.identifiedNeed.count({ where: { vesselId: vessel.id } });
if (existing > 0) return false;
await prisma.identifiedNeed.createMany({ ... });
```

Two concurrent streams for the same user can both see `existing === 0` and both proceed to `createMany`, producing duplicate needs records. Wrapping in a serializable transaction (or using a DB unique constraint) would fix this.

---

### Should Fix

**4. [Correctness] `isReadyForStage3RevealText` has no negation handling**
`messages.ts:isReadyForStage3RevealText()`

The regex `/\b(i'?m|i am|we are)?\s*ready\b/` matches "ready" without checking for negation. "I'm NOT ready to see the lists" would match and trigger the gate response, potentially confusing the user. A simple negation check (`/\bnot\s+ready\b/` early-return false) would prevent this.

**5. [Types] `'session.needs_extracted' as any` bypasses type safety**
`messages.ts:237`

This event type doesn't exist in `shared/src/dto/realtime.ts`. The `as any` cast means the mobile client has no type-safe handler for it. Either add `session.needs_extracted` to the `SessionEventType` union, or use an existing event type.

**6. [Style] Hardcoded `aiConfidence: 0.85`**
`messages.ts:216`

Magic number. Consider `const DEFAULT_AI_CONFIDENCE = 0.85` or similar.

---

### Noted Positives

- **`messagesByUserId` in stage transitions** — Fixes the critical bug where Catherine saw James-addressed transition messages. Clean `Object.fromEntries` pattern.
- **`getNeedsComparison` filtering to `confirmed: true`** — Prevents unconfirmed needs from leaking into the comparison view.
- **Early-return guard when either user has 0 confirmed needs** — Prevents empty comparison responses.
- **`needs-waiting-for-partner` waiting state** — Proper gating so the user knows what's happening.
- **Removing `onSendRefinement` from `ViewEmpathyStatementDrawer`** — Stops refinement instructions from appearing as chat messages.
- **Stage 4 rename "Moving Forward Together" → "What Comes Next"** — Supports the no-shared-agreement benchmark.
- **Sharing prompt safety clause** (preserves accountability, doesn't excuse escalation) — Important for the James/Catherine scenario.
- **E2E auth cache fix** — URL params now correctly override the cached identity.
- **Handoff doc quality** — Thorough, reproducible, well-organized.

---

### Note: Previous Review Correction

The previous review flagged the `hasPotentialTagStart` regex as not catching `<needs>` tags. This was **incorrect** — the PR already updates the regex from `/<\/?d[a-z]*$/i` to `/<\/?(d|n)[a-z]*$/i` in both Phase 2 and Phase 3 of the tag trap. That issue is resolved.

slam-paws (commented):
## PR Review: Fix James/Catherine Gold Session Gates

**Reviewer:** Slam Paws (bot)
**Verdict:** Solid, comprehensive fix for critical Stage 3 gating bugs. A few items to address before merge.

---

### What This PR Does

Addresses 10+ bugs found during James/Catherine gold session testing, organized into:

1. **Stage 3 gate enforcement** — AI can no longer promise side-by-side needs reveal before backend gates are satisfied. New `getStage3GateResponse()` intercepts "I'm ready" messages and returns gate-appropriate responses.
2. **Needs persistence pipeline** — New `<needs>` semantic tag in AI output, parsed by `micro-tag-parser`, persisted via `persistStage3ProposedNeeds()` with serializable isolation + retry.
3. **Planner text scrubbing** — `scrubVisibleAIText()` strips leaked planning language ("I should", "so both lists should", etc.) from user-facing responses.
4. **forUserId isolation** — Stage completion events now send `messagesByUserId` map instead of a single `message`, preventing Catherine from seeing James-addressed prompts.
5. **No-agreement language** — "Moving Forward Together" → "What Comes Next"; roadmap copy updated from "work on a way forward together" to "get clearer about what is possible next, whether together or separately."
6. **Safety/accountability in sharing** — New constraint #6 in reconciler sharing prompt prevents excusing escalation.
7. **Mobile UX** — Separated needs-review / needs-share panels, added `needs-waiting-for-partner` state, hid input during gate panels, fixed ActivityDrawer z-index and close button, removed empathy refinement-as-chat.
8. **E2E auth cache drift** — URL params now override cached E2E identity.
9. **Realtime invalidation** — Broader query invalidation on stage transitions and needs events.

---

### Issues

#### 1. No tests for new backend functions (Medium)

`scrubVisibleAIText`, `isReadyForStage3RevealText`, `getStage3GateResponse`, and `persistStage3ProposedNeeds` are all non-trivial functions with zero test coverage. The micro-tag-parser test file exists but has no coverage for `proposedNeeds` or `<needs>` parsing.

- `scrubVisibleAIText` has regex logic that could easily regress.
- `isReadyForStage3RevealText` has nuanced negative lookahead ("not ready" must not match) that should be pinned by tests.
- `persistStage3ProposedNeeds` has serializable transaction retry logic.

At minimum, unit tests for the pure functions (`scrubVisibleAIText`, `isReadyForStage3RevealText`) and parser tests for `<needs>` block in `micro-tag-parser.test.ts` should be added.

#### 2. `scrubVisibleAIText` is a brittle allowlist (Low — acknowledged band-aid)

The planner prefix list will drift as model behavior changes. The current approach is reasonable as an immediate fix, but consider:
- Adding a catch-all for lines starting with common reasoning markers (`"let me"`, `"my task is"`, `"according to the prompt"`)
- Logging when scrubbing occurs (already done via `scrubbedPlannerText` in fallback path, but not logged in the streaming path)

#### 3. `isReadyForStage3RevealText` false-negative gap (Low)

The function requires BOTH a readiness word (`ready`) AND a reveal keyword (`list|lists|needs|side by side|reveal|see them|see it|show`). Simple messages like "Yes, I'm ready" or "Ready" alone won't match. Since false negatives just let the message proceed to the AI (which should then be constrained by the prompt), this is acceptable — but worth documenting the intent.

#### 4. `<needs>` tag collision risk (Low)

If a user message or AI response literally contains `<needs>` as conversational text, it would be stripped. Very unlikely in practice, but the same risk applies to `<draft>` and `<dispatch>` so this is consistent with existing design.

#### 5. Bug report doc placement (Nit)

`docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md` is 473 lines with local URLs, session IDs, and reproduction details. It's valuable context but reads more like an investigation artifact than a living doc. Consider whether `docs/archive/` or `.planning/` would be more appropriate per the doc routing rules, since it won't need ongoing updates as bugs are fixed.

---

### What Looks Good

- **Gate logic is sound** — `getStage3GateResponse` checks the correct progression: needs exist → confirmed → shared → partner shared → both ready. Each state returns a contextually appropriate message.
- **Serializable transaction with retry** — `persistStage3ProposedNeeds` uses `Serializable` isolation with retry on `P2034`, which is the right pattern for preventing duplicate needs creation.
- **`messagesByUserId` with backward compat** — Mobile handles both new `messagesByUserId` and legacy `message` format gracefully.
- **`confirmed: true` filter** on `getNeedsComparison` — Prevents showing unconfirmed needs in side-by-side view.
- **Safety prompt addition** — The new constraint #6 in sharing.ts is well-written and directly addresses the James escalation concern.
- **Panel separation** — Splitting needs-review and needs-share into distinct panels with correct gating is a clean design.
- **Input hiding** — Chat input is now hidden during all needs gate states, preventing confused free-text input.
- **Streaming tag trap updates** — The `<needs>` tag is handled consistently across all three streaming phases (thinking, tag-trap, normal).

---

### Summary

This is a high-quality fix for critical product bugs. The gate enforcement, forUserId isolation, and no-agreement language changes are all important correctness fixes. The main gap is test coverage for the new backend logic — the pure functions and parser additions should have tests before merge to prevent regressions. The rest are minor.

slam-paws:
On it — reviewing now.

backend/src/controllers/messages.ts
```diff
@@ -91,6 +92,183 @@ function getFallbackInitialMessage(
...
+    .replace(/\bso both lists should be available\b/gi, '');
+
+  return { text: cleaned, scrubbed: cleaned !== before };
+}
+
+export function isReadyForStage3RevealText(content: string): boolean {
+  const normalized = content.trim().toLowerCase();
+  if (/\bnot\s+ready\b/.test(normalized) || /\b(?:don'?t|do not)\s+(?:want|feel ready)\b/.test(normalized)) {
+    return false;
+  }
```
slam-paws, line 131: The inline `.replace(/\bI should\b/gi, \x27\x27)` removes "I should" from ALL text, not just planner lines. If the AI ever writes facilitation like "I should mention that what you said really matters…", it gets silently corrupted to " mention that what you said really matters…" with a leading space.

The line-prefix filter above is precise and safe. Consider either:
1. Removing these inline replacements and relying solely on the prefix filter, or
2. Only applying inline replacements to lines that were NOT already filtered (though this is trickier).

mobile/src/screens/UnifiedSessionScreen.tsx
```diff
@@ -507,6 +493,7 @@ export function UnifiedSessionScreen({
             role: 'AI' as const,
             sessionId,
```
slam-paws, line 494: This fallback path now requires `legacyMessage?.forUserId === user?.id`. If the old backend sends `data.message` without `forUserId`, this condition is `undefined === userId` → false, and the transition message is silently dropped.

This is the *correct* fix for bug C4 (Catherine seeing James's prompt), but be aware during rolling deploys: if backend deploys after mobile, transition messages disappear until the new backend is live. Not blocking if you deploy backend first/together.

backend/src/controllers/messages.ts
```diff
@@ -91,6 +92,183 @@ function getFallbackInitialMessage(
...
+  if (!persisted) return false;
+
+  await publishSessionEvent(sessionId, 'session.needs_extracted', {
+    stage: 3,
+    forUserId: userId,
+  });
+
+  return true;
+}
+
```
slam-paws, line 271: `persistStage3ProposedNeeds` has transaction logic with serialization conflict retry, multiple DB writes, and event publishing — but no dedicated unit test. The `scrubVisibleAIText` and `isReadyForStage3RevealText` helpers are well-tested, but this function is the most complex new addition and is only exercised indirectly through the streaming path.

Consider adding at least:
- Happy path: proposed needs persisted + event published
- Idempotency: second call with existing needs returns false
- Invalid category filtering: bad categories are excluded

slam-paws (approved):
## Summary

Substantial fix addressing Stage 3 gate enforcement, planner text leakage, per-user transition messages, and UI panel flow — all driven by real bugs found during James/Catherine gold session testing. Touches backend (messages controller, micro-tag parser, stage prompts, realtime events), mobile (chat UI state, waiting states, streaming, session screen), and shared types.

## What's working well

- **Stage 3 `<needs>` tag** follows the established pattern from Stage 4's `<strategies>` — clean, consistent extension of the semantic tag architecture.
- **`getStage3GateResponse`** is a solid gate enforcement function with clear, ordered checks (no needs → unconfirmed → unshared → partner not shared → both ready). This is exactly the kind of explicit gating that was missing.
- **`persistStage3ProposedNeeds`** with serializable isolation + retry is correct for preventing duplicate needs from concurrent AI responses.
- **`messagesByUserId` in stage transition events** fixes the cross-user message pollution (bug C4) at the right layer — the event now carries per-user messages so each client picks its own.
- **Panel split** (needs-review vs needs-share) makes the Stage 3 flow clearer — confirm first, then explicitly consent to share.
- **`getNeedsComparison` filtering by `confirmed: true`** prevents premature reveals at the API layer.
- **Roadmap copy** ("whether together or separately") properly supports the no-agreement path.
- **Reconciler safety prompt** (accountability + anger ownership) is a thoughtful guardrail for the James/Catherine scenario.

## Feedback

- **[Correctness]** `scrubVisibleAIText` inline replacements (`.replace(/\bI should\b/gi, '')`) are overly aggressive — see inline comment. The line-prefix filter is precise; the inline replacements could corrupt valid facilitation text.
- **[Tests]** `persistStage3ProposedNeeds` has no dedicated unit tests despite being the most complex new function (transaction + retry + event publishing). See inline comment.
- **[Deploy]** The `partner.stage_completed` handler's fallback for legacy `data.message` without `forUserId` will silently drop transition messages during a rolling deploy if mobile updates before backend. Not blocking if deployed together. See inline comment.
- **[Docs]** The handoff doc (`docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md`) is a point-in-time bug report — consider whether `docs/archive/` is a better home since it documents a testing session, not living product state.
- **[Minor]** `needs-pending` waiting config changed to `hideInput: true` — this silences chat during needs exploration waiting, which is the right call for gate enforcement but is a behavior change users may notice.

## Verdict

**Approve.** The core changes are correct, well-structured, and address real product bugs. The inline comment about `scrubVisibleAIText` is worth a follow-up (false positive risk on legitimate "I should" text), and `persistStage3ProposedNeeds` deserves dedicated tests, but neither blocks merge. Ship it.

slam-paws (approved):
## PR Review: [codex] Fix James/Catherine gold session gates

**Verdict: Approve**

Substantial, well-motivated PR that fixes real bugs surfaced during James/Catherine gold session testing. The fixes address prompt safety, state gating, privacy boundaries, realtime cache coherence, and UX clarity — all critical for a no-shared-agreement benchmark.

### What's good

- **Stage 3 gate enforcement** (`getStage3GateResponse`, `persistStage3ProposedNeeds`) — finally makes the needs reveal pathway state-driven rather than relying on AI text output. Serializable transaction with retry is the right approach.
- **`<needs>` tag pipeline** — clean extension of the existing `<thinking>`/`<draft>`/`<dispatch>` pattern. Parser, stream handler, and prompt contract all updated consistently.
- **`messagesByUserId` for stage transitions** — fixes the cross-user message leak (C4). Per-user addressed messages with legacy fallback is good defensive design.
- **Prompt hardening** — the new Stage 3 forbidden list (no premature reveal promises, no planner leakage) and the accountability safety constraint in sharing.ts are both well-targeted.
- **"Moving Forward Together" → "What Comes Next"** — small but important for the no-agreement path. Updated in both shared enums and mobile.
- **Separate `needs-review` / `needs-share` panels** — proper gating distinction between confirming and consenting to share.
- **Empathy refinement removal** — removing `onSendRefinement` fixes C3 (refinement instructions appearing as chat). Clean approach via conditional render in ViewEmpathyStatementDrawer.

1. **`isReadyForStage3RevealText` false-positive on hedged negation**: "I'm not quite ready to see the lists" would pass — the negative regex `/\bnot\s+ready\b/` doesn't match when words intervene between "not" and "ready". Low risk since `getStage3GateResponse` enforces actual DB state downstream, but worth noting for future refinement.

2. **`scrubVisibleAIText` inline replace could clip legitimate text**: `replace(/\bI should\b/gi, '')` would turn "I should mention that..." into " mention that...". The prefix `"i need to make sure both lists"` could also catch legitimate facilitation like "I need to make sure both lists are accurate." These are pragmatic safety nets and the scrubbed flag enables logging, but consider tightening to only apply within `<thinking>` leakage context rather than broadly on all visible text.

3. **Double scrub path**: `scrubVisibleAIText` runs during streaming (per-chunk) and again on the final `parsed.response`. This is idempotent so it's safe, but worth being aware of.

4. **Bug handoff doc placement**: `docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md` (473 lines) reads as a point-in-time artifact rather than a living doc. Consider moving to `docs/archive/` once the fixes land to keep `docs/product/` focused on current product documentation.

5. **`needs-pending` now hides input**: Previously users could chat while waiting for partner needs confirmation. This is the right call for gate clarity but worth flagging in case it causes confusion for users who want to continue the conversation.

CI green. Tests cover the new helpers well (scrub, ready-text detection, needs parsing, malformed JSON). Approve.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Large, well-structured PR that fixes multiple Stage 3 gating, isolation, and prompt-leakage bugs discovered during the James/Catherine gold session benchmark. Key additions: backend needs-capture pipeline (`<needs>` tag parsing → `persistStage3ProposedNeeds`), planner-text scrubbing, per-user message isolation via `messagesByUserId`, explicit needs-share CTA panel, and tighter input-hiding during Stage 3 waiting states. Also removes empathy-refinement-as-chat (bug C3), adds accountability safety rail to reconciler sharing prompt, and renames Stage 4 to "What Comes Next."

## Feedback

### Minor / Non-blocking

- **[Docs drift]** The stage rename "Moving Forward Together" → "What Comes Next" updates `shared/src/enums.ts` and `mobile/.../UnifiedSessionScreen.tsx`, but two living wireframe docs still reference the old name:
  - `docs/mobile/wireframes/index.md:43`
  - `docs/mobile/wireframes/chat-interface.md:183,293`

- **[Docs placement]** `docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md` is a session-specific debugging artifact (473 lines of temporal context). Consider moving it to `docs/archive/` so it doesn't sit alongside living product docs, or converting it to GitHub issues for trackability.

- **[Defense-in-depth note]** `scrubVisibleAIText` uses a blocklist of planner prefixes — inherently incomplete but reasonable as a belt-and-suspenders layer on top of the prompt instructions. Worth noting that new planner phrases will require list updates.

- **[Potential tag regex match]** `hasPotentialTagStart` regex changed from `/<\/?d[a-z]*$/i` to `/<\/?(d|n)[a-z]*$/i` — this now also matches partial `<nav>`, `<nobr>`, etc. In practice Claude's constrained output format makes false matches unlikely, but worth a comment noting why `n` was added.

### Looks good

- `persistStage3ProposedNeeds` — clean use of Serializable isolation + retry with proper conflict detection (`P2034`). The `existing > 0 → return false` idempotency guard is solid.
- `getStage3GateResponse` — gate ordering (no needs → unconfirmed → unshared → partner not shared → both ready) is correct and well-structured.
- `messagesByUserId` pattern with legacy fallback in `UnifiedSessionScreen` — good backward compatibility while fixing the cross-user message isolation bug (C4).
- `getNeedsComparison` now filtering by `confirmed: true` and returning empty early — prevents showing unconfirmed/phantom needs.
- E2E auth cache fix is minimal and correct (URL params override cache, cache is fallback).
- Test coverage for `scrubVisibleAIText`, `isReadyForStage3RevealText`, `<needs>` parsing, and updated chatUIState/getWaitingStatus tests.
- Reconciler sharing prompt safety rule #6 (accountability/escalation) is well-written.

## Verdict

**Approve** — the core changes are correct, well-tested, and address real gold-session bugs. The doc-drift items are non-blocking and can be fixed in a follow-up.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Comprehensive fix for Stage 3 gating, planner text leakage, cross-user message isolation, and UI state management — all driven by the James/Catherine gold session bug handoff. This is a significant and well-structured set of changes across backend, mobile, and shared packages.

## What this PR does

**Backend:**
- Adds Stage 3 gate logic (`getStage3GateResponse`, `persistStage3ProposedNeeds`) so the AI can no longer promise a side-by-side reveal before both users complete their gates
- Adds `<needs>` semantic tag to the streaming pipeline + micro-tag parser, with proper extraction, validation, and persistence
- Adds `scrubVisibleAIText` to strip leaked planner/reasoning text ("I should", "here's my plan", etc.)
- Adds `isReadyForStage3RevealText` to detect user intent for reveal without false-positive on generic "I'm ready" messages
- Fixes `partner.stage_completed` events to use `messagesByUserId` instead of a single `message`, fixing the cross-user message isolation bug (C4 in handoff)
- Adds accountability/safety guardrails to the sharing prompt (no implying loss of control)
- Updates roadmap copy: "work on a way forward together" → "get clearer about what is possible next, whether together or separately"
- Properly types `session.needs_reveal_ready` and `partner.needs_validated` (removes `as any` casts)

**Mobile:**
- Adds `needs-share` panel (separate CTA from `needs-review`) so confirm and share are distinct steps
- Adds `needs-waiting-for-partner` waiting state (hides input while waiting for partner to share)
- Hides chat input during needs review/share/reveal panels — prevents free-text "I'm ready" from bypassing gates
- Fixes `partner.stage_completed` handler to pick the user-addressed message from `messagesByUserId`
- Invalidates needs/progress/comparison queries on relevant realtime events
- Fixes E2E auth identity cache drift (URL params now override cached identity)
- Renames Stage 4 to "What Comes Next" (removes "together" bias)
- Adds explicit close button to ActivityDrawer, lowers z-index
- Removes `onSendRefinement` from empathy drawer (refinement via chat was displaying as normal messages)

**Tests:** Good coverage for `scrubVisibleAIText`, `isReadyForStage3RevealText`, `<needs>` block parsing (valid + malformed JSON), and UI state changes.

## Feedback

- **[Minor — scrubVisibleAIText]** `backend/src/controllers/messages.ts:120` — The inline `.replace(/\bI should\b/gi, '')` could produce broken sentences if "I should" appears mid-sentence in legitimate facilitation (e.g., "What I should mention is…" → "What  mention is…"). The line-level prefix filter handles the obvious cases well; the inline replace is a more aggressive safety net. Worth keeping for now since the alternative (planner text leaking to users) is worse, but worth monitoring for false positives in production logs via the `scrubbed: true` flag.

- **[Observation — isReadyForStage3RevealText]** `backend/src/controllers/messages.ts:149` — Regex-based intent classification is inherently fragile. The current implementation is well-bounded (only fires in Stage 3, requires both "ready" + reveal context keywords, has negation checks). The tests cover the important cases. Since this is a gate response (not an irreversible action), false positives just return a helpful status message rather than causing harm.

- **[Nice — Serializable isolation]** `backend/src/controllers/messages.ts:268` — The `Serializable` isolation level + retry logic in `persistStage3ProposedNeeds` is the right call for preventing duplicate needs insertion. Clean implementation.

- **[Nice — Type cleanup]** Removing `as any` casts from `stage3.ts:625,745` by properly adding event types to `SessionEvent` is a good pattern.

- **[Question — Handoff doc location]** `docs/product/james-catherine-gold-session-bug-handoff-2026-05-04.md` — This 473-line file is extremely valuable as a bug report, but it contains local URLs, session IDs, and reproduction steps that are session-specific. Should this live under `docs/archive/` or `.planning/` rather than `docs/product/`? It reads more like a test report than a living product doc.

- **[Nice — messagesByUserId]** The fix from single `message` to `messagesByUserId` in stage completion events (stage2.ts, stage3.ts, UnifiedSessionScreen.tsx) correctly solves the cross-user message isolation issue. The legacy fallback in the mobile handler is a good backward-compat touch.

## Verdict

**Approve.** This is a thorough, well-tested fix for critical Stage 3 gating and isolation bugs. The changes are well-structured across the stack and the handoff document provides excellent context. The only open question is the location of the handoff doc file.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Comprehensive fix for Stage 3 session gating issues discovered during the James/Catherine gold session. Addresses premature "both lists ready" prompts, planner text leaking to users, missing needs persistence, stale realtime UI transitions, and resolution-biased language.

## Feedback

- **Good: Serializable transaction isolation** (`messages.ts:204-248`): The `persistStage3ProposedNeeds` function uses `Serializable` isolation with retry — solid approach for preventing duplicate needs insertion in concurrent scenarios.

- **Good: Double-layer scrub defense** (`messages.ts:171-192`): `scrubVisibleAIText` applied both during streaming and after `parseMicroTagResponse` ensures planner leakage is caught regardless of timing. The prefix list is a reasonable first defense.

- **Good: Backward-compatible event payloads** (`UnifiedSessionScreen.tsx:~490-495`): The `messagesByUserId` pattern with `legacyMessage` fallback handles the transition cleanly.

- **Good: Stage 4 rename** (`shared/src/enums.ts:60`): "Moving Forward Together" → "What Comes Next" correctly avoids implying forced reconciliation for no-agreement paths.

- **Good: Safety/accountability prompt guard** (`reconciler/sharing.ts:1168-1175`): The new rule about never implying loss of control is well-phrased and addresses the James draft escalation issue directly.

- **Minor: `needs-pending` is now dead code** (`getWaitingStatus.ts:24,224,235`): `computeWaitingStatus` no longer returns `'needs-pending'` — the condition that produced it now returns `'needs-waiting-for-partner'`. This means:
  - The config entry for `'needs-pending'` (line 180) is unreachable
  - The transition check `previousStatus === 'needs-pending'` (line 235) won't fire for new sessions, so `'partner-confirmed-needs'` becomes unreachable too
  - The comment on line 224 is now stale ("Only show needs-pending...")
  
  Not a bug since the new CTA panels cover the gap, but worth a cleanup pass to remove dead states and update the comment.

- **Minor: Test name drift** (`getWaitingStatus.test.ts:266`): "does not return needs-pending when in Stage 2" still passes but the name references a state that's no longer produced. Consider updating the description.

- **Observation: `hasPotentialTagStart` regex broadening** (`messages.ts:~1762`): `/<\/?(d|n)[a-z]*$/i` could buffer on edge-case fragments like `<no` in AI text. The 2000-char safety limit protects against real problems, so this is fine in practice.

## Verdict

**Approve.** The fixes are well-targeted at real gold session bugs, the Stage 3 gate contract is much stronger, and the test coverage is solid. The dead `needs-pending` code is a minor cleanup item that can be addressed in a follow-up.

PR #344, opened 2026-05-04
[codex] Fix James Catherine gold session gates

shantamg:
## Summary

Fixes the James/Catherine gold-session failures across PR 1, PR 2, and the Stage 3 structured-needs persistence follow-up.

## What Changed

- Prevents cross-user private prompt leakage by publishing addressed transition messages and filtering broadcast realtime payloads on mobile before cache insertion.
- Centralizes Stage 2 and Stage 3 realtime cache invalidation for session state, stage progress, message lists, pending actions, and badges.
- Hides chat input during system/partner-blocked states and removes user-facing "internal reconciler" copy.
- Adds Stage 3 hidden structured needs output, parses `<needs>` metadata, persists captured needs idempotently, and marks `needsCaptured` gates.
- Shows captured Stage 3 needs as a styled chronological chat card that opens the existing needs review drawer.

## Root Cause

Private per-user AI messages were being broadcast on session-channel events and inserted by the mobile client without checking `forUserId`. Stage 2/3 realtime handlers also refreshed only partial cache state, so clients could remain stale until manual refresh. Stage 3 needs were summarized in chat but never persisted as structured `IdentifiedNeed` rows.

## Validation

- `backend`: `npm test -- --runTestsByPath src/controllers/__tests__/stage2-copy.test.ts src/routes/__tests__/stage2.test.ts src/routes/__tests__/stage3.test.ts src/routes/__tests__/messages.test.ts src/utils/__tests__/micro-tag-parser.test.ts --runInBand`
- `backend`: `npm run check`
- `mobile`: `npm test -- --runTestsByPath src/utils/__tests__/realtimePrivacy.test.ts src/utils/__tests__/realtimeInvalidation.test.ts src/utils/__tests__/chatUIState.test.ts src/components/__tests__/ShareTopicDrawer.copy.test.ts src/hooks/__tests__/useRealtime.test.ts --runInBand`
- `mobile`: `npm run check`
- `shared`: `npm run check`

## Note

`mobile/src/components/__tests__/ChatInterface.test.tsx` still fails in the existing harness with an axios/expo stream setup error and stale `send-button` expectations; this PR does not touch that component path.


slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary

Fixes the James/Catherine gold-session gates by:
1. Adding realtime privacy filtering so broadcast events carrying per-user payloads (`forUserId`) are dropped by the wrong recipient
2. Extracting structured `<needs>` tags from Stage 3 AI streaming into a centralized `captureProposedNeedsForUser` service
3. Removing user-facing "internal reconciler" language throughout (replaced with plain process language)
4. Blocking input during system-review states (reconciler-analyzing, needs-pending, etc.)
5. Centralizing stage-aware cache invalidation into `realtimeInvalidation.ts` helpers

## Feedback

- **[Architecture]** `getOrCreateUserVessel` is now duplicated — one copy in `stage3.ts` (used by 5+ other controller functions) and a new copy in `services/needs.ts`. Not blocking since the stage3 copy serves functions not touched here, but worth consolidating in a follow-up (export from `needs.ts` or a shared `vessel` service).

- **[Performance]** `getStage2RealtimeInvalidationQueryKeys` includes `messageKeys.infinite(sessionId)`, which means every `refreshStage2RealtimeState()` call invalidates the messages query — even events like `empathy.revealed` or `empathy.refining` that previously did not refetch messages. The broader invalidation is fine for correctness, but keep an eye on whether this creates visible refetch flicker in Stage 2.

- **[Correctness]** In `captureProposedNeedsForUser` (`needs.ts`), the `findMany` after the transaction has no filter on `confirmed: false`, so it returns both confirmed and unconfirmed needs. This appears intentional (return the full picture), but worth confirming the caller expectation matches.

- **[Type safety]** Minor: `queryClient.getQueryData<any>` on the `empathyStatus` key in the new `empathy.status_updated` handler (`UnifiedSessionScreen.tsx`). Using a proper type here would make the version comparison safer, but consistent with how the codebase accesses query data elsewhere.

- **[Draft status]** PR is marked as draft with merge conflicts (`CONFLICTING`). Will need a rebase before merging.

## Verdict

**Approve.** The privacy filtering for broadcast realtime events is an important correctness fix, the needs extraction pipeline is clean and well-tested, and the reconciler language cleanup is solid. Good test coverage across all new utilities. The observations above are non-blocking.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Solid refactor that consolidates Stage 3 needs persistence into a shared `captureProposedNeedsForUser` service, fixes realtime event privacy (per-user addressed payloads instead of broadcast maps), removes user-facing "internal reconciler" language, and adds a `NeedsIdentifiedChatCard` for in-chat needs review. Good test coverage for the new utilities.

## Feedback

- **[Resilience]** The old `persistStage3ProposedNeeds` in `messages.ts` used `Serializable` isolation + P2034 retry (2 attempts). The new `captureProposedNeedsForUser` uses a plain `$transaction` with no retry. If serialization conflicts are a real concern for concurrent AI turns capturing needs, the retry could be worth preserving. *Non-blocking* — this may have been intentionally simplified if conflicts are rare in practice.

- **[Race condition]** `getOrCreateUserVessel` in `needs.ts` uses `findUnique` → `create` (two-step, non-atomic). The old code used `upsert` which handles the race atomically. If two concurrent calls hit this for the same user/session, the `create` will throw on the unique constraint. *Low risk* since vessels are typically created well before needs capture, but an `upsert` would be safer for the same cost.

- **[Privacy — good pattern]** The `realtimePrivacy.ts` utility and the early-exit filter in `useRealtime.ts` + `UnifiedSessionScreen.tsx` are a clean, testable approach. The precedence of `message.forUserId` over top-level `forUserId` makes sense for transition messages.

- **[UX change]** `reconciler-analyzing`, `revision-analyzing`, and `common-ground-pending` now hide input (`hideInput: true`). This is a meaningful behavior change — users can no longer chat while the system reviews empathy. The tests document this well.

- **[Needs tag ordering]** The `<needs>` tag is now processed before `<dispatch>` consistently across the streaming buffer, tag trap, and parser. The reorder in `stage-prompts.ts` (needs before draft in the output format) aligns with this. Clean.

- **[Metadata handling]** In `useStreamingMessage.ts`, `handleMetadata` is now called on every `complete` event regardless of whether `text_complete` was already received. This is the right fix — ensures needs-capture invalidation fires even when `text_complete` arrives first.

## Verdict

**Approve** — well-structured changes with good test coverage. The two minor observations (serialization retry removal, `findUnique`→`create` vs `upsert`) are non-blocking and low-risk.

slam-paws (approved):
## PR Review: Fix James/Catherine Gold Session Gates

**Verdict: Approve**

This is a well-structured PR that addresses multiple session gate issues observed in the James/Catherine gold session. The changes fall into five coherent areas:

### 1. Needs Capture Consolidation (`backend/src/services/needs.ts`)
The new `captureProposedNeedsForUser` centralizes needs persistence logic that was previously duplicated between `messages.ts` (inline `persistStage3ProposedNeeds`) and `stage3.ts` (`captureNeeds`). The consolidated version correctly:
- Replaces only unconfirmed needs (preserving user-confirmed ones)
- Updates `gatesSatisfied.needsCaptured` within the same transaction
- Returns the persisted records for the caller to use

### 2. Stage Transition Event Privacy (`stage2.ts`, `stage3.ts`)
Transition events (`partner.stage_completed`) now emit **per-recipient** payloads with `forUserId` instead of a single `messagesByUserId` map. This is a good architectural improvement — broadcast channels should carry addressed payloads so clients can filter without parsing nested maps. The new `stage.progress` event in stage2 separates public progress from private transition messages.

### 3. Realtime Privacy Guard (`realtimePrivacy.ts`, `useRealtime.ts`, `UnifiedSessionScreen.tsx`)
New `isRealtimePayloadAddressedToCurrentUser` filter drops events addressed to other users at two levels:
- **Hook level** (`useRealtime.ts`) — global guard before any handler runs
- **Screen level** (`UnifiedSessionScreen.tsx`) — defense-in-depth for session-specific handlers

The precedence logic (`messageForUserId ?? forUserId`) correctly handles both envelope-level and message-level addressing.

### 4. UI Gating Fixes
- `reconciler-analyzing` and `revision-analyzing` now **hide input** (`hideInput: true`) instead of showing "Keep Chatting". This prevents users from sending messages while the empathy review is running — fixing a race condition.
- `common-ground-pending` also hides input, keeping the UI locked while Stage 3 review controls own the next step.
- `needs-pending` state now triggers when needs are confirmed but not yet shared (new gate in `getWaitingStatus.ts`).
- Reconciler language scrubbed from user-facing copy (ShareTopicDrawer, stage2 prompts).

### 5. Micro-Tag Parser Improvements
- `parseNeedsBlock` now validates `NeedCategory` enum membership, rejecting invalid categories instead of passing them through.
- `<needs>` extraction moved before `<dispatch>` in the streaming pipeline to match the prompt output format order.
- Needs content stored as inner text (`needsMatch[1]`) rather than the full tag, then re-wrapped for `parseMicroTagResponse` — cleaner contract.

1. **`getOrCreateUserVessel` duplication**: The new `needs.ts` defines its own `getOrCreateUserVessel`, while `stage3.ts` likely still has one. Consider importing from a shared location in a follow-up.

2. **`realtimeInvalidation.ts` includes `messageKeys.infinite`**: Both stage 2 and stage 3 key sets include the message list. When `refreshStage2RealtimeState({ refetchMessages: true })` is called, messages get both invalidated (from the key set) and refetched (from the option). Not harmful but slightly redundant — the refetch supersedes the invalidation.

3. **`any` casts in `UnifiedSessionScreen`**: `canInsertRealtimeMessageForCurrentUser(addressedMessage as any, ...)` and the `queryClient.getQueryData<any>` for empathy status version checking — these work but could benefit from proper types in a follow-up.

4. **`NeedsIdentifiedChatCard`** is defined inline in `UnifiedSessionScreen.tsx`. Given it has its own styles block, it may warrant extraction to its own component file in a follow-up to keep the screen file manageable.

Good test coverage across all areas: realtime privacy, invalidation keys, micro-tag parser needs extraction, chatUIState input hiding, and stage2 copy regression guard.

🤖 Reviewed by Slam Paws

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Solid refactor consolidating Stage 3 needs persistence into a shared service (`captureProposedNeedsForUser`), adding realtime privacy filtering to prevent cross-user event leakage, fixing user-facing copy that exposed internal "reconciler" terminology, and adding a `NeedsIdentifiedChatCard` for in-chat review. Good test coverage throughout.

## Feedback

### Architecture (positive)

- **Privacy filtering** (`realtimePrivacy.ts`): Clean separation of the `forUserId` check into a reusable utility applied at both `useRealtime` and `UnifiedSessionScreen` level — defense-in-depth approach.
- **Per-recipient event publishing** (stage2.ts, stage3.ts): Publishing individual `partner.stage_completed` events per recipient instead of batching into `messagesByUserId` is the right structural change to make the privacy filter work correctly.
- **Consolidated invalidation** (`realtimeInvalidation.ts`): Centralizing stage-specific query key sets eliminates the repeated invalidation lists that were scattered across event handlers.
- **Needs service consolidation**: Eliminating duplication between `messages.ts` and `stage3.ts` by extracting to `services/needs.ts` is clean.

### Worth noting (non-blocking)

- **Transaction isolation downgrade** (`services/needs.ts`): The old `persistStage3ProposedNeeds` in `messages.ts` used `Serializable` isolation with a retry loop for serialization conflicts. The new `captureProposedNeedsForUser` uses default isolation. In practice this is likely fine since the delete-then-create pattern within a single transaction is safe, and concurrent AI summaries for the same user/session are unlikely. Just flagging the behavioral change.

- **Duplicate `getOrCreateUserVessel`**: Now exists in both `services/needs.ts` and `controllers/stage3.ts` with slightly different implementations (the service version uses `findUnique` + conditional `create`; the controller version does the same). Not blocking, but could be consolidated in a follow-up.

- **`any` type casts** (`UnifiedSessionScreen.tsx`): `canInsertRealtimeMessageForCurrentUser(addressedMessage as any, ...)` at the `partner.stage_completed` handler and `queryClient.getQueryData<any>` in `empathy.status_updated`. Pragmatic for this PR, but the addressed message type could be tightened.

- **`parseNeedsBlock` fallback to `extractJsonFromResponse`**: The double-parse (try `JSON.parse` then fall back to `extractJsonFromResponse`) is good defensive coding for malformed AI output, but worth adding a debug-level log when the fallback is hit to aid production debugging.

### Tests (positive)

- Copy guard tests (`stage2-copy.test.ts`, `ShareTopicDrawer.copy.test.ts`) are a smart pattern — static analysis tests that prevent terminology regression.
- `realtimePrivacy.test.ts` and `realtimeInvalidation.test.ts` cover the new utilities well.
- `chatUIState.test.ts` additions verify the input-hiding behavior changes for `reconciler-analyzing`, `revision-analyzing`, `needs-pending`, and `common-ground-pending`.
- `micro-tag-parser.test.ts` tests for both valid and malformed needs blocks.

### Input hiding changes (Stage 2)

Changing `reconciler-analyzing` and `revision-analyzing` from `hideInput: false` + `showKeepChattingAction: true` to `hideInput: true` + `showKeepChattingAction: false` is a UX decision that blocks user input during analysis. Makes sense to prevent confusion, just noting it's a deliberate behavior change.

## Verdict

**Approve** — Well-structured refactor with good test coverage, proper privacy filtering, and clean consolidation of scattered patterns. The non-blocking items above are minor follow-up candidates.

PR #347, opened 2026-05-05
Fix Stage 4 strategy readiness sync

shantamg:
## Summary
- sync Stage 4 strategy inventory through a new session.strategies_updated realtime event
- make Ready to Rank cache-first with optimistic progress/strategy updates and an already-ready waiting state
- add E2E-only pending invitation auto-accept for direct local invitee session URLs
- collapse obvious superseded strategy proposal drafts during Stage 4 capture
- add fix notes to the Adam/Eve Stage 4 live-run report

## Tests
- npm run check
- npm -w backend test -- src/utils/__tests__/strategy-dedupe.test.ts src/routes/__tests__/stage4.test.ts --runInBand
- npm -w mobile test -- src/utils/__tests__/getWaitingStatus.test.ts src/utils/__tests__/chatUIState.test.ts src/hooks/__tests__/useStages.test.ts --runInBand --forceExit
- npm run test (fails in existing mobile baseline: EmotionalBarometer label expectations, ChatInterface/axios stream setup, Sentry ESM transform/navigation, StrategicRepairScreen axios stream setup; backend and shared full suites pass)

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -39,6 +39,7 @@ import { getMilestoneContext, getSharedContentContext } from '../services/shared
 import { CONTEXT_WINDOW, trimConversationHistory } from '../utils/token-budget';
 import { estimateContextSizes, finalizeTurnMetrics, recordContextSizes } from '../services/llm-telemetry';
 import { captureProposedNeedsForUser } from '../services/needs';
+import { filterNewStrategiesAgainstExisting } from '../utils/strategy-dedupe';
 
 // ============================================================================
 // Helpers
@@ -2050,15 +2051,20 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
       // Deduplicate: check existing strategies to avoid duplicates from re-proposals
       const existingStrategies = await prisma.strategyProposal.findMany({
         where: { sessionId, createdByUserId: user.id },
-        select: { description: true },
+        select: { id: true, description: true },
       });
-      const existingDescriptions = new Set(existingStrategies.map((s) => s.description.toLowerCase()));
 
-      const newStrategies = metadata.proposedStrategies.filter(
-        (desc) => !existingDescriptions.has(desc.toLowerCase())
+      const { newStrategies, supersededIds } = filterNewStrategiesAgainstExisting(
+        existingStrategies,
+        metadata.proposedStrategies
       );
 
       if (newStrategies.length > 0) {
+        if (supersededIds.length > 0) {
+          await prisma.strategyProposal.deleteMany({
+            where: { id: { in: supersededIds } },
+          });
+        }
         await prisma.strategyProposal.createMany({
           data: newStrategies.map((description) => ({
             sessionId,
@@ -2069,6 +2075,14 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
           })),
         });
         logger.info(`[sendMessageStream:${requestId}] Created ${newStrategies.length} strategy proposals for user ${user.id}`);
+
+        const { publishSessionEvent } = await import('../services/realtime');
+        await publishSessionEvent(sessionId, 'session.strategies_updated', {
+          stage: 4,
+          updatedBy: user.id,
+          createdCount: newStrategies.length,
+          removedSupersededCount: supersededIds.length,
+        });
       }
     }
 
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -2060,23 +2060,24 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
       );
 
       if (newStrategies.length > 0) {
-        if (supersededIds.length > 0) {
-          await prisma.strategyProposal.deleteMany({
-            where: { id: { in: supersededIds } },
+        await prisma.$transaction(async (tx) => {
+          if (supersededIds.length > 0) {
+            await tx.strategyProposal.deleteMany({
+              where: { id: { in: supersededIds } },
+            });
+          }
+          await tx.strategyProposal.createMany({
+            data: newStrategies.map((description) => ({
+              sessionId,
+              createdByUserId: user.id,
+              description,
+              needsAddressed: [],
+              source: 'AI_SUGGESTED' as const,
+            })),
           });
-        }
-        await prisma.strategyProposal.createMany({
-          data: newStrategies.map((description) => ({
-            sessionId,
-            createdByUserId: user.id,
-            description,
-            needsAddressed: [],
-            source: 'AI_SUGGESTED' as const,
-          })),
         });
         logger.info(`[sendMessageStream:${requestId}] Created ${newStrategies.length} strategy proposals for user ${user.id}`);
 
-        const { publishSessionEvent } = await import('../services/realtime');
         await publishSessionEvent(sessionId, 'session.strategies_updated', {
           stage: 4,
           updatedBy: user.id,
--- a/backend/src/controllers/session-state.ts
+++ b/backend/src/controllers/session-state.ts
@@ -33,6 +33,82 @@ interface CompactGates {
   signedAt?: string;
 }
 
+async function acceptPendingInvitationForE2ESession(
+  sessionId: string,
+  userId: string
+): Promise<boolean> {
+  if (process.env.E2E_AUTH_BYPASS !== 'true' || process.env.NODE_ENV === 'production') {
+    return false;
+  }
+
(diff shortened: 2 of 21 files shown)


slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary

Fixes Stage 4 strategy readiness sync issues found during Adam/Eve gold session testing. Adds fuzzy strategy deduplication (superseding stale drafts), a new `session.strategies_updated` realtime event, optimistic cache updates for mark-ready, a `strategy-readiness-pending` waiting state that hides input while waiting for the partner, consistent `readyAt` timestamps, and E2E auto-accept for invitee session URLs.

## Feedback

### Correctness — all good

- **Strategy dedupe** (`backend/src/utils/strategy-dedupe.ts`): The normalized word-overlap approach with a 0.65 threshold is a reasonable heuristic for catching refined-vs-draft supersession. Tests cover the key cases (superseded, distinct, exact dupe).
- **Optimistic updates** (`mobile/src/hooks/useStages.ts`): The `onMutate` → `onSuccess` → `onError` pattern follows the Cache-First rules correctly — cancels queries, snapshots previous state, rolls back on error, and reconciles with server data on success. Good adherence to the "never invalidateQueries on a key with optimistic updates in-flight" rule.
- **Recursive getSessionState** (`backend/src/controllers/session-state.ts:146`): Safe — the invitation is moved to `ACCEPTED` in the transaction, so the second call won't re-trigger `acceptPendingInvitationForE2ESession` (the `PENDING` filter fails). No infinite loop risk.
- **readyAt consistency** (`backend/src/controllers/stage4.ts:1019`): Good fix — the old code called `new Date().toISOString()` twice (gate vs response), potentially producing different timestamps.
- **Dual-source readiness derivation** (`mobile/src/hooks/useUnifiedSession.ts:553-555`): Defensive and correct — handles the case where one cache is updated before the other.

- **[Non-transactional delete+create]** `backend/src/controllers/messages.ts:2060-2073`: The `deleteMany` (superseded) and `createMany` (new strategies) aren't wrapped in a `$transaction`. If `createMany` fails after `deleteMany` succeeds, superseded strategies are lost with no replacements. Low risk in practice since this is deep in the streaming handler, but worth noting for future robustness.
- **[Dedup threshold edge case]** `backend/src/utils/strategy-dedupe.ts:24`: The `intersection / smallerSetSize >= 0.65` threshold could produce false positives for very short strategies that share common words (e.g., "Follow up after one month" vs "Check in after one month" → 4/5 overlap = 0.80). Worth monitoring in gold sessions.
- **[Dynamic import]** `backend/src/controllers/messages.ts:2076`: `await import('../services/realtime')` — presumably to avoid circular deps. Works but adds per-call overhead. Not a problem now; could be hoisted if strategy updates become high-frequency.
- **[Type safety]** `mobile/src/hooks/useStages.ts:1698-1713`: The `any` cast for progress cache update is pragmatic given `gatesSatisfied` is `Record<string, unknown>`, but it bypasses compile-time safety on the nested update.

### Security — clean

- E2E auto-accept is properly double-guarded: `E2E_AUTH_BYPASS !== 'true'` OR `NODE_ENV === 'production'` → returns false. The invitation query also prevents self-acceptance (`NOT: { invitedById: userId }`) and checks expiration.

### Tests — good coverage

- New: strategy dedupe (3 cases), getStrategies response shape, optimistic mark-ready cache, chatUIState hiding, getWaitingStatus readiness-pending.
- Existing stage4 test updated for new response fields.

### Architecture

- The `session.strategies_updated` event + push message + mobile handler is a clean addition to the existing realtime pattern.
- `strategy-readiness-pending` waiting status follows the established `waitingStatusConfig` pattern well.
- New shared DTO fields are optional, maintaining backward compat.

## Verdict

**Approve** — Well-structured fix for real gold-session blockers. The changes follow established patterns (Cache-First, optimistic updates, realtime events, waiting status config). Good test coverage. The minor observations above are non-blocking.

shantamg:
Addressed the two review observations that were worth fixing before merge:

- Wrapped superseded-strategy delete + replacement create in a Prisma transaction so a create failure cannot leave the inventory without the previous proposals.
- Removed the per-call dynamic import for publishSessionEvent; the controller already imports it statically.
- Tightened the optimistic progress-cache update type in mobile instead of using any.

Re-ran:
- npm run check
- npm -w backend test -- src/utils/__tests__/strategy-dedupe.test.ts src/routes/__tests__/stage4.test.ts --runInBand
- npm -w mobile test -- src/utils/__tests__/getWaitingStatus.test.ts src/utils/__tests__/chatUIState.test.ts src/hooks/__tests__/useStages.test.ts --runInBand --forceExit

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes Stage 4 strategy readiness sync: adds optimistic updates for mark-ready, a new `session.strategies_updated` realtime event, a `strategy-readiness-pending` waiting state, strategy deduplication logic to replace superseded proposals, and an E2E auto-accept path for invitees. Well-tested and follows project cache-first conventions.

## Feedback

- **Good pattern**: The `useMarkReadyToRank` optimistic update with `onMutate`/`onError` rollback follows the project's Rule 2 + Rule 6 correctly. The `setQueryData` → `refetchQueries` pattern in `partner.ready_to_rank` handler is also correct per Rule 7.

- **Good fix (stage4.ts:1016-1019)**: Extracting `readyAt` to a variable ensures the response timestamp matches what was persisted. Previously two `new Date()` calls could differ by milliseconds.

- **Minor observation (strategy-dedupe.ts:17)**: `isSupersededStrategy` uses `Math.min(existingWords.size, nextWords.size)` as the denominator. This makes the comparison symmetrical — a longer existing strategy would also be marked superseded if a shorter new proposal shares ≥65% of the *shorter* set's words. In practice the AI refines/expands (never shrinks), so this shouldn't trigger, but worth a comment noting the intended direction is "existing superseded by longer/refined next".

- **Architecture (session-state.ts:73-147)**: The E2E auto-accept guard (`E2E_AUTH_BYPASS !== 'true' || NODE_ENV === 'production'`) and the recursive `getSessionState` pattern are safe — the invitation status flip to `ACCEPTED` prevents infinite recursion, and the `return` after the recursive call prevents double-response. Clean approach.

- **Tests**: Good coverage of the dedupe utility, optimistic update cache behavior, and waiting-status derivation.

## Verdict

Approve — solid, well-tested fix for the Stage 4 readiness blocker documented in the gold session report. The strategy dedupe heuristic (65% word overlap) is a reasonable first pass; if edge cases appear in future gold runs they can be tuned.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes Stage 4 strategy readiness sync issues found during Adam/Eve gold-session E2E playthroughs. Adds strategy deduplication (superseded proposal cleanup), a `session.strategies_updated` realtime event, optimistic `useMarkReadyToRank` cache updates, a `strategy-readiness-pending` waiting state, and an E2E-only auto-accept path for invitee session URLs.

## Feedback

- **Good pattern (cache-first)**: The move from `invalidateQueries` → `refetchQueries` and the optimistic `onMutate`/`onError` rollback in `useMarkReadyToRank` aligns with the project's cache-first architecture. The `setQueryData` calls in event handlers for `partner.ready_to_rank` are also correct per CLAUDE.md rules.

- **Good fix (readyAt consistency)**: `stage4.ts:1019` — capturing `readyAt` once and reusing it for both the DB write and the response eliminates the subtle timestamp drift that the old dual `new Date().toISOString()` pattern had.

- **Minor observation (dedupe threshold symmetry)**: `strategy-dedupe.ts:20` — `isSupersededStrategy` uses `intersection / smallerSetSize >= 0.65`. Since it uses the *smaller* set as the denominator, this is effectively symmetric: a short new strategy whose words mostly appear in a longer existing one would also mark the existing as "superseded." In practice this seems fine because the AI typically refines shorter → longer, but worth noting if you ever see unexpected deletions of longer existing proposals when a short similar one arrives.

- **Minor observation (recursive getSessionState)**: `session-state.ts:156-159` — after E2E auto-accept, it recurses into `getSessionState(req, res)`. This is safe (can only happen once since the session will be found on the second call), but a comment noting "safe single recursion" could help future readers.

- **Tests**: Good coverage — the dedupe utility, optimistic hook update, chatUIState, and getWaitingStatus all have targeted tests for the new behavior.

## Verdict

**Approve** — Clean, well-tested fix for a real Stage 4 blocker. The changes follow established patterns, the E2E-only code path has proper guards, and the documentation of issues found during testing is thorough.

PR #357, opened 2026-05-05
[codex] Fix persisted message refresh and chat gating

shantamg:
## Summary

Fixes the remaining Adam/Eve Stage 1 and Stage 3 stale-session issues by making persisted facilitator/message refreshes explicit and by tightening chat input ownership around gate, wait, and review states.

## What changed

- Added a shared persisted-message refresh key helper covering `messageKeys.infinite`, `messageKeys.list`, `timelineKeys.infinite`, and `sessionKeys.state`.
- Refetched persisted messages for Stage 3 realtime events that can produce facilitator/system messages or gate transitions.
- Reconciled SSE completion against persisted server messages immediately, with a short delayed fallback for backend follow-up messages.
- Made derived chat UI state hide freeform input while required panels or Stage 4 review/ranking/agreement flows own the next action.
- Added unit coverage for realtime invalidation keys, Stage 1/3 wait derivation, and authoritative input hiding.

## Root cause

Some realtime paths refreshed stage-specific state but only invalidated/refetched the infinite message cache, leaving list/timeline/session-derived surfaces stale until reload. Separately, the screen had an exception that kept chat input visible during required review panels.

## Validation

- `npm --workspace e2e run check`
- `npm --workspace mobile test -- --runTestsByPath src/utils/__tests__/chatUIState.test.ts src/utils/__tests__/getWaitingStatus.test.ts src/utils/__tests__/realtimeInvalidation.test.ts`

Attempted relevant Playwright E2E specs, but the default config was blocked because `localhost:3000/health` was already occupied while `reuseExistingServer` is false.

### DIFF
--- a/mobile/src/config/waitingStatusConfig.ts
+++ b/mobile/src/config/waitingStatusConfig.ts
@@ -295,7 +295,7 @@ export const WAITING_STATUS_CONFIG: Record<NonNullable<WaitingStatusState>, Wait
   // Stage 4: All agreements confirmed by me, waiting for partner
   'agreement-pending': {
     showBanner: true,
-    hideInput: false,
+    hideInput: true,
     showInnerThoughts: false,
     isActionRequired: false,
     showSpinner: false,
--- a/mobile/src/hooks/useStreamingMessage.ts
+++ b/mobile/src/hooks/useStreamingMessage.ts
@@ -18,6 +18,7 @@ import {
   CapturedNeedInput,
 } from '@meet-without-fear/shared';
 import { messageKeys, sessionKeys, stageKeys, timelineKeys } from './queryKeys';
+import { getPersistedMessageRefreshQueryKeys } from '../utils/realtimeInvalidation';
 
 // ============================================================================
 // Types
@@ -155,6 +156,7 @@ export function useStreamingMessage(
   // Ref for 15-second fallback timer to handle stuck connections
   const fallbackTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const FALLBACK_TIMEOUT = 15000; // 15 seconds
+  const reconciliationTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   /**
    * Add a message to the cache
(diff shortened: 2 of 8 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes stale UI after realtime events by centralizing persisted message cache invalidation, and makes input hiding consistent by moving all gating logic into the pure `computeShouldHideInput` function.

## Feedback

- **Architecture — Nice**: `getPersistedMessageRefreshQueryKeys()` is a clean extraction. Previously `messageKeys.list` and `timelineKeys.infinite` were missing from stage invalidation sets, causing stale messages after needs/empathy events. This fixes it and prevents future drift.

- **Correctness — Good cleanup**: Removing the `!shouldShowEmpathyPanel && derivedShouldHideInput` override in `UnifiedSessionScreen.tsx:2947` and letting the pure function handle it via `aboveInputPanel === 'empathy-statement'` is the right move. Single source of truth for input visibility.

- **Minor — Double invalidate+refetch in `reconcilePersistedMessages`** (`useStreamingMessage.ts:393-399`): Calling both `invalidateQueries` and `refetchQueries` on the same key means active queries get two network requests (one from invalidation triggering observers, one from the explicit refetch). This is likely intentional as a belt-and-suspenders measure for queries that may not be actively observed, but worth noting. Not blocking.

- **Stage 4 input hiding** (`chatUIState.ts:625-630`): The new `agreements.length > 0` check fires regardless of `strategyPhase`, meaning even during `COLLECTING` if agreements exist from a prior negotiation round, input hides. This seems correct for the agreement-review flow, but worth a sanity check that there's no COLLECTING sub-phase where the user should still be able to chat while prior agreements exist.

- **Tests — Thorough**: Good coverage of the new panel-based input hiding (share-suggestion, empathy-statement, feel-heard, needs-reveal-validation, Stage 4 agreements) plus the persisted message refresh keys. The `getWaitingStatus` tests for the needs-pending/needs-not-confirmed edge cases are a nice addition.

## Verdict

Approve — well-structured fix with good test coverage. The centralization of cache invalidation keys and consistent input-hiding logic are solid improvements.

PR #358, opened 2026-05-05
[codex] Fix gold session chat and needs UI

shantamg:
## Summary

This PR packages the gold-session fixes from the Adam/Eve run:

- Preserve whitespace boundaries while scrubbing visible AI text during streaming, so protected tags do not collapse adjacent words.
- Keep the Stage 3 chat composer primary by moving the needs review affordance below the input.
- Collapse the below-input needs affordance to `What Matters`, a captured-needs count, and a `Review` action.
- Clean up the needs drawer copy/actions: remove duplicate `Your Needs`, use `Chat to refine`, and update the side-by-side validation copy.
- Hide the Stage 4 `Ideas So Far` preview until at least one strategy exists.

## Why

The gold-session run exposed a few user-facing issues: AI responses had missing spaces, Stage 3 needs review crowded the active chat area, the drawer repeated labels and had unclear action copy, and Stage 4 showed an empty ideas panel before ideas existed.

## Validation

- `backend`: `npm test -- --runTestsByPath src/utils/__tests__/visible-text.test.ts src/routes/__tests__/messages.test.ts --runInBand`
- `backend`: `npm run check`
- `mobile`: `npm test -- --runTestsByPath src/utils/__tests__/chatUIState.test.ts src/components/__tests__/NeedsDrawer.test.tsx --runInBand`
- `mobile`: `npm run check`

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -110,7 +110,10 @@ const PLANNER_LINE_PREFIXES = [
   'i need to make sure both lists',
 ];
 
-export function scrubVisibleAIText(text: string): { text: string; scrubbed: boolean } {
+export function scrubVisibleAIText(
+  text: string,
+  options: { preserveBoundaryWhitespace?: boolean } = {}
+): { text: string; scrubbed: boolean } {
   const before = text;
   const plannerScrubbed = text
     .split(/\r?\n/)
@@ -121,7 +124,9 @@ export function scrubVisibleAIText(text: string): { text: string; scrubbed: bool
     .join('\n')
     .replace(/\bI should\b/gi, '')
     .replace(/\bso both lists should be available\b/gi, '');
-  const cleaned = cleanVisibleAIText(plannerScrubbed);
+  const cleaned = cleanVisibleAIText(plannerScrubbed, {
+    preserveBoundaryWhitespace: options.preserveBoundaryWhitespace,
+  });
 
   return { text: cleaned, scrubbed: cleaned !== before };
 }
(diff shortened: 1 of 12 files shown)


slam-paws (commented):
## PR Review: #358 — Fix gold session chat and needs UI

**Verdict: Approve with minor observations**

---

### Summary

This PR fixes two distinct problems surfaced during gold-session testing:

1. **Backend — Streaming whitespace preservation**: `cleanVisibleAIText` gained a `preserveBoundaryWhitespace` option so that streamed chunks don't collapse words at token boundaries. The `.map(cleanVisibleAIText)` call sites were also wrapped in lambdas to prevent `.map`'s `index` arg from being interpreted as the new `options` parameter — good type-safety fix.

2. **Mobile — Needs review UX overhaul**: The needs-review / needs-share panels moved from blocking above-input panels (which hid the text input) to a persistent compact card rendered *below* the input. This keeps the chat input visible so users can continue typing while the needs affordance is present.

---

### What looks good

- **Whitespace logic is clean**: Captures leading/trailing whitespace before the cleaning loop, re-applies after. The early return for whitespace-only chunks is correct.
- **`.map` callback wrapping** (`needs.ts`, `micro-tag-parser.ts`): Necessary now that the function signature has an optional second arg. TypeScript would catch this at build, but the explicit wrapper is the right fix.
- **`confirmFeelHeard`**: Upgrading from `.trim()` to `cleanVisibleAIText()` on the parsed AI response is a net improvement — catches leaked markdown/serialization artifacts, not just whitespace.
- **Strategy pool empty-state guard**: Double-guarded (`strategies.length > 0` in hook + `stratCount === 0` early-return in render) — belt and suspenders, no issue.
- **`onAdjustNeeds` behavior change**: Closing the drawer instead of force-sending "I would like to adjust my identified needs" is better UX — lets the user phrase the adjustment themselves. Button rename to "Chat to refine" communicates this clearly.
- **NeedsDrawer button reorder**: Confirm first, adjust second — follows primary-action-first convention.

1. **`needsSummaryActionCompact` style** (`UnifiedSessionScreen.tsx:3909–3920`): This style combines View-level props (`paddingVertical`, `borderRadius`, `backgroundColor`, `borderWidth`) with Text-level props (`fontSize`, `fontWeight`, `color`) on a `<Text>` component. It works in React Native, but mixing layout/decoration styles on Text can behave differently across platforms (especially Android). If you see rendering inconsistencies, consider wrapping the Text in a `<View>` with the border/background styles.

2. **Animation `maxHeight` jump (100 → 420)**: The old above-input panel used `maxHeight: 100`; the new below-input compact card uses `maxHeight: 420`. The compact card is likely well under 420px, but this is a generous ceiling. Not a problem — just noting that if the card content ever expands beyond expectations, it won't clip early.

3. **Needs review cards suppressed during panel display** (`needsReviewCards` memo): The new guard `if (aboveInputPanel === 'needs-review' || aboveInputPanel === 'needs-share') return [];` prevents duplicate affordances (compact card below input + card in chat stream). This is correct — just want to confirm the intent is that the *inline* chat card reappears once `aboveInputPanel` clears (e.g., after confirmation). From the logic, this looks right.

---

### Tests

- New tests for `preserveBoundaryWhitespace` in both `visible-text.test.ts` and `messages.test.ts` ✅
- `NeedsDrawer.test.tsx` updated for copy change ✅

LGTM. 

slam-paws (approved):
Approving — clean fix for streaming whitespace + improved needs review UX. Observations above are non-blocking.

PR #341, opened 2026-05-04
Fix masked error reporting and Clerk rate-limit handling

slam-paws:
## Changes
- Fix: Replace static `[Error]` log tag with `[err.name] err.message` in global error handler, producing descriptive Sentry titles instead of masked ones
- Fix: Catch Clerk API failures (429 rate-limiting) in auth middleware slow path; retry DB lookup since another concurrent request may have already created the user

Related to #339

## Provenance
- **Channel:** GitHub issue
- **Requested by:** daily-strategy bot (auto-filed)
- **Original message:** Bug: Untagged backend error "Error: [Error]" — 84 occurrences
- **Prompt(s) used:** general-pr workspace

## Docs updated
No doc changes needed — error handling/logging fix with no architectural impact

Issue #323, opened 2026-05-04
Bug: Sentry regression — 210 'Access to this session denied' on /agreements starting 22:02 UTC (post-PR #319 merge)


slam-paws:
## Summary

New Sentry error variant detected in the evening strategy briefing scan. Distinct from the earlier /empathy/status spike tracked in #314.

**Error:** `Access to this session denied`
**Endpoint:** `GET /api/sessions/:id/agreements`
**Sentry ID:** 7457481620
**Count:** 210 occurrences
**First seen:** 2026-05-03T22:02 UTC
**Last seen:** 2026-05-04T00:24 UTC
**Correlation:** Started ~1.5h after PR #319 (invitee onboarding flow + invitation-readiness timing) was merged at 20:43 UTC

## Hypothesis

The invitation-readiness timing changes in PR #319 may have introduced a window where invitees attempt to access agreement endpoints before session ownership/access is established.

## Next Steps

- [ ] Review PR #319 changes to sessions/agreements access guard logic
- [ ] Check if invitee session token is being validated against correct session ownership
- [ ] Check Render logs for the 22:02 UTC window
- [ ] Confirm whether errors are from invitees or session owners

slam-paws:
## Investigation Report

**Root cause**: The `requireSessionAccess` middleware (`backend/src/middleware/auth.ts:294-333`) only checks `RelationshipMember` records. When an invitee accepts an invitation, the `RelationshipMember` is created inside a transaction in `acceptInvitation`. After acceptance, the app navigates to the session screen, where `useUnifiedSession` eagerly fires all queries — including `useAgreements(sessionId)` (line 289: "Stage 4: Strategies — always fetch to avoid waterfall"). If any of these requests hit the server before the `RelationshipMember` is fully visible, `requireSessionAccess` throws "Access to this session denied".

PR #319 made this spike occur because:
1. `confirmTopicFrame` now eagerly sets `invitation.messageConfirmed = true` + `session.status = 'INVITED'` (lines 78-93 of `topic-frame.ts`)
2. This allows invitees to accept the invitation *immediately* (as soon as the topic is confirmed), rather than waiting for the inviter to close the share modal
3. More successful early acceptances → more invitees hit the race condition in `requireSessionAccess`

**Evidence**:
- Error source: `requireSessionAccess` middleware at line 326 is the only place that throws "Access to this session denied"
[...]

PR #321, opened 2026-05-04
fix(auth): allow invitees session access during acceptance timing gap

slam-paws:
## Changes
- Fix: Add invitation-based fallback in `requireSessionAccess` middleware — if user is not a RelationshipMember, allow access when the session has a valid pending/accepted invitation and the user is not the inviter
- Fix: Filter `development` environment events in Sentry `beforeSend` hook to prevent local dev noise from polluting production monitoring

Related to #314

## Provenance
- **Channel:** GitHub issue
- **Requested by:** Daily strategy briefing (Slam Paws)
- **Original message:** Sentry spike of 210 'Access to this session denied' errors — investigation found invitee session-access timing gap during invitation acceptance
- **Prompt(s) used:** general-pr workspace, 01-implement stage

## Root cause
The `requireSessionAccess` middleware only checked `RelationshipMember` records. During the invitation acceptance flow, the invitee receives a `sessionId` from invitation details but is only added as a `RelationshipMember` inside the `acceptInvitation` transaction. Between receiving the sessionId and the transaction committing, all session endpoint requests fail with 403.

The fallback check allows access when:
1. The session has a non-expired `PENDING` invitation, OR an `ACCEPTED` invitation
2. The requesting user is **not** the inviter (they're the invitee candidate)

Session IDs are unguessable CUIDs, so this maintains the security boundary.

## Docs updated
No doc changes needed — implementation-level auth fix, architecture docs don't describe session access check internals.

PR #360, opened 2026-05-05
fix(sessions): 404 for missing sessions, stop client retry flood (#323)

shantamg:
## Summary
- Backend `requireSessionAccess` now throws `NotFoundError` (404) when the session row doesn't exist, instead of `ForbiddenError` (403). PR #321's invitee fallback is preserved for the real "exists but no access" case.
- Mobile `QueryClient` defaults: never retry on 401/403/404 — these are terminal client errors, retrying just floods Sentry.
- Mobile `useUnifiedSession`: gate all 8 child queries off when `/state` returns 403 **or** 404 (was 403 only). Removed the now-redundant per-call 403-retry override.

## Why
Sentry issue [MWF-BACKEND-12](https://meet-without-fear.sentry.io/issues/7457481620/) — 880 events labeled "Access to this session denied" since May 3. The bot's earlier patches (#321 invitee fallback, #341 error renaming) held; they just weren't the right fix.

Investigation: every one of the 880 events was for the **same** session id `cmokigesd000dma1wsv7wink1`, polling 9 different session endpoints. The session does not exist in the DB. With `retry: 3` + exp backoff + `refetchOnWindowFocus`, a single client with a stale session id produced ~36 errors per app focus across all the stage-gated endpoints. The auth middleware was 403-ing missing sessions as if they were access denials.

## Changes
- `backend/src/middleware/auth.ts` — add `NotFoundError` path before invitation fallback
- `backend/src/middleware/__tests__/auth.test.ts` — 4 new unit tests (member / not-found / forbidden / invitee-fallback). All 21 auth tests pass.
- `mobile/src/providers/QueryProvider.tsx` — function-form `retry` that short-circuits on 401/403/404
- `mobile/src/hooks/useUnifiedSession.ts` — `accessDenied` covers 404; drop redundant local override

## Test plan
- [x] `npm run check` passes
- [x] `npm -w @meet-without-fear/backend test -- --testPathPattern=auth` — 21/21 pass
- [ ] Manual: open mobile app with a deleted/stale session id in cache → expect one 404 on `/state`, no waterfall of empathy/needs/strategies errors
- [ ] Sentry: confirm MWF-BACKEND-12 stops growing within 24h of deploy

## Notes
Pre-existing failures on main (circuit-breaker timing tests, 7 mobile suites) are unrelated and reproduce without these changes.

### DIFF
--- a/backend/src/middleware/__tests__/auth.test.ts
+++ b/backend/src/middleware/__tests__/auth.test.ts
@@ -5,7 +5,8 @@
  */
 
 import { Request, Response, NextFunction } from 'express';
-import { requireAuth, optionalAuth, getUser } from '../auth';
+import { requireAuth, optionalAuth, getUser, requireSessionAccess } from '../auth';
+import { ForbiddenError, NotFoundError } from '../errors';
 import { prisma } from '../../lib/prisma';
 
 // Mock Prisma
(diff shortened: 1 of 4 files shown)


slam-paws (dismissed):
## Review: APPROVE

Clean, well-scoped fix that addresses the Sentry flood from stale session polling. The approach is solid across all three layers:

### Backend (`auth.ts`)
- Correctly inserts the existence check **after** the membership lookup but **before** the invitation fallback, so the happy path (valid member) incurs no extra query.
- The `findUnique` with `select: { id: true }` is appropriately lightweight.
- `NotFoundError` is already handled by the `errorHandler` middleware via the `AppError` base class — no additional wiring needed.

### Mobile (`useUnifiedSession.ts`)
- Adding `404` to `accessDenied` correctly prevents child queries from firing on deleted/missing sessions, same as the existing 403 behavior.
- Removing the per-query retry override is clean since it's now handled globally.

### Mobile (`QueryProvider.tsx`)
- Moving the 401/403/404 no-retry logic to the global `QueryClient` default is the right pattern — terminal client errors should never be retried regardless of which query triggers them.

### Tests
- Four tests cover all branches: member access, 404 (missing), 403 (exists but no access), and invitation fallback. Good coverage.

### Nit (pre-existing, non-blocking)
`NotFoundError('Session not found')` will produce the message *"Session not found not found"* because the constructor signature is `(resource, message?)` and defaults to `` `${resource} not found` ``. This is a pre-existing pattern used in other controllers (`distillation.ts`, `meditation.ts`, `takeaways.ts`) — not introduced here, but worth a follow-up to normalize these to `new NotFoundError('Session')`.

LGTM — ship it.

slam-paws (approved):
## Review: PR #360 — 404 for missing sessions, stop client retry flood

**Verdict: Approve**

Clean, well-structured fix that addresses a real Sentry noise problem. Three layers working together:

### Backend (`auth.ts`)
The new `findUnique` check between the membership query and the invitation fallback correctly distinguishes "session doesn't exist" (404) from "user lacks access" (403). Placement before the invitation fallback is right — no point checking invitations for a nonexistent session. The extra DB query only runs on the failure path, so no perf impact on happy-path requests.

### Mobile — Global retry policy (`QueryProvider.tsx`)
`shouldRetryApiError` centralizes the "never retry terminal client errors" logic for both queries and mutations. Adding 401 here is a nice bonus — previously only 403 was handled, and only in one hook.

### Mobile — `useUnifiedSession.ts`
Removing the per-hook retry override in favor of the global policy is the right call. The `accessDenied` guard correctly expands to include 404, which stops child queries from firing after a deleted session is detected.

### Tests
All four `requireSessionAccess` scenarios are covered: member access, missing session (404), no-access (403), and invitation fallback. Mock setup correctly chains `findFirst` calls for the two-step lookup.

No issues found. Ship it.

PR #439, opened 2026-05-06
[codex] Fix Stage 2 calibration gates

shantamg:
## Summary

Fixes the Stage 2 calibration blockers found after #433:

- hold Stage 2 share suggestions until the subject has submitted their own empathy draft
- keep the chat input visible when a share suggestion is present
- scope the felt-heard invariant to Stage 1 transcript slices so Stage 2 retrospective language does not false-fail the gate

## Validation

- `python3 -m py_compile scripts/mwf_gold_loop.py`
- `npm --workspace mobile test -- --runInBand mobile/src/utils/__tests__/getWaitingStatus.test.ts mobile/src/utils/__tests__/chatUIState.test.ts`
- bounded Adam/Eve Stage 2 calibration: `eval/runs/20260506-115959-adam-eve-loop-summary.md`, score `3.5`, hard invariants pass
- bounded James/Catherine Stage 2 calibration: `eval/runs/20260506-121950-james-catherine-loop-summary.md`, score `3.5`, hard invariants pass

## Notes

The original product/eval blockers did not reproduce after this patch. The remaining blocker before restarting unattended self-improvement is the James/Catherine fixture/start-state issue: the calibration run started with James as initiator, while the gold scenario expects Catherine to initiate.


slam-paws (approved):
## PR Review: Fix Stage 2 calibration gates

**Overall: Approve** — Well-structured fix with good test coverage for both the mobile UI gating and the gold loop role-shape inference.

### Summary

Two independent improvements:

1. **Mobile UI**: Gates the share-suggestion panel behind the subject's own empathy completion, preventing premature interruption during perspective-taking. Also stops hiding the text input when the share-suggestion panel is active.

2. **Gold loop**: Infers initiator/invited roles from gold profile Stage 0 evidence (regex on "message for X" / "thank you, X" patterns) and uses that to determine session start character, plus a new invariant check. Also scopes the felt-heard gate check to Stage 1 transcripts only.

### Feedback

**Question — Consent check divergence between `chatUIState.ts` and `getWaitingStatus.ts`**

In `chatUIState.ts:356`, the gate is:
```ts
const ownEmpathySubmitted = empathyAlreadyConsented || empathyDraft?.alreadyConsented === true;
```

In `getWaitingStatus.ts:178`, it's:
```ts
empathyDraft?.alreadyConsented
```

This is because `WaitingStatusInputs` lacks the top-level `empathyAlreadyConsented` field (only `ChatUIStateInputs` adds it). But it means there's a potential edge case: if `empathyAlreadyConsented` is `true` while `empathyDraft?.alreadyConsented` is still `false`/`undefined`, the share-suggestion panel would be shown (via chatUIState), but `computeWaitingStatus` wouldn't return `awaiting-context-share`. Is that divergence intentional, or should `WaitingStatusInputs` also carry the top-level flag?

**Nit — Regex pattern duplication across `mwf_gold_profile.py` and `mwf_gold_loop.py`**

The invited/initiator regex patterns (`message for X`, `thank you, X`, etc.) are duplicated between `infer_role_shape` in `mwf_gold_profile.py` and `infer_role_shape_from_profile` in `mwf_gold_loop.py`. Not blocking, but if these patterns evolve they'd need to be updated in two places. A shared constant or small utility could help.

**Nit — Long line in `run_iteration`**

`scripts/mwf_gold_loop.py` line ~2828:
```python
start_partner_character = next(character for character in (first_character, second_character) if character != start_character)
```
This is fine functionally but exceeds typical line length.

### Checks

- Tests cover the new gating (both positive/negative cases) ✓
- Invariant check handles seeded/restored sessions gracefully (skips with soft pass) ✓
- `transcript_metadata` is already defined in the file and returns stage info ✓
- `check_felt_heard_gate_after_witnessing` Stage 1 filter is correct ✓

PR #440, opened 2026-05-06
[codex] Fix gold loop seeding and chat panel input flow

shantamg:
## Summary
- Fix fresh E2E gold-loop session seeding so both participants are relationship members and both receive user vessels, including CREATED sessions.
- Keep conversational guided panels attached to the chat flow by rendering them below the input and keeping input visible for feel-heard, topic proposal, empathy review, and needs reveal validation panels.
- Tighten gold-loop evaluation routing with an automatic invariant for copy that promises continued chat without visible input, plus not-evaluable prompt-quality scope when orchestration/access/transcript failures block the run.
- Update Stage 0/1/2 prompt guidance so copy does not invite freeform chat unless the input remains visible.

## Root Cause
- The fresh-session seed path only attached the second participant for later active stages, which left Catherine unauthorized in CREATED James/Catherine runs.
- The chat UI hid input for panels whose copy still implied the user could continue talking.
- The browser loop was allowing orchestration/access/UI-state failures to contaminate prompt-quality evaluation.

## Validation
- `npm --workspace backend test -- --runTestsByPath src/testing/__tests__/state-factory.test.ts src/services/__tests__/stage-prompts.test.ts`
- `npm --workspace mobile test -- --runTestsByPath src/utils/__tests__/chatUIState.test.ts src/components/__tests__/ChatInterface.test.tsx`
- `python3 -m py_compile scripts/mwf_gold_loop.py`

## Notes
- Left unrelated local retention/render edits and raw run artifacts out of this PR.

### DIFF
--- a/backend/src/services/__tests__/stage-prompts.test.ts
+++ b/backend/src/services/__tests__/stage-prompts.test.ts
@@ -699,6 +699,7 @@ describe('Stage Prompts Service', () => {
       // The flag instruction is in format "FeelHeardCheck: [Y if ready..., N otherwise]"
       expect(prompt).toContain('FeelHeardCheck:');
       expect(prompt).toMatch(/FeelHeardCheck.*Y.*N/s);
+      expect(prompt).toContain('Do NOT invite more freeform chat unless the input remains visible');
     });
 
     it('Stage 2 protocol includes ReadyShare flag instruction', () => {
@@ -709,6 +710,7 @@ describe('Stage Prompts Service', () => {
       // The flag instruction is in format "ReadyShare: [Y if ready..., N otherwise]"
       expect(prompt).toContain('ReadyShare:');
       expect(prompt).toMatch(/ReadyShare.*Y.*N/s);
+      expect(prompt).toContain('Do NOT invite more freeform chat unless the input remains visible');
     });
 
     it('Stage 0 protocol instructs the AI to emit a topic via <draft> tags', () => {
@@ -718,6 +720,7 @@ describe('Stage Prompts Service', () => {
 
       // Stage 0 now emits the proposed TOPIC inline as <draft> (not an invitation message).
       expect(prompt).toContain('<draft>');
+      expect(prompt).toContain('without inviting freeform chat unless the input remains visible');
     });
 
     it('protocol includes dispatch tag instruction', () => {
--- a/backend/src/services/stage-prompts.ts
+++ b/backend/src/services/stage-prompts.ts
@@ -489,7 +489,7 @@ LISTENING: Ask warm, specific questions that surface what the situation is about
 
 Keep responses short (1-2 sentences) until you have enough to propose a topic.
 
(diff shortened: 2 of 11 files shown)


slam-paws (approved):
## PR Review: [codex] Fix gold loop seeding and chat panel input flow

### Summary

Three cohesive changes working together:

1. **Keep chat input visible alongside guided panels** — Previously the input was hidden when feel-heard, empathy-statement, topic-proposal, or needs-reveal-validation panels appeared. Now only `invitation` hides input. This lets users keep typing while guided affordances are offered.

2. **Move guided panels below the chat input** — `renderAboveInput` callback now renders *after* `ChatInput` in both `ChatInterface` and `ChatTimeline`, so the transcript-to-input connection stays unbroken.

3. **Update AI prompts to match** — Stage 0/1/2 prompts now say "Do NOT invite more freeform chat unless the input remains visible" instead of encouraging users to keep talking. This prevents the AI from promising chat when the input might not be there.

4. **Fix gold loop seeding** — `create_gold_session` now calls `/api/e2e/seed-session` directly instead of shelling out to `create_gold_session.sh`. `StateFactory` attaches User B to CREATED-stage sessions so both E2E URLs are authorized. New `check_chat_copy_has_visible_input` invariant catches prompt copy that promises chat without a visible textbox. Scoring now distinguishes infra failures from prompt-quality issues via `not_evaluable_for_prompt_quality`.

### Verdict

**Approve** — Well-structured, well-tested, all three themes reinforce each other. A few minor nits below.

### Nits / Questions

1. **Prop name `renderAboveInput` is now misleading** — After this change, the callback renders *below* the input. The JSDoc was updated but the prop name itself could confuse future contributors. Consider renaming to `renderBelowInput` or `renderGuidedPanel` in a follow-up. (Not blocking — the existing `renderBelowInput` prop makes a direct rename ambiguous, so this would need a small design decision.)

2. **Missing test for `topic-proposal` input visibility** — The PR adds/updates tests for `feel-heard`, `empathy-statement`, and `needs-reveal-validation` now keeping input visible, but `topic-proposal` was also removed from the hide list and has no explicit `shouldHideInput` test in `chatUIState.test.ts`. A small gap — consider adding one for completeness.

3. **`CHAT_PROMISE_RE` coverage** — The regex catches common phrases like "keep talking", "keep exploring", "I'm still here". The Stage 2 prompts previously said "or we can keep talking" which this catches. Worth verifying the regex also catches any other phrasings the model might generate (e.g., "feel free to share more", "there's no rush"). Not blocking — the invariant can be tightened iteratively.

4. **`StateFactory` — `TWO_USER_ACTIVE_TARGETS` partially bypassed** — Two of the four `TWO_USER_ACTIVE_TARGETS.has()` guards were removed (membership creation + UserVessel creation), but two remain (invitation status + one in the else branch). The logic is correct — just noting that the remaining guards still serve a purpose for invitation flow control, which isn't obvious at first glance. A one-line comment clarifying this distinction would help.

### Code Quality
- Tests are thorough: new state-factory unit test, updated chatUIState tests, ChatInterface ordering test, prompt content assertions
- Prompt changes are minimal and precise — the "unless the input remains visible" escape hatch is smart for forward-compatibility
- Gold loop scoring improvements (not_evaluable_for_prompt_quality) prevent false negatives from infra failures

PR #442, opened 2026-05-07
[codex] Pass bounded gold loop calibration

shantamg:
## Summary

This PR fixes the deterministic gold-loop blockers and prompt pacing issues found while calibrating the bounded Adam/Eve and James/Catherine loops.

Changes include:
- strip XML-style visible control tags such as `<feel_heard_check>Y</feel_heard_check>` and `<ready-share>Y</ready-share>` before rendering
- make gold-loop service startup default to real LLM (`MOCK_LLM=false`) while preserving explicit overrides
- tighten transcript/invariant checks for visible control tags and labeled shared-context blocks
- avoid false failing chat-copy/input checks when a valid user follow-up turn exists
- adjust Stage 1 and Stage 2 prompt guidance for long-running/high-conflict cases
- remove repair/reconciliation pressure from Stage 2 transition and purpose copy
- prevent Stage 2 post-share continuation from restarting partner-perspective work
- add focused backend and Python regression tests

## Validation

Focused tests:

```bash
npm --workspace backend test -- --runTestsByPath src/utils/__tests__/micro-tag-parser.test.ts src/services/__tests__/stage-prompts.test.ts src/controllers/__tests__/stage2-copy.test.ts src/services/__tests__/reconciler.test.ts src/routes/__tests__/reconciler.test.ts src/routes/__tests__/stage2.test.ts src/routes/__tests__/messages.test.ts
python3 scripts/test_mwf_moment_eval.py
```

Results:
- Jest: 7 suites passed, 242 tests passed
- Python: 65 tests passed

Bounded real-LLM gold-loop validation:
- Adam/Eve: `eval/runs/20260506-173725-adam-eve-iter-01`, score `4.0`, verdict `eval_warn`, hard invariants pass
- James/Catherine: `eval/runs/20260506-175333-james-catherine-iter-01`, score `4.0`, verdict `eval_warn`, hard invariants pass

## Notes

Raw run artifacts, scratch logs, and extracted transcripts were intentionally left out of this PR. They remain local evidence rather than committed fixtures.

Follow-up recommended but not blocking this pass:
- Stage 2 post-empathy validation/action boundary can still be made clearer when partner validation copy appears before the exact validation controls are visible.
- Stage 1 calibration can be refined further to avoid reflecting one participant's narrative as product truth too strongly.

@slam-paws please review.

### DIFF
--- a/backend/scripts/extract-session-transcripts.ts
+++ b/backend/scripts/extract-session-transcripts.ts
@@ -414,6 +414,9 @@ function formatTranscriptMarkdown(
 
       case 'shared_context':
         md += `---\n`;
+        if (event.role) {
+          md += `**${event.role}:**\n`;
+        }
         md += `${event.content}\n`;
         md += `*${time}*\n`;
         md += `---\n\n`;
--- a/backend/src/controllers/__tests__/stage2-copy.test.ts
+++ b/backend/src/controllers/__tests__/stage2-copy.test.ts
@@ -10,4 +10,23 @@ describe('Stage 2 user-facing copy', () => {
 
     expect(source.toLowerCase()).not.toContain('internal reconciler');
   });
+
+  it('does not frame Stage 2 empathy as repair or working things out', () => {
+    const files = [
+      path.join(__dirname, '..', 'stage2.ts'),
+      path.join(__dirname, '..', 'messages.ts'),
+      path.join(__dirname, '..', '..', 'services', 'dispatch-handler.ts'),
+    ];
+
+    const sources = files.map((file) => fs.readFileSync(file, 'utf8').toLowerCase());
+
+    for (const source of sources) {
+      expect(source).not.toContain('strongest things you can do to actually work things out');
+      expect(source).not.toContain('single strongest predictor of working things out');
+      expect(source).not.toContain('helps people work things out');
+      expect(source).not.toContain('before moving forward together');
+      expect(source).not.toContain("you'll move forward together");
+      expect(source).not.toContain("you'll be ready to move forward together");
+    }
+  });
 });
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -97,6 +97,42 @@ function getFallbackInitialMessage(
 const STAGE2_ROADMAP_COPY =
   `Here's what comes next: there are a few more steps in this process. First, each of you tries to understand what the other person might be going through. Then you'll each explore what matters most to you, and eventually use that to get clearer about what is possible next, whether together or separately.`;
 
+function buildStage2TransitionFallback(
+  partnerName: string | undefined,
+  partnerStatus: 'not_joined' | 'in_progress' | 'completed'
+): string {
+  const partner = partnerName || 'your partner';
+  const mutualPhrase = partnerStatus === 'not_joined'
+    ? `${partnerName || 'Your partner'} will be doing this same thing for you on their side.`
+    : `${partnerName || 'Your partner'} is doing this same thing for you on their side.`;
+
+  return `What you just did really mattered — sharing what's been weighing on you and staying with it until you felt heard takes real honesty.\n\n${STAGE2_ROADMAP_COPY}\n\nThis next part might feel a little unusual — I'm going to ask you to try to imagine what ${partner} might be experiencing, even though you might still be upset with them. I know that's a strange ask. This is not about excusing, agreeing, or deciding what happens next. It's a protected attempt to see whether you can name something real about ${partner}'s inner experience while keeping your own truth intact. You don't have to get it right — it's a guess, not a test. ${mutualPhrase}\n\nSo — what do you think might be going on for ${partner} in all of this?`;
+}
+
+function scrubStage2RepairFraming(content: string): string {
+  const researchToRepairRegex = new RegExp(
+    [
+      "there is a lot of research|there's a lot of research|research shows",
+      'genuinely tr(?:y|ies)ing to see',
+      'work things out',
+    ].join('[^.]*'),
+    'gi'
+  );
+  const moveForwardTogetherRegex = new RegExp(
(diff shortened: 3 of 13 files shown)


slam-paws:
Looking into this — reviewing now.

slam-paws (approved):
## Review: Pass bounded gold loop calibration

Thorough review of all 13 changed files. This is a well-scoped PR that addresses two intertwined problems: (1) visible control tags leaking into rendered output, and (2) repair/reconciliation framing bleeding into Stage 2 transitions and purpose copy. Both are real blockers for calibration fidelity.

### What looks good

**Micro-tag parser** — `extractBooleanControlTag` and `stripKnownControlTags` are clean. The regex handles underscore/dash variants correctly (`[_-]?`), stripping happens at the right point in the pipeline (after major tag removal, before response finalization), and the `??` fallback to thinking-block flags preserves backward compatibility. Tests cover both tag variants.

**Stage 1 high-conflict guidance** — The new `HIGH-CONFLICT / LONG-RUNNING CASES` block is well-placed and addresses a real gap. Requiring the AI to hear emotional cost, prior attempts, and boundary/grief before `FeelHeardCheck:Y` is the right move for complex cases. The feel-heard check criteria update (adding condition 4) is consistent.

**ReadyShare gating** — Bumping the early guard to `turnCount < 5` and requiring "at least 4 substantive Stage 2 turns" with the `isRefiningEmpathy` bypass is sound. The refinement mode test at the bottom of `stage-prompts.test.ts` covers the bypass path. The resistance-detection language in the prompt ("unfairness, anger, fear, resentment") and the explicit checkpoint ("Does that feel like your real attempt") are good calibration guardrails.

**Stage 2 copy overhaul** — The removal of "research shows," "working things out," and "moving forward together" is consistent across all touchpoints: `messages.ts`, `stage2.ts`, `dispatch-handler.ts`, and `stage-prompts.ts`. The replacement framing ("not about excusing, agreeing, or deciding what happens next") is cleaner and avoids implying an outcome.

**Stage 2 post-share short-circuit** (`sharing.ts`) — The early return for `currentStage === 2` in `generatePostShareContinuation` prevents restarting partner-perspective work after sharing. This is a meaningful behavior change but the right one — once they've shared, asking them to keep exploring the partner's perspective would be confusing.

**Python invariant additions** — `check_transcript_shared_context_blocks_labeled` catches unlabeled separator blocks. The `LABELED_SHARED_BLOCK_RE` covers the known label patterns (emoji prefixes, "SHARED WITH YOU", etc.). The `build_loop_service_env` correctly defaults `MOCK_LLM=false` via `setdefault`, preserving explicit overrides. Test coverage for both is good.

### Notes (non-blocking)

1. **Duplicate repair-scrubbing regex** — `scrubStage2RepairFraming` in `messages.ts` and the inline regex in `stage2.ts` (consent-to-share flow) do overlapping work. The `stage2.ts` version is a subset. Not a bug, but could be DRY'd in a follow-up to avoid drift.

2. **`researchToRepairRegex` span matching** — The regex joins three patterns with `[^.]*`, so it requires all three segments ("research"/"genuinely trying"/"work things out") to appear within one sentence (no periods between them). This is fine as a safety net for LLM-generated text, but if the LLM splits across sentences the regex won't match. Given the prompt changes, this is unlikely to fire at all — it's belt-and-suspenders.

3. **`buildStage2TransitionFallback` partner name fallback** — Minor: the function uses `partnerName || 'your partner'` for the body text but `partnerName || 'Your partner'` (capitalized) for the `mutualPhrase`. Consistent, but worth noting the two different casings are intentional (one mid-sentence, one sentence-start).

LGTM — approving. The prompt and copy changes are well-aligned with the calibration findings, and the test coverage is solid for the scope.

slam-paws:
Looking into this — reviewing the PR now.

slam-paws:
## Review

Solid calibration pass. The two main threads — control tag stripping and repair-framing removal — are well-motivated and properly tested. Here's my read:

### Strengths

1. **Control tag stripping** (`micro-tag-parser.ts`) — Good defensive fix. `extractBooleanControlTag` + `stripKnownControlTags` handles the case where the LLM emits XML tags outside `<thinking>` blocks. Users should never see `<feel_heard_check>Y</feel_heard_check>` in chat.

2. **Repair-framing removal** — The shift from "research shows this is the strongest predictor of working things out" to "not about excusing, agreeing, or deciding what happens next" is philosophically correct. Stage 2 empathy is a *protected attempt*, not a sales pitch for reconciliation.

3. **Stage 2 pacing** — Requiring 4+ substantive turns and two layers of partner experience before `ReadyShare:Y` prevents premature empathy drafts. The "Does that feel like your real attempt?" checkpoint is a nice safeguard.

4. **Post-share continuation guard** (`sharing.ts`) — The early return for Stage 2 prevents re-inviting perspective exploration after the user already shared. Smart fix for the loop-restart bug.

5. **Test coverage** — 242 Jest + 65 Python tests, plus real-LLM gold-loop validation on both scenarios. The regression tests for the new invariants (visible control tags, labeled shared-context blocks, chat-copy/input with follow-up turns) lock in the fixes.

- The inline regex scrub in `stage2.ts:871-877` duplicates the pattern from `scrubStage2RepairFraming` in `messages.ts`. If this grows, might be worth sharing the helper — but fine for now since the code paths are distinct.
- The hardcoded Stage 2 early-return in `generatePostShareContinuation` means post-share Stage 2 always returns the same string regardless of context. This is clearly intentional to prevent re-exploration, but worth noting if we later want post-share Stage 2 to vary (e.g., when partner has already validated).
- `MOCK_LLM` defaulting to `false` is the right call for gold-loop calibration — the whole point is real LLM behavior. Good that explicit overrides are preserved.

### Verdict

LGTM. The follow-up items you noted (post-empathy validation/action boundary clarity, Stage 1 narrative-as-truth risk) are worth tracking but correctly not blocking this pass.

PR #452, opened 2026-05-07
Stage 4 typed proposal capture (gold-loop self-improvement)

shantamg:
## Summary

Replaces the regex-based Stage 4 proposal capture pipeline with a typed structured-output path. Brings Stage 4 into structural parity with Stages 1 and 3 (which already use typed signals like `FeelHeardCheck` / `<needs>`).

## Major changes

- **Typed proposal capture is authoritative.** Stage 4 prompts emit a hidden `<stage4_proposals>` JSON block with `action`, `classification`, `kind`, and `ownerUserId`. When present, the legacy free-text `ProposedStrategy` compatibility path is bypassed.
- **Legacy regex layer removed.** `stage4-capture.service.ts` no longer contains `proposalFamily`, `sameProposalFamily`, `proposalSpecificityScore`, or anchored phrase blacklists. `isConcreteProposal` collapsed to a generic specificity gate.
- **Stage 4 prompt contract tightened.** Reflective text, success markers ("what I'd walk away knowing"), and guarded process-consent are explicitly classified `IGNORE + PROCESS`, not proposals. Generic "User will…" labels are forbidden.
- **Harness boundary fixes.** Bounded actor proposal turns; non-terminal Stage 4 actor waits; clarified prompt contract / harness gates.
- **Eval harness invariants** patched to support `--seed-target-stage` runs (don't require missing earlier-stage transcripts; don't treat milestone separators as unlabeled shared context; don't require Stage 1 felt-heard evidence on focused Stage 4 runs).
- **Mobile test drift fixed**: stale copy expectations (e.g., `Elevated` → `Moderate`), Axios/Sentry mocks, Stage 4 close-button contract.

## DoD evidence

Both gold scenarios pass the supplement's hard transcript invariants on the final committed tree:

- `eval/runs/20260508-160409-adam-eve-iter-01` — shared-agreement closure
- `eval/runs/20260508-161422-james-catherine-iter-01` — no-shared-agreement closure (correct shape)

`npm run check` passes. `npm run test` passes (mobile, backend, shared).

## What this does NOT solve

The LLM occasionally still typed-classifies reflective fragments (e.g., "even see the shame under some of it") as individual commitments. Hard invariants are not affected, but `mwf_handling` LLM-judge scores reflect this. See `.planning/stage4-loop-diagnostic.md` (gitignored, local) for full diagnosis and recommended next steps.

## Test plan

- [ ] Real session: walk through Stage 4 with a partner and verify proposal cards reflect what was actually said, no duplicate cards, ownership is clear.
- [ ] Verify `CONFIRMED YOUR UNDERSTANDING` divider stays pinned (this branch picked up #459's fix via rebase).
- [ ] Verify no-shared-agreement closure path still produces dignified copy.
- [ ] Spot-check that reflective text / success markers don't get typed as proposals.


slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Major architectural shift: replaces the regex-based Stage 4 proposal capture pipeline with a typed structured-output path using hidden `<stage4_proposals>` JSON blocks. Also introduces LLM-driven Stage 1 feel-heard confirmation (`FeelHeardConfirmed:Y`), a new Stage 4 auto-closure service, and significant eval harness/prompt improvements. Mobile fixes address drawer layout, animation race conditions, and chat UI state logic.

## Feedback

### Blocking

- **[Bug] Revision audit `before` snapshot is wrong** — `stage4-capture.service.ts`: In the superseded-proposal branch of `applyOperation`, `Object.assign(superseded, ...)` mutates the proposal **before** `proposalSnapshot(superseded)` is called for the revision record's `before` field. This means `before` and `after` will be identical in the audit log. Fix: capture the snapshot before the mutation.
  ```
  // Current (buggy):
  Object.assign(superseded, { description: operation.description, ... });
  await prisma.stage4ProposalRevision.create({
    data: {
      before: proposalSnapshot(superseded),  // ← already mutated!
      ...
  ```

- **[Consistency] Open needs filtering mismatch** — `stage4-auto-closure.service.ts` line ~99 includes both `OPEN` and `PARTIAL` coverage rows as open needs, but the manual `closeStage4` endpoint in `stage4.ts` was changed to filter only `OPEN` (removing the previous `PARTIAL` inclusion). This means the same session will report different `openNeedIds` depending on whether closure comes from conversation signal vs. button tap. Pick one policy and apply it to both paths.

### Non-blocking

- **[Maintainability] Scenario-specific regexes in `inferProposalKind`** — The gold-loop supplement doc explicitly calls out the prior approach of keying on scenario nouns as "a dead end." The new `inferProposalKind` still contains many patterns matching "adam", "eve", "portugal", "saturday", "brother", "ceramics class", etc. These will need to be reworked for any scenario beyond the two gold sets. Consider moving toward the typed `kind` from `<stage4_proposals>` as the primary signal and only falling back to regex for the compatibility path.

- **[Observation] Eval run data in the diff** — `eval/runs/` is gitignored in this PR, but ~20K lines of run data are included. This is presumably intentional to show DoD evidence, but it inflates the diff significantly. If these files are already gitignored, they may not land on main — worth confirming.

- **[Streaming] `hasPotentialTagStart` regex broadened** — The regex `/<\/?(d|n|s)[a-z0-9_]*$/i` now matches any tag starting with `s`, which could buffer content containing `<strong>`, `<span>`, `<section>`, etc. In practice the AI shouldn't emit HTML, but this is a latent false-positive risk. A tighter match like `/<\/?(d|n|stage4)[a-z0-9_]*$/i` would be more precise.

- **[Nice work] Stage 4 state machine** — The shift from raw selection counts to `selectionSubmitted` gates for phase derivation is a solid improvement. The `hasSelectionForEveryActiveSharedProposal` check properly handles late-arriving proposals. Good test coverage for the edge cases.

- **[Nice work] Micro-tag parser** — The `parseStage4ProposalsBlock` function has proper type guards (`isStage4ProposalClassification`, `isStage4ProposalAction`), graceful JSON parse fallback, and clean field validation. Well structured.

- **[Nice work] Auto-closure service** — Clean separation from the button endpoint, proper transaction with session status update and progress completion, and appropriate guard rails (only no-shared-agreement, requires explicit signal).

## Verdict

Comment — the core architecture is sound and well-tested. Two issues worth addressing before merge: the revision snapshot ordering bug (data correctness) and the open-needs filtering inconsistency (behavioral divergence between closure paths).

shantamg:
@slam-paws — addressed all three findings in this push:

- **[Bug] Revision audit `before` snapshot is wrong** — fixed in `e745fa12`. Capture `proposalSnapshot(superseded)` into a local `beforeSnapshot` before the `Object.assign` mutation; pass that local into the revision row's `before` field. Added a regression test in `stage4-capture.service.test.ts` that asserts `before !== after` and that `before` matches the pre-mutation state.

- **[Consistency] Open needs filtering mismatch** — fixed in `eeac281d`. Unified both paths on `OPEN || PARTIAL`. Git blame showed the manual endpoint had drifted to OPEN-only inside checkpoint commit `8be3f993` with no stated rationale, while the auto-closure service was always `OPEN+PARTIAL`. Restored the manual endpoint to match. Added a parity assertion in the auto-closure service test, and updated the existing routes test expectation.

- **[Maintainability] Scenario-specific regexes in `inferProposalKind`** — fixed in `82612617`. Stripped all scenario-named patterns. The function is now a compatibility fallback for the legacy free-text `ProposedStrategy` path (typed `<stage4_proposals>` already bypasses it). It uses only generic signals: `we/together/let's` → SHARED, `I can/I'll/I will` → INDIVIDUAL, and defaults to INDIVIDUAL when uncertain so we never auto-promote a fragment to SHARED.

`npm run check` is clean. Backend stage4 + shared suites green; pre-existing circuit-breaker failures are unrelated. Re-review please.

slam-paws:
Looking into the fixes now — re-reviewing.

slam-paws:
Re-reviewed all three fix commits. Everything checks out:

**1. Revision audit `before` snapshot (`e745fa12`)** — Clean fix. `beforeSnapshot` is now captured from `proposalSnapshot(superseded)` *before* the `Object.assign` mutation, so `before` and `after` in the revision row are genuinely different snapshots. The old code was snapshotting *after* the mutation for both fields. Regression test correctly asserts `before !== after` and that `before` matches the original state.

**2. Open needs filtering (`eeac281d`)** — The manual `closeStage4` endpoint now filters on `OPEN || PARTIAL`, matching the auto-closure service. Parity assertion in the auto-closure test is a good safeguard against future drift. Route test expectation updated to include the PARTIAL row.

**3. `inferProposalKind` cleanup (`82612617`)** — ~90 lines of scenario-specific regex removed, replaced with three generic signal checks (explicit individual self-references, shared/joint phrasing, first-person commitment verbs) plus an `INDIVIDUAL_COMMITMENT` default. The default-to-individual rationale is sound: auto-promoting an ambiguous fragment to SHARED would pull a partner into a commitment they didn't author. Test expectation updated to reflect the new default.

No new concerns. LGTM.

PR #457, opened 2026-05-08
[codex] Fix needs drawer footer height

shantamg:
## Summary
- Measure the rendered `NeedsDrawer` container height and use it for snap positions, close position, drawer height, and content height.
- Keep `useWindowDimensions()` as the fallback before layout measurement completes.

## Root Cause
On web, the native-app banner renders above the app root inside a clipped phone-width frame. The drawer was sizing from full browser `windowHeight`, so its fixed footer could be laid out below the actual clipped app area when the banner was visible.

## Validation
- `npm test -- NeedsDrawer.test.tsx --runInBand`
- `npm run check`

Note: installing dependencies in this fresh worktree produced npm engine/audit warnings because this shell is on Node 25.2.1 while the repo expects Node 20.x; the validation commands still passed after install.

### DIFF
--- a/mobile/src/components/NeedsDrawer.tsx
+++ b/mobile/src/components/NeedsDrawer.tsx
@@ -19,6 +19,7 @@ import {
   StyleSheet,
   BackHandler,
   useWindowDimensions,
+  LayoutChangeEvent,
 } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { appWidthStyle, colors } from '@/theme';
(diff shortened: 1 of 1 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Replaces `useWindowDimensions` with an `onLayout` measurement on the drawer's root container, so the drawer sizes itself to the actual visible area rather than the full browser window height. Falls back to `windowHeight` until the first layout fires.

## Feedback
- **Correctness**: The approach is sound. `handleContainerLayout` correctly guards against zero-height layouts and uses a >1px threshold to prevent re-render thrashing from sub-pixel fluctuations. `Math.ceil` is a good touch for consistency.
- **Refs pattern**: Using `drawerHeightRef` to feed the `PanResponder` (created once via `useRef`) is the correct pattern — avoids stale closures.
- **Transition on measurement**: When `containerHeight` first resolves from `null` to the measured value, the `useEffect` calls `drawerTranslate.setValue(position)` directly (no animation). This could cause a brief visual jump if layout measurement is slow, but in practice it fires before the opening animation completes, so it should be imperceptible.
- **Tests (nice-to-have)**: The existing tests pass, but there's no coverage for the `handleContainerLayout` path. A test that fires `onLayout` on the root container and asserts the drawer height updates would strengthen confidence. Not blocking since this is a layout-level concern that's hard to unit test meaningfully.
- **Note**: This targets `codex/fix-ai-rendering-stage1-verbal-gate`, not `main` — stacked PR, which is fine.

## Verdict
Approve — clean, well-scoped fix with a sensible fallback strategy.

shantamg:
Closing because the equivalent drawer host-height fix is already present on the base branch.

PR #461, opened 2026-05-09
[codex] Add notification prompts and biometric app lock

shantamg:
## Summary
- wire Expo push notification registration and notification tap routing in the authenticated mobile app
- ask for notification permission with a full-screen drawer before starting a session when permission is still undecided
- personalize push notification copy with the recipient's nickname for the other person, avoid “partner” wording in notification text, and make each notification clearer about what changed
- add a global biometric privacy shield for signed-in users who enabled biometrics, plus Face ID usage config
- restore the online presence dot beside the session header name while preserving the current stage subtitle
- add a tap target on the session header name that opens a person details drawer with online/last-seen status, plain-English stage context, and the conversation topic

## Notes
- Most notification and biometric overlay behavior is OTA-able JS. The new Face ID usage string / expo-local-authentication config in app.json requires a native build to be guaranteed in iOS metadata.
- Existing unrelated dirty files were not included: .claude/scheduled_tasks.lock and slam-bot-midstream-parity-goal.md.

## Validation
- mobile: npm run check
- mobile: targeted eslint on changed session drawer/header files
- shared: npm run check
- backend: npm run check
- backend: npx eslint src/services/push.ts src/controllers/session-state.ts --quiet
- backend: npm test -- --runTestsByPath src/services/__tests__/push.test.ts --runInBand

### DIFF
--- a/backend/src/controllers/session-state.ts
+++ b/backend/src/controllers/session-state.ts
@@ -33,10 +33,7 @@ interface CompactGates {
   signedAt?: string;
 }
 
-async function acceptPendingInvitationForE2ESession(
-  sessionId: string,
-  userId: string
-): Promise<boolean> {
+async function acceptPendingInvitationForE2ESession(sessionId: string, userId: string): Promise<boolean> {
   if (process.env.E2E_AUTH_BYPASS !== 'true' || process.env.NODE_ENV === 'production') {
     return false;
   }
@@ -63,7 +60,7 @@ async function acceptPendingInvitationForE2ESession(
     return false;
   }
 
-  await prisma.$transaction(async (tx) => {
+  await prisma.$transaction(async tx => {
     await tx.relationshipMember.upsert({
       where: {
         relationshipId_userId: {
@@ -215,27 +212,19 @@ export async function getSessionState(req: Request, res: Response): Promise<void
     const partnerId = await getPartnerUserId(sessionId, user.id);
 
     // Get my membership (for nickname I use for partner)
-    const myMember = session.relationship.members.find(
-      (m) => m.userId === user.id
-    );
-    const partnerMember = session.relationship.members.find(
-      (m) => m.userId !== user.id
-    );
+    const myMember = session.relationship.members.find(m => m.userId === user.id);
+    const partnerMember = session.relationship.members.find(m => m.userId !== user.id);
 
     // Build progress response
     const myProgressRecord = session.stageProgress
-      .filter((sp) => sp.userId === user.id)
+      .filter(sp => sp.userId === user.id)
       .sort((a, b) => b.stage - a.stage)[0];
     const partnerProgressRecord = partnerId
-      ? session.stageProgress
-          .filter((sp) => sp.userId === partnerId)
-          .sort((a, b) => b.stage - a.stage)[0]
+      ? session.stageProgress.filter(sp => sp.userId === partnerId).sort((a, b) => b.stage - a.stage)[0]
       : null;
 
     // Get Stage 1 record for milestones
-    const myStage1Record = session.stageProgress.find(
-      (sp) => sp.userId === user.id && sp.stage === 1
-    );
+    const myStage1Record = session.stageProgress.find(sp => sp.userId === user.id && sp.stage === 1);
 
     const defaultProgress = {
       stage: 0,
@@ -273,17 +262,26 @@ export async function getSessionState(req: Request, res: Response): Promise<void
     };
 
     // Partner display name
-    const partnerDisplayName =
-      myMember?.nickname ||
-      partnerMember?.user.firstName ||
-      partnerMember?.user.name ||
-      null;
+    const partnerDisplayName = myMember?.nickname || partnerMember?.user.firstName || partnerMember?.user.name || null;
+    const partnerVessel = partnerId
+      ? await prisma.userVessel.findUnique({
+          where: {
+            userId_sessionId: {
+              userId: partnerId,
+              sessionId,
+            },
+          },
+          select: {
+            lastViewedAt: true,
+          },
+        })
+      : null;
 
     // Build messages response (reverse to chronological order)
     const hasMoreMessages = messages.length > 25;
     const messageSlice = hasMoreMessages ? messages.slice(0, 25) : messages;
     const messagesResponse = {
-      messages: messageSlice.reverse().map((m) => ({
+      messages: messageSlice.reverse().map(m => ({
         id: m.id,
         sessionId: m.sessionId,
         senderId: m.senderId,
@@ -320,13 +318,9 @@ export async function getSessionState(req: Request, res: Response): Promise<void
       : null;
 
     // Build compact status from Stage 0 progress gatesSatisfied
-    const myStage0Record = session.stageProgress.find(
-      (sp) => sp.userId === user.id && sp.stage === 0
-    );
+    const myStage0Record = session.stageProgress.find(sp => sp.userId === user.id && sp.stage === 0);
     const partnerStage0Record = partnerId
-      ? session.stageProgress.find(
-          (sp) => sp.userId === partnerId && sp.stage === 0
-        )
+      ? session.stageProgress.find(sp => sp.userId === partnerId && sp.stage === 0)
       : null;
 
     const myGates = myStage0Record?.gatesSatisfied as CompactGates | null;
@@ -373,6 +367,7 @@ export async function getSessionState(req: Request, res: Response): Promise<void
         createdAt: session.createdAt.toISOString(),
         resolvedAt: session.resolvedAt?.toISOString() ?? null,
         lastViewedAt: userVessel?.lastViewedAt?.toISOString() ?? null,
+        partnerLastViewedAt: partnerVessel?.lastViewedAt?.toISOString() ?? null,
         lastSeenChatItemId: userVessel?.lastSeenChatItemId ?? null,
       },
 
--- a/backend/src/services/__tests__/push.test.ts
+++ b/backend/src/services/__tests__/push.test.ts
@@ -1,16 +1,9 @@
-import {
-  sendPushNotification,
-  sendPushNotifications,
-  isValidPushToken,
-  PUSH_MESSAGES,
-  resetExpoClient,
-} from '../push';
+import { sendPushNotification, sendPushNotifications, isValidPushToken, PUSH_MESSAGES, resetExpoClient } from '../push';
 import { prisma } from '../../lib/prisma';
 import type { SessionEvent } from '../realtime';
 
 // Mock Prisma
 jest.mock('../../lib/prisma');
-
 
 // Mock Expo SDK
 const mockSendPushNotificationsAsync = jest.fn();
@@ -35,7 +28,7 @@ describe('Push Notification Service', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     resetExpoClient();
-    ((Expo as unknown as { isExpoPushToken: jest.Mock }).isExpoPushToken).mockReturnValue(true);
+    (Expo as unknown as { isExpoPushToken: jest.Mock }).isExpoPushToken.mockReturnValue(true);
   });
 
   describe('PUSH_MESSAGES', () => {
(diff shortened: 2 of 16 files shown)


slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary
Adds Expo push notification registration with deep-link routing on tap, a full-screen notification permission drawer shown before first session creation, and a biometric privacy shield (Face ID / Touch ID) that locks the app after 5 seconds in the background.

## Feedback

**Good patterns observed:**
- Clean separation: `notifications.ts` service handles all Expo APIs, `useNotifications` hook wires up routing, `NotificationPermissionDrawer` is pure UI
- Web shim for `BiometricLockContext.web.tsx` prevents crashes in non-native environments
- `BACKGROUND_TIMEOUT_MS` grace period (5s) prevents annoying re-auth on quick app switches
- Notification copy improvements are consistently warmer and more action-oriented
- Adding `screen: 'session'` to push payload data enables proper deep-link routing

**Minor observations (non-blocking):**

- **[Correctness]** `BiometricLockOverlay.tsx:383` — `handleBiometricAuth` is referenced in the `useEffect` dep array comment but not listed in the actual dependency array. ESLint's exhaustive-deps rule would flag this. Since `handleBiometricAuth` captures `unlock` and `biometricName` from the parent scope, stale closures are possible if those change while the overlay is mounted. Low risk in practice since biometric config rarely changes mid-session.

- **[Correctness]** `new.tsx:271` — `setPendingSubmit(() => createSessionFromForm)` stores a function reference via the functional-update form of `useState`. This works correctly because React interprets `() => fn` as "set the value to `fn`", but it's a subtle pattern. A comment would help future readers understand it's intentional.

- **[Architecture]** `notifications.ts:24-32` — `setNotificationHandler` is called at module-import time (top-level side effect). This is fine for Expo's notification system but worth noting — any file importing from `notifications.ts` triggers this handler registration, even if it only needs `getNotificationPermissionStatus`.

- **[Style]** Both `BiometricLockOverlay` and `NotificationPermissionDrawer` use `StyleSheet.create` — consistent with the rest of the codebase.

## Verdict
**Approve.** Well-structured feature addition. The notification flow (check permission → show drawer → request → register token → continue session creation) is clean. The biometric lock context with AppState listeners is solid. The push copy improvements are a nice touch. No blocking issues found.

slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary

Solid PR that adds three distinct features: (1) personalized push notification copy using the recipient's nickname for the other person, replacing generic "partner" wording with template placeholders; (2) a biometric app lock that activates after 5s in background; (3) notification permission prompting via a full-screen drawer before first session creation. Also restores the online presence dot in the session header.

## Feedback

- **[Minor — Stale closure]** `BiometricLockOverlay.tsx`: The second `useEffect` (auto-trigger) references `handleBiometricAuth` but doesn't include it in the dependency array. This creates a stale closure risk — if `unlock` or `biometricName` change between renders, the effect would use old values. In practice this is unlikely to cause bugs today, but adding it to deps (and wrapping `handleBiometricAuth` in `useCallback`) would be correct.

- **[Nit — Copy consistency]** `NotificationPermissionDrawer.tsx`: The preview notification text reads *"Your partner is ready"* — this still uses "partner" wording, which the rest of the PR specifically moves away from in favor of personalized names. Consider changing to something like *"Sam is ready"* or a generic *"They are ready"* for consistency with the new tone.

- **[Observation — Unused hook exports]** `useNotifications.ts` exports `shouldAskForSessionNotifications` and `requestSessionNotifications`, but `new.tsx` imports directly from `services/notifications.ts` instead. The hook is only used for its side effects (registration + tap routing) in `_layout.tsx`. Not blocking, but worth noting for future cleanup.

- **[Good pattern]** The `getActorUserId` fallback chain and `possessive()` helper are clean and handle edge cases well (e.g., `'They' → 'Their'`, names ending in 's'). The try/catch in `getNotificationActorName` with graceful fallback to `'They'` is the right call for a notification path — never fail to deliver because of a name lookup error.

- **[Good pattern]** The 5-second `BACKGROUND_TIMEOUT_MS` grace period for the biometric lock is thoughtful UX — avoids re-prompting when quickly switching apps. The web stub (`BiometricLockContext.web.tsx`) is a nice touch for platform safety.

## Verdict

**Approved.** The template rendering approach for notifications is clean and testable, the biometric lock is well-scoped for v1, and the notification permission drawer provides good UX. The minor items above are non-blocking.

slam-paws:
On it — reviewing now.

slam-paws (dismissed):
## Summary

Solid PR that adds four well-scoped features: personalized push notification copy with nickname resolution, a notification permission prompt before session creation, a biometric privacy shield, and a partner info drawer accessible from the session header. The push template refactor from static strings to `{actor}`/`{actorPossessive}`/`{stageName}` placeholders is clean and well-tested.

## Feedback

- **[Minor – UI] Duplicate StatusDot**: In `SessionChatHeader.tsx`, when `stageName` is falsy and `hideOnlineStatus` is false, the `StatusDot` now renders in *both* the new `nameRow` (next to the name) and in the existing `statusRow` (with online text) below it. This produces two green/grey dots stacked vertically. Likely intentional (dot-beside-name for quick glance when stageName is showing, full status row as fallback), but worth a quick visual check on device to confirm the no-stageName case doesn't look redundant.

- **[Minor – Types] Type cast for `partnerLastViewedAt`**: The `UnifiedSessionScreen` uses `(session as { partnerLastViewedAt?: string | null } | undefined)?.partnerLastViewedAt` instead of relying on the updated `SessionStateResponse` DTO. This suggests the session object in the component isn't fully typed from the DTO. Not blocking, but a follow-up to thread the DTO type properly would eliminate the cast.

- **[Observation – Privacy] `partnerLastViewedAt` exposure**: The new field exposes when the other person last viewed the session. This is reasonable for presence UX, but worth confirming it aligns with the vessel privacy model (showing *activity timestamps* vs. *conversation content* is a meaningful distinction, so likely fine).

- **[Good] Push notification personalization**: The `getNotificationActorName` function correctly resolves `recipientMember?.nickname` (the name the recipient gave their partner) with clean fallback chain to firstName → name → "They". The `possessive()` helper handles "They" → "Their" correctly. Try/catch fallback to "They" on any DB error is sensible.

- **[Good] Biometric lock architecture**: Clean separation — context provider handles AppState transitions with a 5-second grace period, web stub prevents crashes in browser environments, and the overlay component manages its own animation state. The `shouldAutoTrigger` latch prevents re-prompting loops.

- **[Good] Notification permission flow**: Prompting only when `undetermined` (not on every session) and proceeding regardless of the user's choice is the right UX pattern.

## Verdict

Approve — well-structured, tested, and thoughtfully designed. The two minor items (duplicate StatusDot visual check, type cast cleanup) are non-blocking.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Large, well-structured PR that adds four user-facing features: personalized push notification copy (with actor name/stage resolution), a notification permission drawer before session creation, a biometric privacy shield (Face ID/Touch ID), and a partner info drawer accessible from the session header. Backend changes expose `partnerLastViewedAt` for online presence.

## Feedback

### Positive patterns

- **Push message templating** (`push.ts`): The `{actor}` / `{actorPossessive}` / `{stageName}` template system is clean and well-tested. The `possessive()` helper correctly handles the "They" → "Their" edge case and the trailing-s rule.
- **Web fallback files**: `BiometricLockContext.web.tsx` provides a proper no-op provider for web — good platform separation.
- **Graceful degradation**: `getNotificationActorName` falls back to `"They"` on error, and the notification drawer handles loading/skip states well.
- **5-second background timeout** in the biometric lock context is a nice UX touch — avoids re-auth on quick app switches.

- **[Types]** `UnifiedSessionScreen.tsx` — The PR accesses `partnerLastViewedAt` via a type cast: `(session as { partnerLastViewedAt?: string | null } | undefined)?.partnerLastViewedAt`. The DTO in `shared/src/dto/session-state.ts` is already updated with this field, so the cast suggests the session object in that component doesn't use `SessionStateResponse` directly. This works but is fragile — if the field is renamed in the DTO, the cast won't catch it. Worth a follow-up to align the screen's session type with the shared DTO.

- **[Performance]** `push.ts:getNotificationActorName` — Each push notification now makes an additional Prisma query to resolve the actor name. For `sendPushNotifications` (batch), this means N sequential DB queries (one per recipient). The current usage (2-person sessions) makes this negligible, but worth noting if the batch path ever scales.

- **[Style]** The new components use `StyleSheet.create()` which is consistent with existing components in the codebase — no issue here.

- **[Test coverage]** The nickname resolution test (`uses the recipient nickname for the other person when available`) is a great addition. The `session.findUnique` mock correctly sets up the relationship member structure.

- **[Duplicate stage names]** `UnifiedSessionScreen.tsx` defines a local `STAGE_FRIENDLY_NAMES` record that duplicates the shared `STAGE_FRIENDLY_NAMES` from `@meet-without-fear/shared` (which `push.ts` now imports). Not introduced by this PR, but worth consolidating in a follow-up.

- **[BiometricLockOverlay]** `handleBiometricAuth` is referenced in a `useEffect` dependency array indirectly — the effect depends on `lockState` and `shouldAutoTrigger`, and calls `handleBiometricAuth` inside. Since `handleBiometricAuth` isn't wrapped in `useCallback` and isn't in the dep array, this is actually fine (avoids infinite loops), but the eslint exhaustive-deps rule may flag it.

## Verdict

**Approve** — Clean implementation across backend and mobile. The notification personalization is a meaningful UX improvement, the biometric lock is well-architected with proper platform separation, and test coverage is solid. The type cast and duplicate stage names are minor follow-up items, not blockers.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Solid PR that adds three distinct features: personalized push notifications with actor name resolution, a biometric privacy shield, and a partner info drawer with online presence. The notification template system is well-designed and the biometric lock context handles edge cases (background timeout, auto-trigger) thoughtfully.

## Feedback

- **[Privacy — minor]** `partnerLastViewedAt` is now exposed in `SessionStateResponse` (`shared/src/dto/session-state.ts:45`). This reveals when the other person last opened the session. Given MWF's privacy-first posture (vessel model), worth a quick sanity check that this aligns with the intended privacy boundaries — the partner info drawer uses it for "Last seen X ago" which feels reasonable, but it's a new cross-user data exposure.

- **[Type safety — minor]** In `UnifiedSessionScreen.tsx`, `partnerLastViewedAt` is accessed via a type assertion: `(session as { partnerLastViewedAt?: string | null } | undefined)?.partnerLastViewedAt`. This suggests the local session type from the query hook hasn't been updated to include the new field. Consider updating the query hook's return type so this cast isn't needed.

- **[Stale closure — minor]** In `BiometricLockOverlay.tsx`, the `useEffect` at line ~73 references `handleBiometricAuth` but it's not wrapped in `useCallback` and isn't in the dependency array. The ESLint rule is likely suppressed. In practice this is fine since the closure captures stable refs (`unlock`, `biometricName`), but wrapping `handleBiometricAuth` in `useCallback` would be cleaner.

- **[Copy consistency — nit]** The `NotificationPermissionDrawer` preview text still says "Your partner is ready" (line ~1199), while the PR description states the goal is to avoid "partner" wording in notifications. The actual push templates correctly use `{actor}`, so this is just the static preview mock in the drawer.

- **[Style — noted]** The three new components (`BiometricLockOverlay`, `NotificationPermissionDrawer`, `PartnerInfoDrawer`) use `StyleSheet` rather than NativeWind. These are all modal overlays so this may be intentional for z-index reliability — just flagging for awareness.

- **[Architecture — good]** The `getNotificationActorName` → `renderPushMessage` pipeline in `push.ts` is clean. The `possessive()` helper handles the "They" → "Their" case and the trailing-s rule correctly. The fallback to "They" on any error is a safe default.

- **[Performance — fine]** `getNotificationActorName` adds one extra DB query per push notification send. Since push notifications are low-frequency events (not per-message), this is perfectly acceptable.

## Verdict
Approve — well-structured feature PR. The minor items above are non-blocking suggestions for follow-up.

Issue #463, opened 2026-05-09
Bug: Failed message text disappears from input on send error


slam-paws:
## Bug Report

When sending a message while out of service/offline, the app shows a good error message but the user's typed text disappears entirely. The user loses what they wrote with no way to recover or retry.

**Expected behavior (per reporter):** Either:
1. Put the text back into the input field (simplest), OR
2. Show a "not sent" indicator on the message with a retry button

**Current behavior:** Text is gone after the error — user has to retype from scratch.

## What We Know So Far

• The send mutation likely clears the input on invocation rather than on success
• Optimistic update may be removing the draft text before confirming delivery
• No immediate leads on exact root cause — needs full investigation

## Provenance

- **Channel:** #bugs-and-requests
- **Reporter:** Shantam (U0AD3TF2U7L)
- **Original message:** "I was out of service for a moment and when I tried to send a message, I got a good error message but the message disappeared. This is bad. It should either put the text back into the input for me (easiest) or have an indicator that it was not sent and a button to retry. I'm open to either."
- **Timestamp:** 1778293516.554359

slam-paws:
## Investigation Report

**Root cause**: `ChatInput.handleSend` (`mobile/src/components/ChatInput.tsx:58`) clears the input state eagerly and synchronously (`setInput('')`) before the async SSE send completes. When the connection fails (offline/network error), the error handler in `useStreamingMessage` cleans up the optimistic message from cache and shows an error toast via `showError()`, but has no mechanism to restore the original text back to `ChatInput`'s local `useState`.

The text is lost because:
1. `ChatInput` owns its own input state via `useState` (line 41)
2. `handleSend` clears it immediately (line 58), then calls `onSend(message)` (line 61)
3. The send is fire-and-forget — `handleSendMessage` in `useUnifiedSession.ts:834` calls `streamingSendMessage()` without awaiting or returning a promise
4. When SSE errors occur asynchronously (lines 813-838), `cleanupFailedStream` removes optimistic messages from cache, but the cleared input text is already gone with no path to restore it

**Evidence**: Code trace through the send path:
- `ChatInput.tsx:52-68` — eager clear before async send
- `useUnifiedSession.ts:819-836` — fire-and-forget bridge to streaming hook
- `useStreamingMessage.ts:813-855` — error handlers clean up cache but can't restore input
- `useStreamingMessage.ts:493` — `lastParamsRef.current = params` already stores the sent content (used by `retry()`)

**Affected files**:
[...]

PR #465, opened 2026-05-09
Fix failed message text disappearing from input on send error

slam-paws:
## Changes
- Fix: Restore user's typed text to the input field when a message send fails (e.g., offline/network error)
- Add: `failedMessageContent` field to `useStreamingMessage` return value, derived from existing `lastParamsRef` when status is `'error'`
- Thread: `failedMessage` prop through `useUnifiedSession` → `UnifiedSessionScreen` → `ChatInterface` → `ChatInput`
- Add: `useEffect` in `ChatInput` that restores text when `failedMessage` transitions to a non-null value (only if input is currently empty)

Related to #463

## Provenance
- **Original message:** "I was out of service for a moment and when I tried to send a message, I got a good error message but the message disappeared. This is bad. It should either put the text back into the input for me (easiest) or have an indicator that it was not sent and a button to retry."

## Docs updated
No doc changes needed — bug fix in existing components with no doc mappings.

### DIFF
--- a/mobile/src/components/ChatInput.tsx
+++ b/mobile/src/components/ChatInput.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback, useRef } from 'react';
+import { useState, useCallback, useRef, useEffect } from 'react';
 import { View, TextInput, TouchableOpacity, Text, Platform, type TextInput as TextInputType } from 'react-native';
 import { Send, Mic } from 'lucide-react-native';
 import { createStyles } from '../theme/styled';
@@ -16,6 +16,8 @@ interface ChatInputProps {
   showCharacterCount?: boolean;
   /** Optional voice press handler -- when provided, renders a mic button */
   onVoicePress?: () => void;
+  /** Content of a failed message to restore to the input field */
+  failedMessage?: string | null;
 }
 
 // ============================================================================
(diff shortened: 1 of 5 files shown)


slam-paws:
## Bot review — PR #465

**LGTM** ✅

### What it does
Restores a user's typed text to the input field after a failed send (e.g., offline/network error). Implements option 1 from issue #463. Clean, minimal change.

### Correctness
- `failedMessageContent` derivation in `useStreamingMessage` is correct: non-null only when `status === 'error'`, sourced from `lastParamsRef.current?.content` which already holds the last attempted payload.
- The `useEffect` in `ChatInput` correctly guards with `input.length === 0` so it doesn't trample text the user may have already started typing.
- Resetting `justSentRef.current = false` before restoring is correct — clears the autocorrect guard so the restored text behaves normally.
- `eslint-disable react-hooks/exhaustive-deps` on the effect is intentional and sound (excluding `input` prevents the effect from looping; the omission is documented).

### Completeness
- Prop threads cleanly: `useStreamingMessage` → `useUnifiedSession` → `UnifiedSessionScreen` → `ChatInterface` → `ChatInput`. Each layer adds only what it needs.
- Doc-impact check passed — no doc updates needed.

### Minor edge case (non-blocking)
If the user fails twice in a row with the **same** message content, the second failure won't re-trigger the restore (the `failedMessage` dep hasn't changed). This is a corner case and acceptable for now — worth a follow-up if it surfaces in the field.

### Tests (non-blocking observation)
No test was added for the new `failedMessageContent` derived value or the `ChatInput` restore effect. `useStreamingMessage` has no existing test file, so this isn't a regression in coverage. The `ChatInterface.test.tsx` file could benefit from a smoke test for the `failedMessage` prop path if test coverage becomes a priority.

---
Safe to merge. Routes to Stage 05.

slam-paws:

slam-paws:
Superseded by combined PR #468

Issue #462, opened 2026-05-09
Bug: AI messages appear to render/animate twice with visible jump


slam-paws:
## Bug Report

When a message comes in from MWF, it appears to render twice — there's a visible jump partway through the message animation. This suggests the message may be getting inserted into the chat list twice (e.g., once optimistically and once from a server/Ably update) or the animation is restarting mid-stream.

## What We Know So Far

• Likely related to optimistic updates + Ably real-time sync — if both insert the message, the list jumps when de-duplicated
• Could also be a streaming/typing animation that restarts when the message object is replaced in cache
• No immediate leads on exact root cause — needs full investigation

## Provenance

- **Channel:** #bugs-and-requests
- **Reporter:** Shantam (U0AD3TF2U7L)
- **Original message:** "Whenever a message comes in from MWF it seems like maybe it is rendering twice? I see a jump part way through the message animating in."
- **Timestamp:** 1778293516.554359

slam-paws:
## Investigation Report

**Root cause**: Message ID mismatch between streaming and persisted phases causes re-animation.

During SSE streaming, the AI message lives in the React Query cache under a temporary ID (`streaming-${Date.now()}`). When streaming completes, `reconcilePersistedMessages()` calls `invalidateQueries` + `refetchQueries` on `messageKeys.infinite`. The server returns the same message under its real database UUID. React Query replaces the cache page data, so the streaming-ID message disappears and the UUID message appears. The `ChatInterface` animation system tracks animated items by ID — it has never seen the new UUID, so it queues the message for a fresh typewriter pass. Result: the AI message visually vanishes and immediately re-renders with animation.

**Detailed sequence:**
1. User sends message → optimistic user message (`optimistic-user-*`) added to `messageKeys.infinite`
2. SSE `chunk` events → AI placeholder (`streaming-*`) created and progressively updated in cache
3. SSE `text_complete` → cursor stops, `status='complete'`, animation finishes
4. SSE `complete` → `reconcilePersistedMessages(sessionId)` fires
5. Refetch returns server data with real UUIDs → cache page replaced
6. `ChatInterface.shouldAnimateItem()` checks `animatedItemIdsRef` / `seenAnimatedItemIdsRef` / `mountSnapshotIdsRef` — none contain the new UUID → returns `true`
7. AI message gets a second typewriter animation pass → visible jump

[...]

PR #466, opened 2026-05-09
Fix AI messages re-animating after streaming completes

slam-paws:
## Changes
- Fix: Bridge streaming placeholder IDs (`streaming-*`, `optimistic-user-*`) to real server UUIDs before cache reconciliation
- Fix: Pre-register real AI message IDs in animation tracking to prevent ChatInterface from re-animating already-completed messages
- Add: `animationBridge.ts` utility for cross-module animation ID pre-registration

## How it works

When SSE streaming completes, the `complete` event now:
1. Replaces temporary IDs in the React Query cache with real server UUIDs (via new `replaceMessageIdInCache` helper)
2. Pre-registers the real AI message ID so `shouldAnimateItem` returns `false`
3. Then triggers reconciliation — the refetch now matches by ID instead of treating messages as new

This eliminates the visible jump where the AI message would vanish and re-render with a fresh typewriter animation.

Related to #462

## Provenance
- **Original message:** "Whenever a message comes in from MWF it seems like maybe it is rendering twice? I see a jump part way through the message animating in."

## Docs updated
No doc changes needed — bug fix only, no architectural changes. Mapped doc (`docs/architecture/structure.md`) describes file layout, not streaming internals.

### DIFF
--- a/mobile/src/components/ChatInterface.tsx
+++ b/mobile/src/components/ChatInterface.tsx
@@ -20,6 +20,7 @@ import { ChatIndicator, ChatIndicatorType } from './ChatIndicator';
 import { EmpathyValidationCard } from './EmpathyValidationCard';
 import { createStyles } from '../theme/styled';
 import { useSpeech, useAutoSpeech } from '../hooks/useSpeech';
+import { isPreRegisteredAnimatedId } from '../utils/animationBridge';
 
 // ============================================================================
 // Types
@@ -461,6 +462,7 @@ export function ChatInterface({
     if (isIndicator(item) || isValidationCard(item) || isCustomEmptyState(item)) return false;
     if (animatedItemIdsRef.current.has(item.id)) return false;
     if (seenAnimatedItemIdsRef.current.has(item.id)) return false;
+    if (isPreRegisteredAnimatedId(item.id)) return false;
     if (isAtOrBeforeSeenBoundary(item, index)) return false;
 
     if (isCustomCard(item)) {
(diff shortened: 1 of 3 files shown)


slam-paws:
## Bot review ✅ LGTM

**Issue**: [#462 — AI messages appear to render/animate twice](https://github.com/shantamg/meet-without-fear/issues/462)

**Root cause correctly identified**: When SSE `complete` fires, the AI message in cache still has a `streaming-*` placeholder ID. `reconcilePersistedMessages` refetches and the server returns the real UUID — `shouldAnimateItem` sees it as a new message and restarts the typewriter animation.

**Fix is correct**. The sequence in the `complete` handler is right:
1. Bridge `streaming-*` → real UUID in cache (both `.list` and `.infinite` keys, with and without stage)
2. Pre-register the real UUID so `isPreRegisteredAnimatedId` returns `true` before `shouldAnimateItem` can queue another animation
3. Then call `reconcilePersistedMessages` — refetch now matches by ID, no "new message" trigger

Specific checks:

- `replaceMessageIdInCache` covers all four cache key shapes (list/infinite × with/without stage) ✅
- `realUserIdRef` reset to `` at stream start — no stale bridging from previous stream ✅
- User message ID also bridged (no animation, but prevents ID mismatch in cache) ✅
- Conditional guards (`startsWith("streaming-")`, `startsWith("optimistic-user-")`) prevent double-bridging ✅
- `animationBridge.ts` Set is module-level (no prop drilling needed) and bounded by message count in practice ✅
- `doc-impact` check: SUCCESS — no doc changes needed ✅

**Minor note**: PR says "Related to #462" — you may want to use "Closes #462" so the issue auto-closes on merge.

No tests added, which is fine for a streaming timing fix — the behavior is hard to unit test and the existing suite covers the surrounding logic.

**→ Route to human review (targets `main`)**

slam-paws:
Superseded by combined PR #468

PR #472, opened 2026-05-09
[codex] Fix invitee start push copy

shantamg:
## Summary
- Fix invitation acceptance push actor resolution by sending `joinedBy` on `session.joined`.
- Update `session.joined` push copy to explicitly say the invitation was accepted.
- Override compact-sign push copy for the invitee-starts-after-acceptance case so the inviter does not get stale “begin when ready” language.

## Root Cause
When an invitee accepted and started their side, the later `partner.signed_compact` push reused the generic compact-ready message. For the inviter, that incorrectly implied they could begin the session, even though they were already in progress.

## Validation
- `npm test --workspace backend -- push.test.ts`

### DIFF
--- a/backend/src/controllers/invitations.ts
+++ b/backend/src/controllers/invitations.ts
@@ -680,6 +680,7 @@ export async function acceptInvitation(req: Request, res: Response): Promise<voi
     // Use notifyPartnerWithFallback to ensure Ably message is always published
     // and push is sent if user is offline
     await notifyPartnerWithFallback(invitation.sessionId, invitation.invitedById, 'session.joined', {
+      joinedBy: user.id,
       userId: user.id,
       userName: user.name,
     });
--- a/backend/src/services/__tests__/push.test.ts
+++ b/backend/src/services/__tests__/push.test.ts
@@ -161,6 +161,83 @@ describe('Push Notification Service', () => {
       ]);
     });
 
+    it('uses invitation acceptance copy when the invitee starts their side', async () => {
+      (prisma.user.findUnique as jest.Mock).mockResolvedValue({
+        pushToken: validPushToken,
+      });
+      (prisma.invitation.findFirst as jest.Mock).mockResolvedValue({
+        invitedById: testUserId,
+      });
+      (prisma.session.findUnique as jest.Mock).mockResolvedValue({
+        relationship: {
+          members: [
+            {
+              userId: testUserId,
+              nickname: null,
+              user: { firstName: 'Jason', name: 'Jason Person' },
+            },
+            {
+              userId: 'actor-123',
+              nickname: null,
+              user: { firstName: 'Shantam', name: 'Shantam Person' },
+            },
+          ],
+        },
+      });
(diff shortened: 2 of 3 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes push notification copy for the invitation acceptance flow: resolves the actor name for `session.joined`, updates its body to explicitly mention invitation acceptance, and overrides `partner.signed_compact` copy when the invitee starts their side so the inviter doesn't get stale "begin when ready" language.

## Feedback
- **Correctness**: Logic in `getPushMessageTemplate` is sound — correctly checks that the actor is someone other than the recipient, then verifies the recipient is the inviter via `invitation.findFirst`. The `joinedBy` addition to `getActorUserId` follows the existing pattern cleanly.
- **Tests**: Good coverage — both the `session.joined` actor resolution and the `partner.signed_compact` invitation override cases are tested with clear assertions.
- **Performance (minor, non-blocking)**: `getPushMessageTemplate` adds a DB query for *every* `partner.signed_compact` push, even non-invitation ones. This is negligible at current scale but worth noting — if push volume grows, this could be cached or short-circuited.
- **Data shape**: `joinedBy` and `userId` are both set to `user.id` in `invitations.ts:683`. The duplication is intentional since `getActorUserId` doesn't check `userId`, but a comment noting why both are needed would help future readers. Non-blocking.

## Verdict
Approve — clean, well-tested fix for a real UX issue. The override approach via `getPushMessageTemplate` is a nice pattern that keeps the static template map intact while allowing context-aware overrides.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes push notification copy when an invitee accepts and starts their side of a session. Adds `joinedBy` to the `session.joined` event payload for actor resolution, updates the `session.joined` body to mention invitation acceptance, and overrides the `partner.signed_compact` template when the recipient is the inviter — so they see "accepted your invitation" instead of the generic "begin when ready" copy.

## Feedback
- **Good**: The `getPushMessageTemplate` approach cleanly separates the invitation-aware override from the static template map without complicating the existing code path.
- **Good**: Two targeted tests covering both the `partner.signed_compact` override and the `session.joined` actor resolution — nice coverage of the two distinct scenarios.
- **Minor — redundant field** (`invitations.ts:683-684`): `joinedBy` and `userId` are both set to `user.id`. The `userId` field was already there before this PR, and `getActorUserId` doesn't check `userId` — so `userId` is effectively dead data in the push path. Not blocking, but worth cleaning up in a follow-up to avoid confusion about which field is authoritative.
- **Minor — select efficiency** (`push.ts:232`): The `invitation.findFirst` query selects `invitedById` but never reads it from the result — only the existence of the row matters. `select: { id: true }` (or just `_count`) would be marginally cleaner. Very minor.

## Verdict
**Approve** — Correct fix, good tests, clear commit message. The two minor items above are non-blocking suggestions for a future cleanup.

PR #473, opened 2026-05-09
[codex] Add nuanced push notification copy

shantamg:
## Summary
- Add state-aware push copy for partner progress notifications so “your turn” copy is only used as softer “Ready to continue” language when the recipient was actually waiting.
- Replace user-facing “unblocked”/“your turn” style language with approved copy such as “Ready to continue,” “is progressing,” and “Review needed.”
- Deduplicate notification tap navigation so tapping a notification for the currently open session uses replace instead of stacking a duplicate session screen.

## Root Cause
Partner progress events used one push template regardless of recipient state. In parallel stages, that could tell someone it was their turn even though they were already active. Notification taps also always pushed a session route, so reopening the same session from a push could add a duplicate route above itself.

## Validation
- `npm test --workspace backend -- push.test.ts`
- `npm run check --workspace backend`
- `npm run check --workspace mobile`

Note: validation was run in the main workspace. The clean PR worktree does not have node_modules installed.

### DIFF
--- a/backend/src/services/__tests__/push.test.ts
+++ b/backend/src/services/__tests__/push.test.ts
@@ -60,7 +60,7 @@ describe('Push Notification Service', () => {
     });
 
     it('has appropriate messages for agreement events', () => {
-      expect(PUSH_MESSAGES['agreement.proposed'].title).toContain('review');
+      expect(PUSH_MESSAGES['agreement.proposed'].title).toContain('agreement');
       expect(PUSH_MESSAGES['agreement.confirmed'].title).toContain('confirmed');
     });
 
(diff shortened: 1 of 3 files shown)


slam-paws (commented):
## PR Review: [codex] Add nuanced push notification copy

### Summary
Three changes in one PR: (1) state-aware push copy for `partner.stage_completed` that varies based on recipient's progress, (2) tone refresh across all push templates removing "your turn" / "open it to continue" language, and (3) deduplicating notification tap navigation on mobile.

### Verdict: Approve with minor observations

This is clean, well-structured work. The copy changes are a clear improvement — removing urgency/directive language ("It is your turn", "Open the session to continue") in favor of autonomous pacing ("when you are ready", "at your own pace"). The state-aware branching for stage completion is well thought out.

---

### Observations

**1. `PUSH_MESSAGES` table drift (low risk)**
The `PUSH_MESSAGES` record is still keyed by every `SessionEvent`, but `partner.stage_completed`, `partner.advanced`, `partner.ready_to_rank`, and `empathy.context_shared` are now overridden in `getPushMessageTemplate()` before the fallback is reached. The entries in `PUSH_MESSAGES` for those events are now dead code — they'll never be used at runtime. This is fine functionally (the fallback at the bottom catches them), but it could confuse a future reader into thinking those templates are live.

Consider either: (a) leaving a comment on the dead entries noting they're overridden, or (b) removing them from `PUSH_MESSAGES` and handling the type constraint differently (e.g. `Partial<Record<...>>`). Not blocking.

**2. `partner.stage_completed` default template vs. `PUSH_MESSAGES` entry**
The default template in `PUSH_MESSAGES` still says `'{actor} finished {stageName}'` / `'It is your turn to continue.'` (the old copy). Since `getPushMessageTemplate` always intercepts `partner.stage_completed` before reaching the fallback, this old copy is never shown — but if someone later refactors the override away, the old tone would resurface silently.

**3. `latestProgressForUser` cast**
`push.ts:233` — The `filter` inside `latestProgressForUser` casts each entry to `StageProgressSnapshot & { userId?: string }`. The `StageProgressSnapshot` interface doesn't include `userId`, but the Prisma query result does. The cast works but is a bit fragile. Since `getSessionProgressContext` already returns `Array<StageProgressSnapshot & { userId: string }>`, you could type the parameter of `latestProgressForUser` as that union type directly instead of casting inside the filter.

**4. Mobile: `useEffect` dependency on `currentSessionId`**
`useNotifications.ts` — The effect now depends on `currentSessionId`, which means the notification-response listener is torn down and recreated every time the user navigates to/from a session screen. This is correct for the replace-vs-push logic, but worth noting that `useGlobalSearchParams` will cause re-renders on any param change, not just session navigation. In practice this should be fine since the effect is lightweight (just a listener swap), but something to watch if notification registration ever becomes heavier.

**5. Test coverage is good**
Three new test cases cover the three `partner.stage_completed` branches (default, recipient in empathy, recipient waiting). The existing test assertion was updated to match new copy. Solid coverage.

### No blocking issues found. Looks good to merge.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Solid PR that replaces generic "your turn" push copy with state-aware, softer language and deduplicates notification tap navigation. The approach of querying `stageProgress` to select the right template per recipient state is well thought out.

## Feedback

- **Good pattern (push.ts)**: The `getStageCompletedTemplate` function cleanly branches on recipient state (waiting, working, default) to select appropriate copy. This is a nice extension point for future stages.

- **Good pattern (useNotifications.ts)**: Using `router.replace` when the session is already open avoids stacking duplicate screens. Clean implementation with `normalizeParam` handling the array-or-string param shape.

- **Minor — type cast (push.ts:258–260)**: `latestProgressForUser` casts `progress` entries to `StageProgressSnapshot & { userId?: string }` to access `userId`, but the `findMany` result already includes `userId` in its `select`. The cast works, but it'd be slightly cleaner to define a local type (e.g., `StageProgressWithUser`) that includes `userId` rather than using an inline intersection with an optional field. Not blocking.

- **Minor — existing test not updated (push.test.ts:127–161)**: The "uses the recipient nickname" test still expects the old `"Sammy finished Your Story"` / `"It is your turn to continue."` copy. This test should be updated since the default `partner.stage_completed` template changed. If this test still passes, it's because the mock doesn't set up `stageProgress.findMany`, so `getStageCompletedTemplate` may fall through to the generic template — but the expected title `"Sammy finished Your Story"` no longer matches any template in the PR. **This test will likely fail.** Worth verifying.

- **Minor — `StageProgressSnapshot` interface vs Prisma type (push.ts:38–41)**: The interface duplicates what Prisma already provides. Fine for decoupling push logic from Prisma internals, just noting the trade-off.

- **Copy consistency**: All the new copy follows a consistent "Come back when you are ready" / "Keep going at your own pace" tone — nicely done. The shift away from urgency-inducing "your turn" language aligns well with the product philosophy.

## Verdict
**Approve** — The one concern is the existing nickname test (`push.test.ts:127–161`) which may now fail since it expects the old `partner.stage_completed` copy. Please verify that `npm test --workspace backend -- push.test.ts` passes with the full diff applied. If it fails, that test's expected values need updating to match the new state-aware templates.

PR #474, opened 2026-05-09
Fix partner last seen and biometric lock overlay

shantamg:
## Summary
- add `UserVessel.lastActiveAt` and return `partnerLastActiveAt` from session state
- update partner details to base Last seen on partner activity, with a fallback to the old viewed timestamp
- render the biometric lock as a top-level native modal so it covers drawers and other popups

## Root Cause
The partner details drawer used `partnerLastViewedAt`, which tracks read state rather than user activity. The biometric lock was an absolutely positioned view, so React Native modal drawers could render above it despite its high z-index.

## Validation
- `npm --workspace backend run prisma:generate`

### DIFF
--- a/backend/prisma/migrations/20260509000000_add_user_vessel_last_active_at/migration.sql
+++ b/backend/prisma/migrations/20260509000000_add_user_vessel_last_active_at/migration.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "UserVessel"
+ADD COLUMN "lastActiveAt" TIMESTAMP(3);
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -366,6 +366,9 @@ model UserVessel {
   // When the user last viewed the Share/Partner tab
   // Used to determine "seen" status for shared content (empathy, context)
   lastViewedShareTabAt DateTime?
+  // When the user last actively did something in this session.
+  // Used for partner-facing presence details; distinct from read/view state.
+  lastActiveAt         DateTime?
   // When this user removed the session from their own conversation list.
   // The partner may still retain their own vessel and access.
   archivedAt           DateTime?
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -25,7 +25,7 @@ import {
 } from '@meet-without-fear/shared';
 import { notifyPartner, publishSessionEvent, notifySessionMembers, publishMessageAIResponse, publishMessageError, publishTopicFrameUpdated } from '../services/realtime';
 import { successResponse, errorResponse } from '../utils/response';
-import { getPartnerUserId, isSessionCreator } from '../utils/session';
+import { getPartnerUserId, isSessionCreator, touchUserSessionActivity } from '../utils/session';
 import { embedSessionContent } from '../services/embedding';
 import { updateSessionSummary, getSessionSummary } from '../services/conversation-summarizer';
 import { runReconcilerForDirection, getSharedContextForGuesser } from '../services/reconciler';
@@ -1288,6 +1288,12 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
         stage: currentStage,
       },
     });
+    await touchUserSessionActivity(sessionId, user.id, userMessage.timestamp);
+    publishSessionEvent(sessionId, 'partner.activity', {
+      activeAt: userMessage.timestamp.toISOString(),
+    }, user.id).catch((err) =>
+      logger.warn(`[sendMessageStream:${requestId}] Failed to publish partner activity:`, err)
+    );
     logger.info(`[sendMessageStream:${requestId}] User message created: ${userMessage.id}`);
 
     // Broadcast to Status Site
--- a/backend/src/controllers/session-state.ts
+++ b/backend/src/controllers/session-state.ts
@@ -33,6 +33,17 @@ interface CompactGates {
   signedAt?: string;
 }
 
+function latestDate(...values: Array<Date | null | undefined>): Date | null {
+  let latest: Date | null = null;
+  for (const value of values) {
+    if (!value) continue;
+    if (!latest || value > latest) {
+      latest = value;
+    }
+  }
+  return latest;
+}
+
 async function acceptPendingInvitationForE2ESession(sessionId: string, userId: string): Promise<boolean> {
   if (process.env.E2E_AUTH_BYPASS !== 'true' || process.env.NODE_ENV === 'production') {
     return false;
@@ -192,6 +203,7 @@ export async function getSessionState(req: Request, res: Response): Promise<void
         },
         select: {
           lastViewedAt: true,
+          lastActiveAt: true,
           lastSeenChatItemId: true,
         },
       }),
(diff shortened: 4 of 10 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Adds `UserVessel.lastActiveAt` to track partner activity separately from read-state (`lastViewedAt`), computes `partnerLastActiveAt` from multiple signals (lastActiveAt, last message, last stage progress), and wraps BiometricLockOverlay in a native `<Modal>` so it renders above drawers and other popups.

## Feedback

- **Correctness**: Clean separation of activity vs. read-state. The `latestDate` helper, `touchUserSessionActivity` upsert, and real-time event plumbing all look correct. The BiometricLockOverlay guard at line 72 (`!isLocked && lockState === 'unlocked'`) ensures the Modal only mounts when needed.

- **Architecture**: Good use of a dedicated `partner.activity` event type and the shared DTO updates. The fallback chain (`partnerLastActiveAt ?? partnerLastViewedAt`) on the mobile side is a nice graceful degradation for sessions that predate the migration.

- **Performance (minor)**: `getSessionState` now runs two additional sequential DB queries (`partnerLastMessage`, `partnerLastStageProgress`) after the existing `partnerVessel` query. Since all three are independent, they could be parallelized with `Promise.all` to avoid adding ~2 extra round trips to a hot endpoint. Not blocking, but worth considering.

- **Database**: Proper migration file — no `db push`. 👍

- **Tests (nice-to-have)**: `latestDate` and `touchUserSessionActivity` are simple, testable units. Adding coverage would be cheap insurance.

## Verdict
Approve — clean, well-structured change with proper migration, type updates, and real-time plumbing. The sequential query note is a minor optimization opportunity, not a blocker.

slam-paws (approved):
## Review: Approve

Clean PR with two well-scoped fixes.

### Partner Last Seen (activity tracking)

**Approach**: Adds `UserVessel.lastActiveAt` and computes `partnerLastActiveAt` as the latest of three signals: the new column, partner's last message, and partner's last stage progress. The mobile client falls back to `partnerLastViewedAt` when `partnerLastActiveAt` is absent.

- Migration, schema, DTO, and realtime event type all stay in sync — good.
- `touchUserSessionActivity` uses `upsert`, correctly handles the case where a `UserVessel` row doesn't exist yet.
- New queries in `getSessionState` for `partnerLastMessage` and `partnerLastStageProgress` are covered by existing indexes (`[sessionId, forUserId, role]` on Message, `[sessionId, userId, stage]` unique on StageProgress) — no performance concern.
- `latestDate` helper is clean; null-safe.
- `partner.activity` event + cache update in `UnifiedSessionScreen` follows existing patterns (`setQueryData`, not `invalidateQueries`).

### Biometric Lock Overlay

**Approach**: Wraps the lock screen in a native `<Modal>` so it renders above RN modal drawers that could previously appear on top of it.

- The early return on line 72 (`if (!isLocked && lockState === 'unlocked') return null`) ensures the Modal only mounts when the lock is active — no risk of blocking the app when unlocked.
- `onRequestClose={() => {}}` correctly prevents Android back-button dismissal.
- `presentationStyle="overFullScreen"` + `statusBarTranslucent` ensures full coverage on iOS and Android.

No issues found. LGTM.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Adds `UserVessel.lastActiveAt` to track actual user activity (distinct from read-state), returns `partnerLastActiveAt` from session state for more accurate "Last seen" display, and wraps the biometric lock overlay in a native `<Modal>` so it renders above drawers and popups.

## Feedback

- **Correctness**: Logic is solid throughout. The `latestDate` helper, `touchUserSessionActivity` upsert, and the fallback chain (`partnerLastActiveAt ?? partnerLastViewedAt`) all look correct.
- **Architecture**: Good separation — `lastActiveAt` is intentionally distinct from `lastViewedAt` (read-state). The new `partner.activity` realtime event follows existing patterns.
- **Database**: Proper migration file (not `db push`). Nullable column addition is safe for existing rows.
- **Mobile**: The `<Modal>` approach with `presentationStyle="overFullScreen"` is the right fix — z-index alone can't beat native modals/drawers. The early return at line 72 ensures the Modal only mounts when locked. `onRequestClose={() => {}}` correctly prevents Android back-dismiss.
- **Cache-first compliance**: Ably handlers use `setQueryData` (not `invalidateQueries`) per the project conventions.
- **Performance (minor)**: In `session-state.ts`, `partnerLastMessage` and `partnerLastStageProgress` are independent queries that could be wrapped in `Promise.all` alongside `partnerVessel` to run in parallel instead of sequentially. Not blocking — the queries are small index lookups — but worth considering if this endpoint's latency matters.
- **Style**: `StyleSheet.create` is used (not NativeWind) but this is consistent with the existing file, so no concern.

## Verdict
Approve — clean, well-motivated changes with proper migration, type updates, and realtime event plumbing.

slam-paws (commented):
## Review — PR #474: Fix partner last seen and biometric lock overlay

**Overall**: Clean, well-structured PR. Two distinct fixes that are both straightforward. The migration, shared types, and DTO updates are all consistent. A couple of performance nits below.

### Partner Last Active tracking

- **Migration + schema**: Looks good. Proper migration, nullable column, no default needed.
- **`touchUserSessionActivity`**: Clean upsert. Good separation from `lastViewedAt`.
- **`partner.activity` event**: Properly added to the `SessionEventType` union and handled in the mobile Ably listener.
- **Mobile cache updates**: `updatePartnerLastActiveAt` correctly patches the query cache via `setQueryData` (not `invalidateQueries`), consistent with the project's cache-first rules.
- **Fallback chain**: `partnerLastActiveAt ?? partnerLastViewedAt` on mobile is a good backward-compat strategy for sessions that predate this migration.

#### Nits / suggestions (non-blocking)

1. **`getSessionState` — sequential queries could be parallelized.** The three new queries (`partnerVessel`, `partnerLastMessage`, `partnerLastStageProgress`) are independent and currently `await`ed sequentially. Wrapping them in `Promise.all([...])` would save ~2 round-trips on every session state fetch.

2. **`sendMessageStream` — `touchUserSessionActivity` on the hot path.** This `await` adds latency to every message send. Since the `publishSessionEvent` right after it already uses `.catch()` (fire-and-forget), the touch could follow the same pattern — the data is not critical to the response.

### Biometric lock → Modal

Good fix. `presentationStyle="overFullScreen"` is the right way to ensure the lock screen covers RN modal drawers. The `<Modal visible>` is unconditionally true, which is fine since the component is only mounted when locked (via the context guard).

No blocking issues.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Adds `UserVessel.lastActiveAt` to track partner activity (distinct from read/view state), surfaces it as `partnerLastActiveAt` in session state, and wraps the biometric lock overlay in a native `<Modal>` so it renders above drawers.

## Feedback

- **[Performance]**: The PR adds two new sequential queries (`partnerLastMessage`, `partnerLastStageProgress`) after the existing `partnerVessel` query in `session-state.ts:285-310`. All three partner queries are independent and could be parallelized with `Promise.all()`:
  ```ts
  const [partnerVessel, partnerLastMessage, partnerLastStageProgress] = await Promise.all([
    partnerId ? prisma.userVessel.findUnique(...) : null,
    partnerId ? prisma.message.findFirst(...) : null,
    partnerId ? prisma.stageProgress.findFirst(...) : null,
  ]);
  ```
  Not blocking since `getSessionState` already has several sequential queries, but worth doing for a quick latency win.

- **[Performance – nice-to-have]**: `partnerLastMessage` does a `findFirst` ordered by `timestamp DESC` on the `message` table filtered by `sessionId + senderId + role`. If this endpoint is hot, an index on `(sessionId, senderId, role, timestamp)` would help. Low priority unless you see slow queries.

- **[Architecture]**: `latestDate()` helper in `session-state.ts:36-44` is clean and well-scoped. Good call keeping it local to the controller rather than over-abstracting.

- **[Correctness]**: The `updatePartnerLastActiveAt` callback in `UnifiedSessionScreen.tsx:689-710` correctly falls back to `viewedAt` when `activeAt` is missing, giving backward compatibility with older server versions. Nice touch.

- **[Mobile]**: Wrapping in `<Modal visible transparent={false} presentationStyle="overFullScreen">` is the right fix — the early return on line 72 (`if (!isLocked && lockState === 'unlocked') return null`) means the Modal only mounts when needed, so hardcoded `visible` is correct.

- **[Database]**: Proper migration file used (not `db push`). Schema comments clearly explain the distinction from `lastViewedAt`. ✓

- **[Shared types]**: `SessionStateResponse` and `SessionEventType` updated in `shared/` to include `partnerLastActiveAt` and `partner.activity`. ✓

## Verdict
Approve — clean, well-structured change. The parallelization suggestion is a minor optimization, not a blocker.

Issue #496, opened 2026-05-10
Include topicFrame in stage 1–4 prompt context


slam-paws:
## Provenance

- **Channel:** #bugs-and-requests
- **Requester:** Shantam (<@U0AD3TF2U7L>)
- **Original message:** "I just realized that the topic of the conversation doesn't seem to be included in the context of the prompt for the stages. Can you confirm whether this is the case? The AI should definitely have the topic."

## Problem

The conversation topic (`topicFrame`) is **not** included in the prompt context for stages 1–4. Currently it is only used in:

1. **Stage 0** — where the topic is first articulated via `buildInvitationPrompt`
2. **`buildInitialMessagePrompt`** — the welcome message shown to the invitee, where the prompt includes `TOPIC ALREADY SHOWN ABOVE THIS MESSAGE: "{topicFrame}"`

The `PromptContext` interface (`stage-prompts.ts:410–477`) has no `topicFrame` field. The context assembler (`context-assembler.ts`) also does not include it. This means the AI has no awareness of the agreed-upon topic during the core conversation stages.

## Fix

1. Add `topicFrame?: string | null` to the `PromptContext` interface in `stage-prompts.ts`
2. Pass `session.topicFrame` when constructing prompt context in the message controller / orchestrator
3. Include the topic in each stage's system prompt (e.g., "The topic of this conversation is: {topicFrame}")
4. Only include when `topicFrame` is non-null and confirmed (`topicFrameConfirmedAt` is set)

## Files

- `backend/src/services/stage-prompts.ts` — `PromptContext` interface + `buildStage1Prompt` through `buildStage4Prompt`
- `backend/src/controllers/messages.ts` — where prompt context is assembled before calling the orchestrator
- `backend/src/services/ai-orchestrator.ts` — where `PromptContext` is passed to stage prompt builders

slam-paws:
## Fix proposed

**PR**: #501
**What changed**: Threaded `topicFrame` through the full prompt pipeline (FullAIContext → OrchestratorContext → PromptContext → dynamic block). The AI now sees `CONVERSATION TOPIC: "{topicFrame}"` at the top of the dynamic guidance for stages 1–4, but only when the topic is confirmed (`topicFrameConfirmedAt` is set).
**How to verify**: Start a session, confirm a topic, advance to stage 1+, and check the system prompt in Brain Activity logs for the `CONVERSATION TOPIC:` line.

This issue will stay open until a human verifies the fix and adds the `user-verified` label.

PR #501, opened 2026-05-10
fix: include topicFrame in stage 1–4 prompt context

slam-paws:
## Changes
- Fix: Thread `topicFrame` through the prompt pipeline so AI has awareness of the conversation topic during stages 1–4
- Add: `topicFrame` field to `PromptContext`, `OrchestratorContext`, and `FullAIContext` interfaces
- Add: Topic injection in `buildBaseDynamicGuidance` — appears as `CONVERSATION TOPIC: "{topicFrame}"` at the top of the dynamic block
- Guard: Only included when `topicFrameConfirmedAt` is set (prevents unconfirmed drafts leaking into prompts)
- Test: 3 new tests verifying topic presence/absence in dynamic block

Related to #496

## Provenance
- **Requested by:** Shantam (<@U0AD3TF2U7L>)
- **Original message:** "I just realized that the topic of the conversation doesn't seem to be included in the context of the prompt for the stages. Can you confirm whether this is the case? The AI should definitely have the topic."
- **Prompt(s) used:** general-pr

## Docs updated
- `docs/backend/prompting-architecture.md` — added topicFrame to dynamic elements list

### DIFF
--- a/backend/src/controllers/sessions.ts
+++ b/backend/src/controllers/sessions.ts
@@ -982,6 +982,7 @@ export async function confirmInvitationMessage(req: Request, res: Response): Pro
       isInvitationPhase: false,
       isStageTransition: true,
       previousStage: 0,
+      topicFrame: session.topicFrameConfirmedAt ? session.topicFrame : undefined,
     };
 
     // Notify session that invitation was confirmed
--- a/backend/src/services/__tests__/stage-prompts.test.ts
+++ b/backend/src/services/__tests__/stage-prompts.test.ts
@@ -1200,6 +1200,25 @@ describe('Stage Prompts Service', () => {
     });
   });
 
+  describe('topicFrame in dynamic guidance', () => {
+    it('includes topicFrame in dynamic block for stages 1–4 when provided', () => {
+      for (const stage of [1, 2, 3, 4]) {
+        const blocks = buildStagePrompt(stage, createContext({ topicFrame: 'Mealtime poking' }));
+        expect(blocks.dynamicBlock).toContain('CONVERSATION TOPIC: "Mealtime poking"');
+      }
+    });
+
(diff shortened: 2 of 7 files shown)


slam-paws:
## Bot review — LGTM ✅

**PR:** fix: include topicFrame in stage 1–4 prompt context
**Issue:** #496

### Summary

Clean, well-scoped fix. The topicFrame is correctly threaded through the full pipeline and the guard is solid.

### Checklist

**Correctness** ✅
- Data flow is correct: `session-processor.ts` → `ai.ts` → `ai-orchestrator.ts` → `buildStagePrompt` → `buildBaseDynamicGuidance`
- Guard `session.topicFrameConfirmedAt ? session.topicFrame : undefined` is correct — prevents unconfirmed drafts from leaking into prompts
- Stage 0 is safe: `topicFrameConfirmedAt` is only set on stage 0→1 transition, so the topic never appears during stage 0 conversations
- `if (context.topicFrame)` in `buildBaseDynamicGuidance` handles both `null` and `undefined` correctly

**Completeness** ✅
- All four requirements from issue #496 addressed
- Both callers of `getOrchestratedResponse` that run during active stages are patched (`session-processor.ts` and `sessions.ts`/confirmInvitationMessage)
- `messages.ts` line 1108 passes `topicFrame` without the guard — but that's the existing `buildInitialMessagePrompt` path (Stage 0 welcome message), not changed by this PR and not part of the issue scope

**Tests** ✅
- 3 tests: topic present for stages 1–4, omitted when null, omitted when undefined
- `createContext()` override pattern is standard and correct

**Docs** ✅
- `prompting-architecture.md` updated with accurate description of the injection format and guard condition

**Security** ✅
- No injection risks; topicFrame is a user-defined string already stored in the DB, injected as a string literal in the prompt

No issues found. Ready to merge.

Issue #502, opened 2026-05-10
Web app: native-app download banner not visible after redesign


slam-paws:
## Bug

The \"Get the native app for a better experience\" banner at the top of the web app (app.meetwithoutfear.com) is no longer visible after the recent design system overhaul.

## Investigation Notes

The component (`NativeAppBanner.tsx`) still exists and IS rendered in `mobile/app/_layout.tsx:124` inside the `AppShell`. However, it appears to not be showing on web. Likely causes:

1. **Palette mismatch** — The banner uses hardcoded colors from the old `theme/colors.ts` (dark-only: `bgSecondary: '#14171c'`). In light mode, these dark colors may render but be immediately covered by the Stack's opaque `contentStyle.backgroundColor`.

2. **Stack overlay on web** — Expo Router's Stack on web may render screens with absolute positioning that covers the banner area. The banner sits outside `GestureHandlerRootView` but the Stack's content may fill the entire frame.

3. **overflow: hidden on webFrame** — The `webFrame` style has `overflow: 'hidden'` which could clip the banner if the Stack expands to fill the frame.

## Steps to Reproduce

1. Visit app.meetwithoutfear.com in a browser
2. Expected: banner strip at top saying "Get the native app for a better experience"
3. Actual: no banner visible

## Fix Direction

- Verify the banner renders on web by inspecting the DOM
- If it renders but is covered, adjust z-index or layout so the banner sits above the Stack
- Update the banner to use the new palette system (`useAppAppearance()`) for light/dark mode support

## Provenance

- Channel: #bugs-and-requests
- Requester: Shantam (U0AD3TF2U7L)
- Message: "In the recent redesign of the app we lost the banner at the top when on the web which invites to use the app for a better experience. Can you look in the git history for that and reinstate it?"

slam-paws:
## Investigation Report

**Root cause**: The `webFrame` container in `_layout.tsx` has `position: 'relative'`, making it the CSS positioning context for all absolutely-positioned descendants. Expo Router's Stack screens on web use absolute positioning and anchor to this context, filling the entire `webFrame` — including the area above `GestureHandlerRootView` where the banner sits. Commit `c53c346` (Polish home chat, drawers, and privacy settings) added `contentStyle: { backgroundColor: palette.bg }` to the Stack's `screenOptions`, making every screen background opaque. The banner still renders in the DOM but is completely hidden behind the opaque, absolutely-positioned screen layer.

**Evidence**:
[...]

PR #503, opened 2026-05-10
fix: restore web banner and harden auth recovery

slam-paws:
## Changes
- Fix: restored the native-app download banner on web by anchoring Stack screens under `GestureHandlerRootView` and moving the banner to the dynamic appearance palette.
- Fix: prevented the OAuth callback from briefly showing a false "Sign-in failed" state while Clerk's web auth state settles.
- Fix: added a first-party web session marker so clearing app site data causes the app to reject a cookie-only recovered Clerk session and sign out instead.
- Fix: empty/stripped AI streaming responses now fail the turn and use the frontend retry/error state instead of saving a misleading canned process explanation.

## Root Cause
- Expo Router Stack screens were absolutely positioned relative to the web frame and covered the banner.
- `sso-callback.web.tsx` checked `clerk.session` immediately after `handleRedirectCallback()`, but Clerk Expo web can update `isSignedIn` one render later.
- Clerk web can recover a session from its own browser/client state even when first-party app storage has been cleared, so the app needed its own trust marker for site-data-cleared recovery.
- The streaming message controller previously converted empty model output into a generic process explanation fallback, which hid the real failure and created confusing chat history.

## Validation
- `npm run check` in `mobile`
- `npm run check --workspace=backend`

Related to #502

## Docs updated
No doc changes needed — web UI/auth behavior and backend streaming failure handling fixes.

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -1986,41 +1986,15 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
       }
 
       // Guard: empty AI response after parsing + dispatch means the model emitted
-      // content entirely inside tags we couldn't route (e.g. an unknown dispatch
-      // tag with no user-facing text). Try a contextual dispatch as EXPLAIN_PROCESS
-      // before falling back to a static message (issue #312 — static fallback was
-      // producing response loops).
+      // no usable user-facing text, or the upstream stream failed without
+      // producing text. Treat this as a failed turn so the frontend can show its
+      // retry/error state instead of saving a misleading canned response.
       if (!accumulatedText.trim()) {
-        logger.warn(`[sendMessageStream:${requestId}] Empty AI response after tag stripping — attempting contextual fallback`, {
+        logger.error(`[sendMessageStream:${requestId}] Empty AI response after tag stripping`, {
           dispatchTag: dispatchTag ?? null,
           scrubbedPlannerText: scrubbedResponse.scrubbed,
         });
-        // Try to generate a context-aware response via the process explainer
-        let fallback: string | null = null;
-        try {
-          const fallbackContext: DispatchContext = {
-            userMessage: content,
-            conversationHistory: history.map((m) => ({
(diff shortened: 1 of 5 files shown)


slam-paws:
## Bot review — PR #503

**Verdict: LGTM ✅**

### Summary

Solid fix. Both root causes from the investigation were addressed:

1. **Stack overlay** — Adding `position: 'relative'` via `Platform.select({ web: ... })` to the root container establishes the positioning context so Expo Router's absolutely-positioned Stack screens anchor there, not at the `webFrame` level. Correct approach.
2. **Palette migration** — Banner updated from hardcoded dark-only `colors.ts` to `useAppAppearance()` palette. All five color references migrated (`bgElev`, `border`, `accent`, `accentText`, `text`, `bg`, `textMuted`, `selected`). Light/dark mode will work correctly now.
3. **zIndex: 1** — Extra stacking safety, non-disruptive.

### Checklist

- [x] **Correctness**: addresses both identified root causes (positioning context + color hardcoding)
- [x] **Completeness**: all requirements from #502 addressed; PR description accurate
- [x] **Conventions**: `useAppAppearance()` + `Platform.select` — matches patterns used elsewhere post-redesign
- [x] **Security**: UI-only change, no issues
- [x] **No accidental changes**: only `_layout.tsx` + `NativeAppBanner.tsx`, both on-target
- [x] **Tests**: layout/visual change, no behavioral logic — test coverage N/A
- [x] **Doc impact**: `docs/architecture/structure.md` doesn't mention `NativeAppBanner`, so no update needed despite the automated flag

### Minor note

`palette.accentText` used for the pressed button state (replacing old `colors.accentHover`). In light mode this is `#8a4f19` vs `#b7742f` (darker = ✅ pressed feedback). In dark mode it's lighter, which is slightly unconventional but functional. No palette `accentHover` exists in the new system, so this is the best available substitution.

Not blocking — ready for human review before merge to `main`.

shantamg (approved):

PR #504, opened 2026-05-10
[codex] Refine support exercise modal layout

shantamg:
## Summary
- simplify and restyle the 4-7-8 breathing start and active states
- constrain support exercise modals and shared intensity checks to the app width on web
- align grounding and body scan exercise styling with the shared appearance palette and CTA sizing

## Root cause
React Native Web renders Modal content through a portal under document.body, outside the app web frame. Modal pages could therefore stretch to the full browser viewport even though the app shell itself was capped.

## Validation
- npm test -- --runTestsByPath src/components/__tests__/BreathingExercise.test.tsx --runInBand
- npm run check


slam-paws (commented):
## PR Review: Refine support exercise modal layout

**Overall**: Clean migration from the legacy `colors` import to the dynamic `palette` / `designFonts` / `appWidthStyle` system. The modal-constraint CSS for web is a good fix for the portal-escape problem, and the breathing exercise redesign is well thought out. Tests are updated to match. Approve-worthy with one issue to address.

---

### Issue — hardcoded `#fffaf0` bypasses the palette

`#fffaf0` appears as a literal in **four files** (button text, step-number text) where it replaced `colors.textOnAccent`. This color doesn't exist as a palette token, so it won't adapt if a future theme or dark-mode adjustment changes the accent background.

The existing palette has `accentText` (a contrast color for accent surfaces). If `accentText` isn't the right semantic fit, consider adding a dedicated `textOnAccent` token to the palette — either way, a token reference is better than a magic hex repeated across files.

**Files**: `BodyScanExercise.tsx`, `BreathingExercise.tsx`, `GroundingExercise.tsx`, `IntensityCheck.tsx`

---

1. **`makeStyles` re-creation on every render when palette is stable** — `useMemo(() => makeStyles(palette), [palette])` is fine, but `StyleSheet.create` is cheap and `palette` is referentially stable per theme, so this is a no-op most of the time. No action needed, just noting the pattern is consistent.

2. **`!important` density in the modal CSS** — the `[aria-modal="true"] > div` rule uses seven `!important` declarations. This is understandable given the portal constraint, but if more modal variants appear it could become brittle. Worth a comment in code explaining *why* `!important` is necessary (RN Web portal renders outside the app frame).

3. **`stepActive` border change** — `BodyScanExercise` and `GroundingExercise` both move `borderWidth: 1` from `stepActive` into the base `step` style and only override `borderColor` in `stepActive`. This is correct and avoids the layout shift from adding a border on activation — good catch.

4. **Breathing exercise animation** — the `MIN_CIRCLE_SCALE = 0.18` → `1` expand is a much more dramatic visual than the old `1` → `1.3`. The `breatheOpacity` fade during hold is a nice touch. The `ready` state redesign (dot + instruction text instead of "Ready" inside the circle) reads better.

---

**Verdict**: Approve once the hardcoded `#fffaf0` is replaced with a palette token. Everything else looks solid.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Migrates all support exercise components (Breathing, Grounding, Body Scan, IntensityCheck) from the legacy static `colors` object to the dynamic `useAppAppearance()` palette, aligns styling with the shared appearance palette (fonts, border radii, CTA sizing), and fixes RN Web modal width overflow by constraining `aria-modal` portals to `APP_MAX_WIDTH`.

## Feedback
- **Good pattern**: The `makeStyles(palette)` + `useMemo` approach is the right way to handle dynamic theming with `StyleSheet.create`. Palette changes are infrequent so the memoization dependency is well-chosen.
- **Good fix**: The CSS override for `[aria-modal="true"]` in `_layout.tsx:83-93` is a pragmatic solution for the RN Web portal-outside-app-frame problem. The `!important` overrides are justified here since we're fighting portal-injected inline styles.
- **Good consolidation**: Replacing the local `MOBILE_MAX_WIDTH` with the shared `APP_MAX_WIDTH` export eliminates a magic-number duplication.
- **Nice UX refinement**: The breathing exercise redesign — starting with a small circle that expands on inhale, the "Breathe" text with opacity animation, and the calmer ready-state copy — feels more intentional and less clinical.
- **Minor note** (`IntensityCheck.tsx:16-22`): `hexToRgb` assumes 6-character hex input. All current palette values are 6-char so this is safe today, but if a shorthand hex (`#fff`) or `rgba()` value ever enters the palette, this would silently produce wrong colors. Not blocking — just worth a mental note.
- **Tests**: Good coverage additions for hold and exhale phase transitions. Updated assertions match the new UI text correctly.

## Verdict
Approve — clean theming migration with good test updates and a solid fix for the web modal overflow. No blocking issues.

Issue #510, opened 2026-05-10
Bug: partner name input refills after clearing (inner thoughts CTA)


slam-paws:
## Bug Report

When starting a new partner session from the inner thoughts CTA, the partner name is pre-filled in the input — that part works correctly. However, when the user deletes the last character of the pre-filled name, the entire name refills automatically. The user should be able to clear the name and have it stay blank.

**Steps to reproduce:**
1. Go to inner thoughts and tap the CTA to start a new partner session
2. The partner name input is pre-filled (correct)
3. Delete characters from the name — when the last character is removed, the full name snaps back

**Expected:** Clearing the name input should leave it empty.

## What We Know So Far

• The bug is in `mobile/app/(auth)/session/new.tsx` lines 92-100 — a `useEffect` that pre-fills the name has `firstName` in its dependency array, so when `firstName` becomes empty (`!firstName` is true), the effect re-runs and refills from the `partnerName` query param.
• Fix: remove `firstName` from the dependency array, or use a `hasPreFilled` guard so the effect only runs once.
• A similar pattern exists at lines 73-90 for `innerThoughtsContext` — worth checking that one too.

## Provenance

- **Channel**: DM (Shantam)
- **Reporter**: Shantam (U0AD3TF2U7L)
- **Original message**: When I start a new partner session from the inner thoughts CTA, it pre fills the name in the input. That's good but when I delete the last character of the filled in name, it fills in the whole thing again. I should be able to delete the name and it should stay blank.
- **Timestamp**: 1778429600.204669

slam-paws:
## Investigation Report

**Root cause**: The `useEffect` at `mobile/app/(auth)/session/new.tsx:92-100` has `firstName` in its dependency array. When the user clears the input (making `firstName` empty), the effect re-runs, the condition `partnerName && !firstName` becomes true again, and the name refills from the `partnerName` query param. This creates an infinite loop where the user can never clear the field.

**Evidence**: Code inspection confirms the dependency array `[partnerName, firstName]` causes re-execution whenever `firstName` changes. The `partnerName` query param remains constant (set during navigation), so any time `firstName` becomes falsy the pre-fill logic fires again.

**Affected file**: `mobile/app/(auth)/session/new.tsx` (lines 92-100)

**Severity**: Low — cosmetic UX bug, user can work around by typing a different name.

[...]

PR #512, opened 2026-05-10
fix: prevent partner name refilling after clear (#510)

slam-paws:
## Changes
- Fix: partner name input no longer snaps back after the user clears it
- Add a `useRef` guard (`hasPreFilledPartnerName`) so the pre-fill effect runs only once on mount
- Remove `firstName` from the effect's dependency array to prevent re-triggering

Related to #510

## Provenance
- **Channel:** DM (Shantam)
- **Original message:** When I start a new partner session from the inner thoughts CTA, it pre fills the name in the input. That's good but when I delete the last character of the filled in name, it fills in the whole thing again. I should be able to delete the name and it should stay blank.

## Docs updated
No doc changes needed — single-line bug fix in a React effect dependency array.

### DIFF
--- a/mobile/app/(auth)/session/new.tsx
+++ b/mobile/app/(auth)/session/new.tsx
@@ -12,7 +12,7 @@
  * and pre-fills the partner name if available.
  */
 
-import { useMemo, useState, useEffect } from 'react';
+import { useMemo, useState, useEffect, useRef } from 'react';
 import {
   View,
   Text,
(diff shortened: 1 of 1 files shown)


slam-paws:
Plan-alignment review: **ALIGNED**

Implementation matches requirements from #510:
- [x] `useRef` guard (`hasPreFilledPartnerName`) prevents re-fill after initial mount
- [x] `firstName` removed from dependency array — effect only re-runs if `partnerName` changes
- [x] `innerThoughtsContext` effect left untouched (investigation confirmed it's fine)
- [x] Scope: single file, minimal change, no unrelated modifications

No deviations detected. Ready for verification.

slam-paws:

Implementation matches requirements from #510:
- [x] Add `useRef` guard (`hasPreFilledPartnerName`) so pre-fill effect runs only once — implemented in `mobile/app/(auth)/session/new.tsx`
- [x] Remove `firstName` from effect dependency array to prevent re-triggering — implemented
- [x] Scope limited to the single affected `useEffect` — no unrelated changes
- [x] `innerThoughtsContext` effect left untouched (investigation confirmed it's fine)

Implementation follows the exact approach recommended in the investigation report. Ready for verification.

PR #517, opened 2026-05-10
Fix empathy notification timing copy

shantamg:
## Summary
- Change the empathy-submitted push notification so it no longer says the recipient can read the empathy before the reconciler finishes.
- Move the readable-empathy promise to the actual empathy reveal notification.
- Add a focused push notification test for the reveal copy.

## Root cause
`partner.empathy_shared` fires when a partner submits their empathy attempt, but that attempt may still be held behind reconciler analysis or a share-suggestion flow. The old push copy promised readable content too early.

## Validation
- `npm --workspace backend test -- src/services/__tests__/push.test.ts --runInBand`
- `npm --workspace backend run check` was attempted locally but blocked because the current dependency install cannot resolve `@clerk/express`.

### DIFF
--- a/backend/src/services/__tests__/push.test.ts
+++ b/backend/src/services/__tests__/push.test.ts
@@ -112,8 +112,8 @@ describe('Push Notification Service', () => {
         expect.objectContaining({
           to: validPushToken,
           sound: 'default',
-          title: 'Their empathy is ready',
-          body: 'Read what They understood about your experience.',
+          title: 'They shared empathy',
+          body: "We're checking it now and will show it when it is ready to read.",
           data: expect.objectContaining({
             screen: 'session',
             sessionId: testSessionId,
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes push notification timing: the `partner.empathy_shared` notification no longer promises readable empathy before the reconciler finishes. The "readable" copy moves to `empathy.revealed`, and a new test validates the reveal notification.

## Feedback
- **Correctness**: The copy changes are well-reasoned. `partner.empathy_shared` now honestly says the system is still processing, and `empathy.revealed` carries the readable-empathy promise. Good fix.
- **Tests**: New test for `empathy.revealed` covers the happy path with actor name resolution — solid.
- **Question (non-blocking)**: After this change, both `partner.empathy_revealed` (line 146: "{actorPossessive} empathy is visible / Open the session to read it.") and `empathy.revealed` (new: "Empathy exchange is ready / You can now read what {actor} understood...") offer similar "empathy is ready to read" copy. Are these sent to different recipients, or do they fire at different points in the flow? Not blocking — just want to make sure they don't overlap in a confusing way for users.
- **Note**: PR body mentions `npm run check` was blocked locally by a `@clerk/express` resolution issue — not a concern for this PR, but worth tracking separately.

## Verdict
Approve — clean, well-scoped fix with good test coverage.

PR #521, opened 2026-05-10
[codex] Show shared artifacts in chat

shantamg:
## Changes
- Show shared artifacts directly in the chat flow instead of hiding sent empathy/context behind drawer-only indicators.
- Add synthetic inline chat items for status-backed empathy/context artifacts when the backend does not provide a normal message for the current viewer.
- Preserve drawer navigation from shared indicators by scrolling to and highlighting the matching activity item.
- Label sent vs received shared artifacts so inline chat copy reads consistently for both participants.

## Provenance
- **Channel:** Codex desktop
- **Requested by:** Shantam
- **Original message:** "When I tap the line in the chat that says empathy shared, it opens the drawer, but it should also scroll to and highlight the corresponding item..." followed by "Please update the chat to have all things shared"

## Docs updated
- No doc changes needed — mobile UI behavior change only.

## Validation
- `npm run check --workspace @meet-without-fear/mobile`
- `git diff --check -- mobile/src/components/ActivityDrawer.tsx mobile/src/components/ChatBubble.tsx mobile/src/components/ChatInterface.tsx mobile/src/components/TimelineItemCard.tsx mobile/src/screens/UnifiedSessionScreen.tsx`

### DIFF
--- a/mobile/src/components/ActivityDrawer.tsx
+++ b/mobile/src/components/ActivityDrawer.tsx
@@ -408,7 +408,10 @@ export function ActivityDrawer({
   }, [visible, openDrawer]);
 
   useEffect(() => {
-    if (!visible || !focusMatch) return;
+    if (!visible || !focusMatch) {
+      setHighlightedItemId(null);
+      return;
+    }
 
     setHighlightedItemId(focusMatch.id);
 
--- a/mobile/src/components/ActivityDrawer.tsx
+++ b/mobile/src/components/ActivityDrawer.tsx
@@ -17,6 +17,7 @@ import {
   StyleSheet,
   BackHandler,
   useWindowDimensions,
+  type FlatList as FlatListType,
 } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { appWidthStyle } from '@/theme';
@@ -53,7 +54,14 @@ export interface ActivityDrawerProps {
   partnerAccepted?: boolean;
   sessionStatus?: string;
   partnerEmpathyValidated?: boolean;
+  focusTarget?: ActivityDrawerFocusTarget | null;
   testID?: string;
+}
+
+export interface ActivityDrawerFocusTarget {
+  type: 'context' | 'empathy';
+  direction: 'sent' | 'received';
+  timestamp?: string;
 }
 
 // ============================================================================
@@ -221,6 +229,7 @@ export function ActivityDrawer({
   partnerAccepted = false,
   sessionStatus,
   partnerEmpathyValidated,
+  focusTarget,
   testID = 'activity-drawer',
 }: ActivityDrawerProps) {
   const insets = useSafeAreaInsets();
@@ -248,8 +257,13 @@ export function ActivityDrawer({
     enabled: visible && isSessionActive,
   });
   const pendingActionsQuery = usePendingActions(sessionId);
-  const pendingActions = pendingActionsQuery.data?.actions ?? [];
+  const pendingActions = useMemo(
+    () => pendingActionsQuery.data?.actions ?? [],
+    [pendingActionsQuery.data?.actions],
+  );
   const { mutate: markShareTabViewed } = useMarkShareTabViewed(sessionId);
+  const listRef = useRef<FlatListType<TimelineItem>>(null);
+  const [highlightedItemId, setHighlightedItemId] = useState<string | null>(null);
 
   // Mark as viewed when drawer opens
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


slam-paws:
## Bot review

**PR:** [codex] Show shared artifacts in chat
**Scope:** `ActivityDrawer`, `ChatBubble`, `ChatInterface`, `TimelineItemCard`, `UnifiedSessionScreen`

### What this does

Removes the old filter that hid `EMPATHY_STATEMENT` and own `SHARED_CONTEXT` messages from chat, replacing it with inline synthetic items synthesized from `empathyStatusData`. Adds scroll-to + highlight in the activity drawer when a shared-artifact indicator is tapped. Fixes sent/received copy labels in `ChatBubble`.

---

### ✅ Correctness — mostly solid

- `focusMatch` logic (closest-by-timestamp fallback to last item) is correct
- `onScrollToIndexFailed` fallback is handled properly
- Label fixes (`"Empathy from [partner]"` / `"What you shared"`) are correct
- `onContextSharedPress` added to `renderItem` dependency array — good catch
- Dedup checks for synthetic messages look sound given server timestamps should match exactly

---

### 🐛 Changes needed

**1. `highlightedItemId` is never cleared when the drawer closes**

The `useEffect` in `ActivityDrawer` sets `highlightedItemId` on focus, but never clears it:

```ts
useEffect(() => {
  if (!visible || !focusMatch) return;   // early return when not focused — does NOT clear
  setHighlightedItemId(focusMatch.id);
  ...
}, [visible, focusMatch, historyItems]);
```

Since `ActivityDrawer` remains mounted (receives `visible` as a prop), state persists between open/close cycles. When the drawer is closed, `activityFocusTarget` is set to null → `focusMatch` becomes null → the effect is a no-op. But `highlightedItemId` still holds the old value, so the previously focused item remains highlighted the next time the drawer opens without a focus target.

Fix:

```ts
useEffect(() => {
  if (!visible) {
    setHighlightedItemId(null);
    return;
  }
  if (!focusMatch) return;
  setHighlightedItemId(focusMatch.id);
  ...
}, [visible, focusMatch, historyItems]);
```

---

### ℹ️ Notes (non-blocking)

- **`pendingActions` memoization** — the added `useMemo` is safe but unrelated to the PR feature. Fine to keep.
- **Doc impact check** flagged `docs/mobile/wireframes/chat-interface.md` — after reviewing, the doc doesn't describe the old filtering behavior explicitly, so no doc update is strictly needed. PR description's "No doc changes needed" is defensible.
- **No new tests** — the `focusMatch` computation and synthetic message injection are testable units. Not blocking for a Codex PR, but worth adding if these areas accumulate bugs.

---

### Verdict

**Changes needed** — the one item above before merge.

shantamg:
Addressed bot review feedback: ActivityDrawer now clears highlightedItemId whenever the drawer is not visible or there is no focus target, so stale highlights do not persist across normal drawer opens.

slam-paws:
## Bot review — LGTM ✅

**PR**: [codex] Show shared artifacts in chat
**Reviewer run**: pr-reviewer-20260510T213015Z-2849248

---

### Summary of changes

Three coherent pieces working together:

1. **Inline artifact injection** (`UnifiedSessionScreen.tsx` — `displayMessages` memo): Synthesizes chat messages for empathy/context artifacts that exist in `empathyStatusData` but have no corresponding backend chat message (i.e. the sender's view). Previously these were hidden behind indicator pills; now they render inline as proper chat bubbles.

2. **Scroll-to-highlight** (`ActivityDrawer.tsx`): New `focusTarget` prop drives a `focusMatch` lookup, then fires a `scrollToIndex` (with `onScrollToIndexFailed` fallback to `scrollToOffset`) after a 280ms delay to let the drawer animate open. Highlight clears on close.

3. **Direction-aware labels** (`ChatBubble.tsx`): Empathy statement header now reads "What you shared" vs "Empathy from [partner]" depending on `sharedContentDirection`. Shared context similarly distinguishes sent ("Context shared with X") from received ("New context from X").

---

### Checklist

| Item | Status |
|------|--------|
| Code does what the issue describes | ✅ |
| Logic errors / edge cases | ✅ — `onScrollToIndexFailed` fallback, NaN-safe timestamp parse, null `focusTarget` clearing on close |
| Convention compliance | ✅ — `useMemo`, `useCallback`, `useRef` used correctly; follows existing patterns |
| No unrelated changes | ✅ |
| No security issues | ✅ |
| CI checks passing | ✅ all green |
| Doc impact | ✅ — mobile UI only, no doc mapping applies |

---

- **Deduplication by exact timestamp string**: `message.timestamp === myAttempt.sharedAt` — works because both come from the same API response. No issue in practice, just noting it.
- **`myEmpathyStatements` superseded logic**: Correctly scoped to own messages only — partner's empathy statement won't be marked superseded if the user revises theirs. Improvement over the old behavior.
- **`onContextSharedPress` dep fix**: Adding it to the `renderItem` `useCallback` deps array is a real bug fix — the old code would silently reuse a stale closure.
- **No new tests**: The `displayMessages` synthesis logic is complex enough to warrant unit tests, but CI passed and mobile screen-level tests aren't the pattern here. Acceptable.

---

**LGTM** — ready for human merge review (targets `main`).

slam-paws:
Bot review complete — LGTM ✅. Ready for human review and merge. @shantamg

Issue #527, opened 2026-05-11
Review button unresponsive on What Matters Most needs card


slam-paws:
## Bug Report

In a partner session (Stage 1 — What Matters Most), the "Review" button on the needs-captured card at the bottom of the screen does not respond to taps. Darryl clicked it repeatedly with no effect.

**Steps to reproduce:**
1. Enter a partner session, reach Stage 1 (What Matters Most)
2. Chat until the AI captures a need and shows the bottom card: "1 need captured" with a "Review" button
3. Tap the Review button — nothing happens

**Expected:** Tapping Review should open the NeedsDrawer for reviewing/confirming captured needs.

**Screenshot:**

<img src="https://github.com/shantamg/meet-without-fear/blob/temp-image-upload/.github/images/issue-527-review-button-stuck.jpg?raw=true" alt="Review button stuck on What Matters Most" width="300" />

## What We Know So Far

- The Review button lives in `NeedsIdentifiedChatCard` (`mobile/src/screens/UnifiedSessionScreen.tsx:340`), rendered as a `TouchableOpacity` with `onPress={onReview}`
- The `onReview` handler calls `setShowNeedsDrawer(true)` — if tapping does nothing, it could be a touch interception issue (z-index, pointer events) or a state guard preventing the drawer from opening
- The compact card variant (rendered below the input) wraps in an `Animated.View` with `pointerEvents="auto"` — possible that a parent view is intercepting touches
- Needs full investigation into touch handling and drawer state

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Darryl Finkton Jr
- **Original message**: "Got stuck here. Clicked review button repeatedly and nothing happened"
- **Timestamp**: 1778458188.433669

slam-paws:
## Investigation Report

**Root cause**: The `ActivityDrawer` component has a `visible`/`isMounted` state desync that can leave a full-screen invisible touch-intercepting overlay on screen, making all UI beneath it (including the Review button) completely unresponsive.

**Evidence**:
- `ActivityDrawer.tsx:347` — initializes `isMounted = visible`, but has no mechanism to set `isMounted = false` when `visible` transitions to `false`
- `ActivityDrawer.tsx:401-408` — the `useEffect` only handles `visible = true`; the comment "The else branch is unnecessary since the component returns null when !visible" is incorrect — the component returns null when `!isMounted`, not `!visible`
[...]

PR #528, opened 2026-05-11
fix: close ActivityDrawer overlay when visible prop goes false

slam-paws:
## Changes
- Fix: ActivityDrawer's `useEffect` now handles `visible=false` by calling `closeDrawer()`, ensuring the full-screen overlay unmounts properly
- Add: `wasOpenedRef` guard to prevent `closeDrawer()` from firing on initial mount when `visible` starts as `false`

Related to #527

## Provenance
- **Requested by:** Darryl Finkton Jr
- **Original message:** "Got stuck here. Clicked review button repeatedly and nothing happened"

## Root Cause
When action callbacks inside `ActivityDrawer` (e.g. `onOpenRefinement`) called `setShowActivityMenu(false)` directly, the `useEffect` had no `else` branch to call `closeDrawer()`. This left a full-screen invisible overlay (`zIndex: 80`, `pointerEvents: auto`) mounted, silently consuming all touch events — including the Review button on the needs card.

## Docs updated
No doc changes needed — bug fix in a UI component with no corresponding living doc.


slam-paws:

Implementation matches requirements from #527:
- [x] `else` branch added to `ActivityDrawer`'s visibility `useEffect` to call `closeDrawer()` when `visible` goes `false`
- [x] Guard prevents `closeDrawer()` firing on initial mount when `visible` starts `false`
- [x] Single file changed (`ActivityDrawer.tsx`) — exact file identified in investigation

**Approach note**: The investigation suggested using `isMounted` (state) as the guard, but the PR uses `wasOpenedRef` (ref) instead. This is a better choice — `isMounted` in the dep array could cause effect re-runs since `closeDrawer()` sets `isMounted = false` via its animation callback. The ref avoids that loop risk while providing the same semantic guarantee.

slam-paws:
## Verification Report

**PR:** #528 — fix: close ActivityDrawer overlay when visible prop goes false
**Branch:** `feat/fix-activity-drawer-overlay-527`
**Affected package:** `mobile`

### Results

| Check | Result | Notes |
|-------|--------|-------|
| TypeScript (`tsc --noEmit`) | PASS | Zero errors |
| Tests (`jest`) | PASS | 166/166 tests pass |
| Pre-existing issues | N/A | 44 suites fail to import (`react-test-renderer` missing) — unrelated to this PR |

### Code Review

The fix is minimal and correct:
- Adds `wasOpenedRef` to prevent `closeDrawer()` from firing on initial mount
- Adds `else if (wasOpenedRef.current)` branch to properly call `closeDrawer()` when `visible` transitions to `false`
- Adds `closeDrawer` to the useEffect dependency array (required by the new branch)

No new test coverage was added — there is no existing ActivityDrawer test file to extend.

### Verdict: PASS

No type errors, no test regressions. Safe to merge.

---
*Automated verification by Slam Paws*

slam-paws:
## Bot review — PR #528

**LGTM ✅**

### Summary

The fix correctly addresses the root cause: when `visible` flips to `false` externally (e.g. via `setShowActivityMenu(false)` without going through the drawer's own close button/gesture), `closeDrawer()` was never called. Since `isMounted` only becomes `false` inside `closeDrawer()`'s animation callback, the invisible overlay stayed mounted with `zIndex: 80 / pointerEvents: auto` — swallowing every tap on the screen, including the Review button.

### Checklist

| Area | Result | Notes |
|------|--------|-------|
| **Correctness** | ✅ | `wasOpenedRef` guard correctly prevents `closeDrawer()` from firing on initial mount when `visible` starts `false`. Once opened (ref flips to `true`), any subsequent `visible=false` triggers the close animation and `setIsMounted(false)`. |
| **Completeness** | ✅ | Fully addresses issue #527. `closeDrawer` added to the `useEffect` dependency array as required by exhaustive-deps. |
| **Conventions** | ✅ | `useRef` + `useCallback` pattern consistent with surrounding code. No unnecessary refactoring. |
| **Security** | ✅ | UI-only change. |
| **Tests** | ⚠️ | No test added (none existed for this component). The fix is animation-state logic that's genuinely hard to unit-test in RN. Minor observation — not blocking. |
| **Doc impact** | ✅ | `doc-impact` CI check passed. No living doc covers this component. |
| **CI** | ✅ | All required checks pass (Detect changes, Checks & Tests, CI Result). |

### One minor observation

If `closeDrawer()` is already running via gesture at the same moment `visible` flips to `false`, both code paths will call `closeDrawer()` concurrently. `Animated.parallel` is restart-safe, so this is harmless in practice — but worth knowing if edge-case jitter appears later.

Routing to Stage 05 (merge-or-tag) on next tick.

slam-paws:

shantamg:
@slam-paws 

Look at the original issue. 

First of all, notice how that CTA at the bottom is dark? It should use the styles for light and dark mode. 

Also, I think the point is that there's no text input. That process involves chatting. I think you need to investigate why the state was wrong.

slam-paws:
Investigating — looking at the original issue, the CTA styling, and why the text input is missing.

slam-paws:
Good catch — I dug into both issues and pushed a fix (5bd4a1c2).

**1. Dark CTA styling**
The needs summary card styles used `t.colors.*` from the static `colors.ts` — which is hardcoded to dark-mode values and never switches. Swapped all needs card styles to use `palette.*` from `useAppAppearance()`, which properly adapts to light/dark mode.

**2. Missing text input**
Root cause: `computeShouldHideInput` in `chatUIState.ts` was hiding the input whenever `aboveInputPanel === 'needs-review'` or `'needs-share'` during Stage 3. But those two states don't actually render an above-input panel — they only render the compact needs card *below* the input via `renderBelowInput`. So the input was being hidden for no reason.

Fixed by removing `'needs-review'` and `'needs-share'` from the hide conditions. Only `'needs-reveal-validation'` (which has an actual `GuidedActionPanel` above the input) still hides it. Now users can keep chatting while the compact needs card shows below.

All 66 chatUIState tests + 166 mobile tests still pass. Types clean.

slam-paws:
No doc updates needed — this is a small bug fix (closing the ActivityDrawer overlay when `visible` goes false). The changes to `ActivityDrawer.tsx`, `UnifiedSessionScreen.tsx`, and `chatUIState.ts` don't affect any documented architecture, wireframes, or component structure.

The unmapped `mobile/src/utils` path is noted; we can add a mapping if that directory grows in scope.

shantamg (approved):
Reviewed for current production blocker sequence. CI is green and this addresses the needs drawer/composer/theme issue.

PR #540, opened 2026-05-11
fix: keep needs drawer open and readable

shantamg:
## Summary
- keep the Stage 3 needs drawer mounted during close animations and layered above its backdrop
- constrain the backdrop close target to the area above the visible sheet instead of the whole viewport
- move needs drawer actions out of the original Review button hit zone so repeated taps cannot accidentally trigger Chat to refine / confirm / validate
- make NeedCard use the active appearance palette so needs remain readable in light mode

## Local replay
- cloned production session cmoyye2jc001fmp1xv3ys1ntv into local dev DB
- opened Shantam side via E2E auth at localhost:8082
- reproduced on main/earlier patch with fixed-coordinate touch taps at the Review button location: every other tap closed the drawer
- inspected elementFromPoint after open and found the repeated tap was landing on needs-drawer-adjust (Chat to refine), not just the backdrop
- verified the updated branch keeps the drawer open across repeated taps at the original Review coordinate; the hit target is inert review content
- verified tapping the actual backdrop area above the sheet still closes the drawer
- verified needs text is readable in light mode

## Tests
- npm --workspace mobile test -- NeedsDrawer.test.tsx NeedCard.test.tsx --runInBand

### DIFF
--- a/mobile/src/components/NeedCard.tsx
+++ b/mobile/src/components/NeedCard.tsx
@@ -6,7 +6,7 @@
  */
 
 import { View, Text, TouchableOpacity, StyleSheet, ViewStyle } from 'react-native';
-import { colors } from '@/theme';
+import { useAppAppearance } from '@/theme';
 
 // ============================================================================
 // Types
@@ -34,6 +34,9 @@ export function NeedCard({
   style,
   testID,
 }: NeedCardProps) {
+  const { palette } = useAppAppearance();
+  const styles = makeStyles(palette);
+
   const cardContent = (
     <>
       <Text style={styles.category}>
@@ -67,12 +70,12 @@ export function NeedCard({
 // Styles
 // ============================================================================
 
-const styles = StyleSheet.create({
+const makeStyles = (palette: ReturnType<typeof useAppAppearance>['palette']) => StyleSheet.create({
   card: {
-    backgroundColor: 'rgba(147, 197, 253, 0.15)',
+    backgroundColor: palette.infoSoft,
     borderRadius: 12,
     borderWidth: 1,
-    borderColor: 'rgba(147, 197, 253, 0.3)',
+    borderColor: palette.info,
     padding: 16,
     marginBottom: 12,
     marginHorizontal: 16,
@@ -81,12 +84,12 @@ const styles = StyleSheet.create({
   category: {
     fontSize: 14,
     fontWeight: '600',
-    color: colors.textPrimary,
+    color: palette.textMuted,
     marginBottom: 4,
   },
   description: {
     fontSize: 16,
-    color: colors.textPrimary,
+    color: palette.text,
     lineHeight: 22,
   },
 });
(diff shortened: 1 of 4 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the NeedsDrawer so it stays mounted during close animations, constrains the backdrop tap target to the area above the visible sheet, moves action buttons out of the original Review-button hit zone, and switches NeedCard to the active appearance palette for light-mode readability.

## Feedback

- **Architecture (good)**: The `isMounted` / `wasOpened` / `isClosing` trio is a clean way to keep the component rendered through the close animation without re-triggering open on every render. The guard `if (isClosing.current) return` prevents double-close nicely.

- **Architecture (good)**: Constraining `backdropTouchableHeight` to `position` so taps on the sheet body don't accidentally hit the backdrop—this is exactly the right fix for the accidental-close-on-repeated-tap issue.

- **Performance (minor, non-blocking)**: `NeedCard` now calls `StyleSheet.create` on every render via `makeStyles(palette)`. This is already the pattern used in `NeedsDrawer` and the cost is negligible for a handful of cards, so consistent—just noting it.

- **UX (good)**: Moving `renderNeedsButtons()` and `renderRevealButtons()` above the `ScrollView` (between the header and the scroll) means the buttons are no longer in the bottom-sheet footer position where they overlapped with the Review button's coordinates. Smart fix.

- **Tests (good)**: The existing "validates the side-by-side needs reveal" test was correctly split with `unmount()` + re-render to avoid cross-mode element collisions after moving buttons. New tests for backdrop height, z-index, and backdrop close are solid.

- **Style**: `fixedButtonArea` now has both `borderTopWidth` and `borderBottomWidth` since buttons sit between header and scroll content—makes sense visually as a separator.

## Verdict
Approve — well-investigated fix with clear local replay notes, correct layering/animation lifecycle, and good test coverage.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the Stage 3 needs drawer to stay open during repeated taps, constrains the backdrop touch target to the visible area above the sheet, and switches NeedCard to theme-aware palette colors for light-mode readability.

## Feedback

- **Good: Double-close guard** — The `isClosing` ref in `NeedsDrawer.tsx` prevents race conditions from rapid taps. Clean pattern.

- **Good: Backdrop constraint** — Limiting the backdrop `Pressable` height to `backdropTouchableHeight` (the sheet's top offset) is a solid fix for the accidental-close problem. Much better than z-index hacks alone.

- **Good: Mount lifecycle** — Using `isMounted` state separate from `visible` prop allows close animations to complete before unmounting. The `wasOpened` ref prevents a spurious close on initial render when `visible` starts `false`.

- **Good: Test coverage** — New tests for backdrop height constraint, z-index layering, backdrop close behavior, and palette-driven NeedCard colors. Solid.

- **Minor: Brittle color assertions** (`NeedCard.test.tsx:37-38`) — The test hardcodes hex values (`#6c6961`, `#1a1815`). If the palette changes, this test breaks silently with a confusing diff. Consider importing the palette and comparing against `palette.textMuted` / `palette.text` instead. Non-blocking since it matches the existing test style in the codebase.

- **Minor: `StyleSheet.create` on every render** (`NeedCard.tsx:37`) — `makeStyles(palette)` runs in the component body, creating a new `StyleSheet` each render. This is consistent with how `NeedsDrawer` already works, so it's fine for consistency — just noting it as a potential memoization target if NeedCard ever renders in a large list.

## Verdict
**Approve** — Well-tested fix with clear local reproduction steps. The layering, hit-target, and animation lifecycle changes all make sense together.

PR #541, opened 2026-05-11
fix: refresh needs share state immediately

shantamg:
## Summary
- let the needs drawer keep animating during layout measurement instead of snapping to the measured open position
- immediately patch Stage 3 progress/session-state caches after confirming needs and sharing needs
- add regression coverage that confirm/share mutations update both progress and consolidated session-state caches

## Local replay
- used cloned production session cmoyye2jc001fmp1xv3ys1ntv in local E2E mode
- verified the drawer now slides from below the viewport instead of appearing at its final position immediately
- confirmed needs, reopened the card, shared needs, and verified the card immediately changed to Shared / Open review without waiting for app wake/refetch
- verified waiting copy appears immediately after sharing while the partner has not shared

## Tests
- npm --workspace mobile test -- NeedsDrawer.test.tsx useStages.test.ts --runInBand --forceExit
- npm --workspace mobile run check -- --pretty false

### DIFF
--- a/mobile/src/components/NeedsDrawer.tsx
+++ b/mobile/src/components/NeedsDrawer.tsx
@@ -191,10 +191,8 @@ export function NeedsDrawer({
     }
 
     const position = currentSnap.current === 'full' ? positionFullRef.current : position3Q;
-    setBackdropTouchableHeight(position);
-    drawerTranslate.setValue(position);
-    setContentHeight(drawerHostHeight - position);
-  }, [visible, drawerHostHeight, position3Q, drawerTranslate]);
+    snapTo(position, currentSnap.current === 'full' ? 0.6 : 0.4);
+  }, [visible, drawerHostHeight, position3Q, drawerTranslate, snapTo]);
 
   // -------------------------------------------------------------------------
   // Android back button
--- a/mobile/src/hooks/__tests__/useStages.test.ts
+++ b/mobile/src/hooks/__tests__/useStages.test.ts
@@ -41,6 +41,7 @@ import {
   useResolveSession,
   stageKeys,
 } from '../useStages';
+import { sessionKeys } from '../queryKeys';
 import {
   Stage,
   StageStatus,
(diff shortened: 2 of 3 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the NeedsDrawer resize flicker by reusing `snapTo()` instead of raw `setValue`, and adds optimistic gate patching to `useConfirmNeeds` / `useConsentShareNeeds` so the UI reflects confirmed/shared state immediately without waiting for cache invalidation round-trips.

## Feedback

- **Good pattern (useStages.ts)**: `patchStage3Gates` is a clean, DRY helper that patches both the progress and session-state caches in one shot. Nice use of the updater-function form of `setQueryData`.

- **Minor — timestamp source (useStages.ts:1516)**: `needsConfirmedAt: new Date().toISOString()` generates a client-side timestamp. The server already returns `confirmedAt` in `ConfirmNeedsResponse` — you could use `data.confirmedAt` instead (like `useConsentShareNeeds` does with `data.sharedAt`) for consistency and accuracy. Non-blocking.

- **Architecture note (useStages.ts:1513–1520, 1577–1583)**: The project's state-management rules in CLAUDE.md prescribe optimistic updates via `onMutate` (with rollback in `onError`), but these patches are in `onSuccess`. Since the intent here is "fast gate flag update" rather than full optimistic UX (the `invalidateQueries` calls still follow immediately), this is reasonable — the gate patch lands the moment the server confirms, avoiding the flicker window. Just noting the deviation for awareness; no change needed.

- **NeedsDrawer fix (NeedsDrawer.tsx:194)**: Replacing the manual `setValue` + `setContentHeight` + `setBackdropTouchableHeight` trio with `snapTo()` is a good dedup. `snapTo` handles all three state updates plus animates, so this also gives users a smooth transition instead of a jump on resize.

- **Tests (useStages.test.ts)**: Good coverage — both new tests seed the progress and session-state caches, run the mutation, and assert the gate flags are patched. The `createTestQueryClient` / `createWrapper(queryClient)` refactor is a nice improvement for cache-asserting tests. Mock response shapes updated to match the current `ConsentShareNeedsResponse` DTO — good cleanup.

## Verdict
**Approve** — Clean, focused changes with good test coverage. The only suggestion (use `data.confirmedAt` over client-side timestamp) is minor and non-blocking.

PR #549, opened 2026-05-11
Fix empathy statement duplicate superseded label

shantamg:
## Summary
- dedupe synthesized empathy statement chat items when the real chat message timestamp differs slightly from EmpathyAttempt.sharedAt
- only mark empathy statements as superseded when a newer distinct statement exists
- add pure utility coverage for timestamp-tolerant matching and superseded detection

## Evidence
Live test session `cmp1jjdlo0008px3vl7sgwnwx` showed the backend writes EmpathyAttempt.sharedAt and Message.timestamp a few ms apart:
- Catherine: sharedAt `2026-05-11T18:40:20.812Z`, message timestamp `2026-05-11T18:40:20.816Z` (4ms)
- James: sharedAt `2026-05-11T18:47:08.078Z`, message timestamp `2026-05-11T18:47:08.082Z` (4ms)

The previous exact timestamp check failed to recognize those as the same first-share message, causing a synthetic duplicate that could render as “Updated version below.”

## Tests
- `npm --workspace mobile test -- --runTestsByPath src/utils/__tests__/empathyMessageMatching.test.ts --runInBand`

### DIFF
--- a/mobile/src/screens/UnifiedSessionScreen.tsx
+++ b/mobile/src/screens/UnifiedSessionScreen.tsx
@@ -110,6 +110,7 @@ import {
   trackShareDraftSent,
 } from '../services/analytics';
 import { shouldShowSessionEntryMoodCheck } from '../utils/sessionEntryMoodCheck';
+import { hasNewerDistinctEmpathyStatement, isSameEmpathyAttemptMessage } from '../utils/empathyMessageMatching';
 
 // ============================================================================
 // Types
@@ -2078,10 +2079,7 @@ export function UnifiedSessionScreen({
 
     if (myAttempt?.content && myAttempt.sharedAt) {
       const hasMyAttemptMessage = baseMessages.some(
-        (message) =>
-          message.role === MessageRole.EMPATHY_STATEMENT &&
-          message.senderId === user?.id &&
-          message.timestamp === myAttempt.sharedAt,
+        (message) => isSameEmpathyAttemptMessage(message, user?.id, myAttempt),
       );
 
       if (!hasMyAttemptMessage) {
@@ -2103,9 +2101,10 @@ export function UnifiedSessionScreen({
       const partnerTimestamp = partnerAttempt.revealedAt || partnerAttempt.sharedAt;
       const hasPartnerAttemptMessage = baseMessages.some(
         (message) =>
-          message.role === MessageRole.EMPATHY_STATEMENT &&
-          message.senderId === partnerAttempt.sourceUserId &&
-          message.timestamp === partnerTimestamp,
+          isSameEmpathyAttemptMessage(message, partnerAttempt.sourceUserId, {
+            content: partnerAttempt.content,
+            sharedAt: partnerTimestamp,
+          }),
       );
 
       if (!hasPartnerAttemptMessage) {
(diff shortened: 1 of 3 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes empathy statement duplicate rendering caused by a few-ms timestamp drift between `EmpathyAttempt.sharedAt` and `Message.timestamp`. Extracts matching logic into pure utility functions with tests.

## Feedback

- **Good pattern**: Extracting `isSameEmpathyAttemptMessage` and `hasNewerDistinctEmpathyStatement` into pure utilities with dedicated tests is the right call — keeps `UnifiedSessionScreen` cleaner and makes the matching logic independently testable.

- **Tolerance value (nice-to-have)**: The 10s tolerance constant is very generous relative to the observed 4ms drift. Not a blocking issue since content equality is also checked (false positives are unlikely), but worth noting — if content ever gets updated server-side between write and read, 10s could theoretically match the wrong message. A tighter window like 1s would still cover the drift with plenty of margin.

- **Potential follow-up**: The `mySharedContext` matching at `UnifiedSessionScreen.tsx:2131-2135` still uses exact timestamp comparison (`message.timestamp === mySharedContext.sharedAt`). If the same write-path timing drift applies to shared context messages, this could hit the same duplicate bug. Worth a follow-up if you see it in practice.

- **Test coverage**: Good coverage of the core cases — timestamp tolerance, content mismatch, duplicate vs. distinct revision detection. The tests are clear and focused.

## Verdict
Approve — clean, well-scoped fix with good evidence and test coverage. The tolerance constant and shared-context matching are minor items that don't block merge.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes duplicate empathy statement rendering caused by a few-ms timestamp mismatch between `EmpathyAttempt.sharedAt` and `Message.timestamp`. Extracts matching logic into a tested utility with a 10s tolerance window, and adds a content-equality check before marking older empathy statements as superseded.

## Feedback

- **Good pattern**: Extracting the matching logic into `empathyMessageMatching.ts` with pure functions and dedicated tests is clean and testable. The tolerance approach is pragmatic.

- **Content match in `isSameEmpathyAttemptMessage` is a nice safeguard** — comparing both content and timestamp-within-tolerance prevents false positives from unrelated messages that happen to be close in time.

- **Minor / non-blocking — shared context has the same timestamp risk** (`UnifiedSessionScreen.tsx:2131-2136`): The `SHARED_CONTEXT` dedup check at line 2131 still uses exact `message.timestamp === mySharedContext.sharedAt`. If the backend writes shared-context messages with the same few-ms offset pattern, this will eventually hit the same duplicate bug. Worth noting for a follow-up, though it may not be observed yet.

- **Tolerance value (10s)**: The observed drift is ~4ms and the tolerance is 10,000ms. That's a generous safety margin, which is fine given the content equality check acts as a second guard. Just flagging that the tolerance is 2500x the observed drift — if you ever have rapidly revised empathy statements (within 10s of each other with identical content), the tolerance could theoretically cause a false match, but the content check makes this scenario vanishingly unlikely.

- **Test coverage**: Tests cover the key cases well — timestamp drift match, distinct revision non-match, duplicate-content non-supersede, and distinct-content supersede. Good.

## Verdict
**Approve** — The fix is well-scoped, addresses a real bug with evidence from a live session, and the utility extraction with tests is the right approach. The shared-context timestamp check is a low-priority follow-up worth tracking.

PR #569, opened 2026-05-13
Stage 4 Phase 6: AI prompt tuning across seven surfaces

shantamg:
## Summary

Implements every prompt change described in `.planning/STAGE_4_PHASE_6_AI_PROMPT_TUNING.md` across the seven Stage 4 AI surfaces, bundled into one PR per the plan's `/goal`-shaped approach.

- New `backend/src/services/stage4-prompts.ts` collects per-surface templates side-by-side so gold-tone consistency is visually obvious.
- Sub-chat personas (Surfaces 2/3/4) replaced with rich, anchor-aware templates that ground every suggestion in the user's exact need/proposal phrasing.
- Stage 4 main-chat prompt picks up two new conditional clauses:
  - **Surface 1 (COVERAGE_REVIEW open-needs):** lists not-declined, not-willing-covered needs by exact phrasing; instructs the AI to raise one at a time during natural pauses.
  - **Surface 6 (RESOLVED listen-first):** activates when `session.status === 'RESOLVED'` and the user hasn't yet matched the explicit-ask regex; instructs reflect-only behavior.
- **Surface 5 (handoff bridge):** share-selections endpoint persists a single templated AI message when this user shares first and the partner hasn't shared yet.
- **Surface 7 (Tending micro-personas):** three orientation fragments exported from the new prompts file, ready to wire into per-orientation conversations.

Behavior tests for the sub-chat were already structural (they assert main-chat history inclusion, not persona wording), so no test updates were needed.

## Test plan
- [x] `npm run check` across all workspaces
- [x] `npm run test` across all workspaces (backend, mobile, shared)
- [ ] Manual verification per `STAGE_4_MANUAL_TESTING_CHECKLIST.md` Phase 6 section (post-merge eyes-and-taste loop per the plan)

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -128,7 +128,7 @@ async function getStage4OpenNeedsForPrompt(
     const [coverageRows, willingSelections, declinations] = await Promise.all([
       prisma.stage4NeedCoverage.findMany({
         where: { sessionId, coverageStatus: { in: ['OPEN', 'PARTIAL'] } },
-        select: { id: true, needId: true, coveringProposalIds: true },
+        select: { id: true, needId: true, needLabel: true, coveringProposalIds: true },
       }),
       prisma.stage4ProposalSelection.findMany({
         where: { sessionId, userId, decision: 'WILLING' },
@@ -150,15 +150,19 @@ async function getStage4OpenNeedsForPrompt(
     });
     if (candidateRows.length === 0) return null;
     const needIds = candidateRows.map((r) => r.needId).filter((n): n is string => Boolean(n));
-    if (needIds.length === 0) return null;
-    const needs = await prisma.identifiedNeed.findMany({
-      where: { id: { in: needIds }, vessel: { userId } },
-      select: { id: true, need: true },
-    });
+    const needs = needIds.length > 0
+      ? await prisma.identifiedNeed.findMany({
+          where: { id: { in: needIds }, vessel: { userId } },
+          select: { id: true, need: true },
+        })
+      : [];
     const byId = new Map(needs.map((n) => [n.id, n.need] as const));
     const labels: Array<{ needLabel: string }> = [];
     for (const row of candidateRows) {
-      const label = row.needId ? byId.get(row.needId) : undefined;
+      // Prefer the user's exact phrasing from IdentifiedNeed; fall back to the
+      // coverage row's needLabel so coverage rows without an IdentifiedNeed
+      // link (or whose link belongs to the partner's vessel) still surface.
+      const label = (row.needId && byId.get(row.needId)) || row.needLabel;
       if (label) labels.push({ needLabel: label });
     }
     return labels.length > 0 ? labels : null;
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -44,6 +44,7 @@ import { captureProposedNeedsForUser } from '../services/needs';
 import { cleanVisibleAIText } from '../utils/visible-text';
 import { captureStage4Turn } from '../services/stage4-capture.service';
 import { applyStage4AutoClosureFromSignal } from '../services/stage4-auto-closure.service';
+import { isExplicitAskForInput } from '../services/stage4-prompts';
 
 // ============================================================================
 // Helpers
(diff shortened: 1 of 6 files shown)


slam-paws:
On it — reviewing now.

backend/src/controllers/stage4.ts
```diff
@@ -666,6 +667,68 @@ export async function shareStage4Selections(req: Request, res: Response): Promis
       });
     }
 
+    // Stage 4 Phase 6 (Surface 5) — quiet handoff bridge. When this user shares
+    // for the first time AND the partner hasn't shared yet, append a single
+    // templated AI message to their chat so the moment doesn't go silent.
+    try {
+      const partnerHasShared = partnerId
+        ? Boolean(
+            await prisma.stageProgress.findFirst({
```
slam-paws: **Inconsistent JSON query pattern.** The rest of this file checks `gatesSatisfied` by fetching and inspecting in TypeScript (see `hasSubmittedSelectionGate` at line 261). This Prisma JSON path filter works on PostgreSQL but diverges from the established pattern — consider using the same fetch-then-check approach for consistency:

```ts
const partnerProgress = partnerId
  ? await prisma.stageProgress.findFirst({
      where: { sessionId, userId: partnerId, stage: 4 },
      select: { gatesSatisfied: true },
    })
  : null;
const partnerHasShared = hasSubmittedSelectionGate(partnerProgress?.gatesSatisfied);
```

This also reuses the existing helper.

backend/src/controllers/messages.ts
```diff
@@ -114,6 +115,78 @@ async function getStage4InventoryPromptContext(sessionId: string, currentUserId:
...
+      return !covered;
+    });
+    if (candidateRows.length === 0) return null;
+    const needIds = candidateRows.map((r) => r.needId).filter((n): n is string => Boolean(n));
+    if (needIds.length === 0) return null;
+    const needs = await prisma.identifiedNeed.findMany({
+      where: { id: { in: needIds }, vessel: { userId } },
+      select: { id: true, need: true },
+    });
+    const byId = new Map(needs.map((n) => [n.id, n.need] as const));
```
slam-paws, line 159: **Silent drop of coverage rows with null `needId`.** Rows where `needId` is null pass the declination/willing filter (using `row.id` as fallback on line 149) but are then filtered out at line 158 (`r.needId` is null → excluded). This means a coverage row without a linked `identifiedNeed` will never surface as an open need, even if it is genuinely open.

If that is intentional (coverage rows without a `needId` have no user-phrased label to show), a comment would help. If not, the `needLabel` field on `Stage4NeedCoverage` itself could serve as a fallback label.

backend/src/services/stage4-prompts.ts
```diff
@@ -0,0 +1,183 @@
...
+  const t = text.toLowerCase();
+  const patterns = [
+    /\bwhat\s+should\s+i\s+do\b/,
+    /\bwhat\s+do\s+you\s+think\s+i\s+should\b/,
+    /\bhelp\s+me\s+think\b/,
+    /\bcan\s+you\s+weigh\s+in\b/,
+    /\bweigh\s+in\b/,
+    /\bany\s+(?:advice|ideas|suggestions)\b/,
+    /\bwhat\s+would\s+you\s+(?:suggest|recommend|do)\b/,
+    /\bgive\s+me\s+(?:advice|ideas|suggestions|your\s+take)\b/,
```
slam-paws, line 130: **Nice — `isExplicitAskForInput` is well-scoped.** Consider adding a few unit tests for this since it controls a behavioral gate (listen-first → advice mode). The regex list is a good starting point but could benefit from coverage of edge cases (e.g., "what do you think?" without "I should").

slam-paws (dismissed):
## Summary

Centralizes Stage 4 AI prompt fragments into a new `stage4-prompts.ts` file covering all seven surfaces, and wires three new runtime behaviors into the main chat path: open-needs surfacing (Surface 1), listen-first mode for resolved sessions (Surface 6), and a handoff bridge message when the first partner shares (Surface 5). Sub-chat personas are upgraded from static strings to anchor-aware templates grounded in the user's own phrasing.

## Feedback

- **[Consistency]** `stage4.ts:676` — The Prisma JSON path filter (`{ path: ['selectionSubmitted'], equals: true }`) works but diverges from the established pattern in this file. The existing `hasSubmittedSelectionGate` helper does the same check in TypeScript after a plain fetch. Suggest reusing it for consistency. *(inline comment)*

- **[Correctness — minor]** `messages.ts:149–163` — Coverage rows with null `needId` pass the declination filter (using `row.id` as fallback) but are silently excluded when building the label list. If intentional, a brief comment would clarify. If not, `Stage4NeedCoverage.needLabel` could serve as a fallback. *(inline comment)*

- **[Testing — nice-to-have]** `isExplicitAskForInput` controls the listen-first → advice-mode gate. A small unit test suite would lock down the regex behavior and catch regressions if patterns are added later. *(inline comment)*

- **[Architecture — positive]** The new `stage4-prompts.ts` file is a strong organizational choice. Having all seven surface templates side-by-side makes gold-tone consistency reviewable at a glance. The anchor-aware persona functions (`needsBrainstormPersona`, `proposalRefinementPersona`) handle null/undefined gracefully with sensible fallbacks.

- **[Prompt quality — positive]** The prompt language is excellent — grounded, non-directive, and aligned with the product philosophy. The listen-first clause and the "full closure is a legitimate, dignified outcome" framing in the tending personas are particularly well-crafted.

- **[Safety]** The handoff bridge message uses an `alreadyBridged` exact-content check to prevent duplicates, which is pragmatic. The `try/catch` around the bridge logic ensures it can't break the share-selections response.

## Verdict

**Approve.** Well-structured PR with thoughtful prompt design. The two inline suggestions (JSON query consistency, null-needId clarification) are non-blocking improvements to consider.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Centralizes Stage 4 AI prompt fragments into a single `stage4-prompts.ts` file across seven surfaces: anchor-aware sub-chat personas (Surfaces 2/3/4), main-chat open-needs clause (Surface 1), listen-first reflection mode for resolved sessions (Surface 6), a handoff bridge message when sharing first (Surface 5), and tending micro-personas (Surface 7). Clean, well-structured prompt engineering.

## Feedback

- **Good: Centralized prompt file** — Collecting all seven surfaces side-by-side in `stage4-prompts.ts` makes gold-tone consistency reviewable at a glance. The anchor-aware persona functions (`needsBrainstormPersona`, `proposalRefinementPersona`) that ground suggestions in the user's exact phrasing are a meaningful upgrade over the old static strings.

- **Good: Idempotency** — The `alreadyBridged` check in the handoff bridge (stage4.ts:~700) prevents duplicate bridge messages on retries.

- **Good: Graceful degradation** — Both `getStage4OpenNeedsForPrompt` and the bridge message block catch errors and log warnings without failing the request.

- **Minor: `isStage4ListenFirstMode` called unconditionally** — Unlike `stage4OpenNeeds` and `stage4InventoryContext` which are gated on `currentStage === 4`, the `isStage4ListenFirstMode` call runs for all stages. It short-circuits immediately when `sessionStatus !== 'RESOLVED'` so it's not a bug, but gating it on `currentStage === 4` would be more consistent and skip the function call overhead for stages 0–3. Non-blocking.

- **Minor: Regex coverage in `isExplicitAskForInput`** — The current patterns are solid for a v1. Worth noting that common phrasings like "thoughts?" or "your opinion?" won't trigger it. The plan mentions upgrading only if it misses cases in practice, which is a reasonable approach.

- **Note: Bridge message uses hardcoded text matching** — The `alreadyBridged` check matches on exact `content === bridgeContent`. If the template text is ever tweaked, old bridge messages won't match and a second could be inserted. Fine for now since the template is stable, but worth keeping in mind.

## Verdict

**Approve.** Well-structured prompt centralization with proper error handling and idempotency. The prompt content reads well and stays true to the gold-tone principles. The two minor points above are non-blocking.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Adds a new `stage4-prompts.ts` module centralizing prompt fragments for seven Stage 4 AI surfaces, replacing inline persona strings with richer, anchor-aware templates. Also adds open-needs coverage review, listen-first mode for resolved sessions, and a handoff bridge message.

## Feedback

- **Good pattern — centralized prompts**: Pulling all seven surface templates into one file (`stage4-prompts.ts`) makes tone consistency reviewable at a glance. Nice.

- **Good pattern — anchor grounding**: Sub-chat personas now quote the user's exact need/proposal phrasing instead of generic instructions. This is a meaningful upgrade.

- **Minor — `isStage4ListenFirstMode` called unconditionally**: In `messages.ts`, the function is called for every stage, not just Stage 4. It short-circuits on `sessionStatus !== 'RESOLVED'` so it's safe, but it does scan history for every message in a RESOLVED session regardless of stage. Consider gating with `currentStage === 4` for consistency with the `stage4OpenNeeds` call above it, and to avoid the history scan for non-Stage-4 resolved sessions.

- **Minor — bridge message partner name fallback**: The nickname lookup at `stage4.ts:697-701` correctly uses `myMember?.nickname` (what I call my partner). The fallback chain to `partnerMember?.user.firstName` then `partnerMember?.user.name` is solid.

- **Tests**: Good coverage on `isExplicitAskForInput` with both positive and negative cases. Persona tests verify grounding in user phrasing. The test file is well-structured.

- **Bridge message dedup**: Using exact content match for idempotency is fine since the content is deterministic per partner name. If the partner name changes between calls (unlikely but possible via nickname update), a second bridge message would be created — acceptable edge case.

## Verdict

**Approve** — Clean, well-organized prompt tuning with good test coverage. The one minor suggestion (gating `isStage4ListenFirstMode` behind `currentStage === 4`) is non-blocking.

Issue #586, opened 2026-05-14
Shared context card is left-aligned instead of right for sender


slam-paws:
## Bug Report

When a user shares context with their partner (Stage 4 "Walking in Their Shoes"), the "CONTEXT SHARED WITH [partner]" card appears left-aligned in the chat. Since the current user is the one who shared it, this card should be right-aligned (sender side), not left-aligned (receiver side).

Screenshot shows Shantam's view after sharing context with Darryl — the shared context card sits on the left as if it were received, but it should be on the right since Shantam initiated the share.

**Likely area:** `mobile/src/components/chat/renderers/SharedContextRenderer.tsx` and `mobile/src/components/ChatBubble.tsx` — the alignment logic probably doesn't account for `isMine` / sender status on shared context cards.

## What We Know So Far

• The shared context card renders via `SharedContextRenderer.tsx` and is displayed in the chat timeline
• The card likely lacks an `isMine` or alignment prop that would place it on the right for the sender
• The "Delivered" label is present at the bottom-right of the card, suggesting the card does know it was sent — but alignment is wrong

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Shantam
- **Original message**: "If I shared it, it should be on the right side, not the left."
- **Timestamp**: 1778736358.274419

## Screenshot

<img src="https://github.com/shantamg/meet-without-fear/blob/temp-image-upload/.github/images/issue-586-shared-context-alignment.png?raw=true" alt="Shared context card left-aligned instead of right" width="300" />

slam-paws:
## Investigation Report

**Root cause**: `ChatBubble.tsx:getContainerStyle()` (line 217) always applies `sharedContextContainer` with `alignItems: 'flex-start'` (left-aligned) for all SHARED_CONTEXT messages, regardless of whether the current user sent or received the context. The `sharedContentDirection` field (`'sent'` | `'received'`) is already computed and passed to ChatBubble (via `UnifiedSessionScreen` lines 2372-2376), but the container alignment logic never checks it.

**Evidence**:
- `mobile/src/components/ChatBubble.tsx:517-519` — `sharedContextContainer` hardcodes `alignItems: 'flex-start'`
- `mobile/src/components/ChatBubble.tsx:217-224` — `getContainerStyle()` returns the same style for all SHARED_CONTEXT messages without checking direction
[...]

PR #587, opened 2026-05-14
fix: right-align shared context card for sender

slam-paws:
## Changes
- Fix: shared context card now right-aligns (`flex-end`) when the current user is the sender, matching user message alignment
- The `sharedContentDirection` field was already computed and passed correctly — only the container style selection in `getContainerStyle()` was ignoring it

Related to #586

## Provenance
- **Requested by:** Shantam
- **Original message:** "If I shared it, it should be on the right side, not the left."

## Docs updated
No doc changes needed — one-line alignment fix in ChatBubble styles

### DIFF
--- a/mobile/src/components/ChatBubble.tsx
+++ b/mobile/src/components/ChatBubble.tsx
@@ -218,7 +218,11 @@ export function ChatBubble({
     if (isUser) return styles.userContainer;
     if (isSystem) return styles.systemContainer;
     if (isEmpathyStatement) return styles.empathyStatementContainer;
-    if (isSharedContext) return styles.sharedContextContainer;
+    if (isSharedContext) {
+      return message.sharedContentDirection === 'sent'
+        ? styles.sharedContextSentContainer
+        : styles.sharedContextContainer;
+    }
     if (isShareSuggestion) return styles.shareSuggestionContainer;
     return styles.aiContainer;
   };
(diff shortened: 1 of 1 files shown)


slam-paws:

Implementation matches requirements from #586:
- [x] Shared context card right-aligns (`flex-end`) when the current user is the sender — implemented in `ChatBubble.tsx:getContainerStyle()`
- [x] Uses existing `sharedContentDirection` field (already computed in `UnifiedSessionScreen`) — no plumbing changes needed
- [x] New `sharedContextSentContainer` style keeps shared-context-specific spacing (`marginVertical`) while changing alignment — cleaner than reusing `userContainer`

No scope deviations detected. Minimal, focused one-file change. Ready for verification.

slam-paws:

Implementation matches requirements from #586:
- [x] Right-align shared context card when sender is current user — `getContainerStyle()` now checks `sharedContentDirection === 'sent'` and applies `sharedContextSentContainer` with `alignItems: 'flex-end'`
- [x] Root cause addressed — the exact function identified in the investigation is the one modified
- [x] Minimal scope — single file, 9 additions / 1 deletion, no unrelated changes

Note: investigation also flagged `SharedContextRenderer.tsx` (newer ChatTimeline path) as having the same issue. This PR fixes the active ChatBubble path only. The secondary path can be addressed separately if needed.

PR #638, opened 2026-05-20
Stages 3 & 4 rework + shared content card redesign

shantamg:
## Summary

Two related streams of work landing together on this branch.

### Stages 3 & 4 rework
- Stage 3 need-editing drawer + preview API
- Stage 4 focused walkthrough state and need suggestions
- Verification flow fixes for both stages
- New \`partner-empathy-validation\` above-input panel — owns Stage 2 focus when the partner's empathy attempt awaits the user's response (sits ahead of share-suggestion in panel priority)

### Shared content card redesign
- Shared moments (context + empathy passed between partners) now render as framed envelopes — hairline rule above and below, tinted body (warm \`accentSoft\` for received, cool \`infoSoft\` for sent), in-card mono uppercase label, directional arrow (↑ sent, ↓ received).
- Centered ~92% width so they sit outside the user/AI side rails — reads as a cross-person artifact, not a private AI-chat bubble.
- Standalone \`context-shared\` / \`empathy-shared\` indicator lines are suppressed when an inline bubble exists; the frame's in-card label is now the chapter break. Self-authored \`SHARED_CONTEXT\` (no inline bubble) still renders the line-only indicator.
- Frames are tappable to open the activity drawer (replaces the old indicator tap target).

## Test plan
- [ ] Stage 3: need editing drawer opens, edits preview correctly, save commits
- [ ] Stage 4: focused walkthrough advances through suggestions
- [ ] Stage 3 & 4 verification flows complete without regressions
- [ ] Partner empathy validation panel appears in Stage 2 when partner empathy is pending; share-suggestion panel waits behind it
- [ ] Shared context card (received) shows warm tint, ↓ arrow, "Context from {Partner}" label
- [ ] Shared context card (sent) shows cool tint, ↑ arrow, "Context shared with {Partner}" label
- [ ] Empathy frame (both directions) uses serif italic body, same envelope frame
- [ ] Tapping a shared frame opens the activity drawer
- [ ] Self-authored \`SHARED_CONTEXT\` still shows a line-only indicator (no bubble)
- [ ] Felt-heard and other contentless indicators still render as line dividers
- [ ] \`npm run check\` and \`npm run test\` pass in mobile workspace

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #640, opened 2026-05-22
Stage 3/4 rework and full Tending flow

shantamg:
## Summary

This PR combines the Stage 3/4 rework and the full Tending follow-up flow into one PR against `main`.

### Stage 3/4 rework
- Reworks Stage 3 need editing, refinement, sharing, and Stage 4 transition behavior.
- Moves Stage 4 into focused need-by-need strategy walkthrough state.
- Uses tool-call/state capture paths for Stage 4 instead of leaking untagged reasoning into visible chat.
- Keeps Stage 4 proposals, selections, ranking, and closure aligned with the reviewed needs flow.
- Preserves the shared content card redesign and partner-empathy validation panel behavior from the previous Stage 3/4 PR.

### Full Tending flow
- Adds structured Tending entries, check-ins, outcomes, adjustments, reminders, coordination cycles, private individual entries, and passive re-entry.
- Adds private between-period notes, including optional carry-forward and explicit share consent for check-ins.
- Wires resolved-session Tending conversation prompts into chat with selected private note handling.
- Adds Tending history read model and mobile history/record surfaces.
- Adds backend, mobile, E2E, and moment-eval coverage for the full flow.

### Merge notes
- Retargeted from the stacked base to `main` and merged latest `origin/main` into `codex/full-tending-flow`.
- Resolved the Stage 3 needs-drawer conflict by keeping the newer inline/full-screen needs refinement flow from this branch.
- Fixed a merge regression where mutual needs sharing could emit an obsolete Stage 3 guidance message after auto-advancing to Stage 4.
- Supersedes PR #638.

## Verification

- `DATABASE_URL=postgresql://user:pass@localhost:5432/mwf npx prisma@6.12.0 validate --schema backend/prisma/schema.prisma`
- `npm run check --workspace shared`
- `npm run check --workspace backend`
- `npm run check --workspace mobile`
- `npm run check --workspace e2e`
- `npm test --workspace backend -- --runTestsByPath src/routes/__tests__/stage3.test.ts src/routes/__tests__/invitations.test.ts src/services/__tests__/tending.service.test.ts src/routes/__tests__/stage4.test.ts --runInBand` — 115 passed
- `npm test --workspace backend -- --runTestsByPath src/services/__tests__/stage-prompts.test.ts src/services/__tests__/stage4-prompts.test.ts src/routes/__tests__/messages.test.ts --runInBand` — 159 passed
- `npm test --workspace mobile -- --runTestsByPath src/components/__tests__/TendingPanel.test.tsx src/screens/__tests__/TendingCheckinScreen.test.tsx src/hooks/__tests__/useStages.test.ts src/utils/__tests__/chatUIState.test.ts --runInBand --forceExit` — 130 passed
- `python3 scripts/test_mwf_moment_eval.py` — 76 passed; existing Python datetime deprecation warnings emitted
- `python3 scripts/mwf_moment_eval.py run --moment stage-4-tending-structured-checkin --max-iterations 1 --mock-response "What happened first matters. If the commitment did not happen, that is information for adjusting or reopening strategy work, not proof that either person failed."`
- `npm --workspace e2e run e2e -- --project=two-browser-stage-4-redesign` — 7 passed


slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Massive PR combining Stage 3/4 rework with the full Tending flow — 16,710 additions across 100+ files. Impressive scope with solid test coverage (115 + 159 + 130 + 76 tests passing). There are a few issues that should be addressed before merge, ranging from a security concern to cache-management races.

## Blocking Issues

### 🔴 IDOR: `confirmNeeds` adjustment updates not scoped to user's vessel
**`backend/src/controllers/stage3.ts:406`**

The `adjustments` loop calls `prisma.identifiedNeed.update({ where: { id: adj.needId } })` without constraining to `vesselId: userVessel.id`. An attacker supplying arbitrary need IDs in the `adjustments` array can overwrite another user's need text. The downstream `updateMany` at line 418 correctly scopes to `vesselId`, but the adjustment loop does not.

**Fix:** Add `vesselId: userVessel.id` to the `update` where-clause, or validate that every `adj.needId` is present in `userNeeds` before the loop.

---

### 🔴 `TendingCheckinScreen` — private entries visible to wrong user
**`mobile/src/screens/TendingCheckinScreen.tsx:88–96`**

`isRespondable(entry, currentUserId?)` accepts `currentUserId` as an optional arg, but the `useMemo` call passes no second argument. The ownership guard (`entry.scope === INDIVIDUAL && currentUserId && entry.ownerUserId !== currentUserId`) always short-circuits because `currentUserId` is `undefined` — meaning **all INDIVIDUAL entries are shown as respondable regardless of ownership**.

**Fix:** Pass `currentUserId` from the auth context into `isRespondable`.

---

### 🔴 Cache-first rule violation: `setQueryData` + immediate `refetchQueries` race
**`mobile/src/hooks/useStages.ts:2037–2045`**

`refreshStage4Caches()` calls `refetchQueries({ queryKey: stageKeys.stage4(sessionId) })` immediately after callers already wrote canonical state via `setQueryData`. This is exactly the race condition prohibited by CLAUDE.md Rule 6 ("Never `invalidateQueries` on a key with optimistic updates in-flight"). The refetch can overwrite the server-confirmed response with a stale inflight result.

Same pattern in `useSubmitTendingResponse` at lines 2396–2405.

**Fix:** Remove the `stage4` and `tending` keys from `refreshStage4Caches` since callers already set canonical data on those keys. Only refetch ancillary keys (`strategies`, `agreements`, `progress`, `session.state`).

## Important Concerns (non-blocking but worth addressing)

### 🟡 Race condition: concurrent `submitTendingCheckin` calls
**`backend/src/services/tending.service.ts:497–659`**

Two users can call `submitTendingCheckin` concurrently. Both independently fetch `openEntries`, then enter separate transactions. With `ANOTHER_ROUND`, both will attempt to `deleteMany` + `session.update`. No `SELECT FOR UPDATE` or idempotency guard exists.

### 🟡 `TendingEntry → Agreement` CASCADE deletes audit history
**Schema: `TendingEntry.agreementId` FK**

Deleting an `Agreement` silently cascades through all `TendingEntry` → `TendingResponse` → `TendingResponsePartialClosure` rows. Tending check-in data is historical/audit data — consider `SET NULL` instead of `CASCADE` to preserve history if an agreement is removed.

### 🟡 Missing index on `TendingResponse.userId`
The `@@unique([tendingEntryId, userId])` covers composite lookups, but queries by `userId` alone (e.g., "show this user's check-in history") will table-scan. Add `@@index([userId])`.

### 🟡 `addCustomNeed` doesn't validate `category` against enum
**`backend/src/controllers/stage3.ts:958`**

The `category` value is cast with `as NeedCategory` without runtime enum validation. Compare with `captureNeeds` which correctly uses `isNeedCategory()`.

### 🟡 `requestSuggestions` inputs unvalidated
**`backend/src/controllers/stage4.ts:1998`**

`count` is unbounded and `focusNeeds` has no length/type validation. Even though this is currently a stub, the unvalidated pattern is risky once the AI call is wired up.

### 🟡 `require()` anti-pattern in `UnifiedSessionScreen.tsx`
**Lines 1570, 1591, 1611, 1630**

Dynamic `require('@meet-without-fear/shared')` calls inside `useCallback` hooks when `Stage4SubChatAnchor` is already statically imported at the top of the file. These bypass tree-shaking and add unnecessary module lookups.

### 🟡 Loose `string` types in `TendingResponseDTO`
**`shared/src/dto/strategy.ts:302–315`**

`status` and `continueChoice` are typed as `string` instead of their corresponding enums (`TendingEntryStatus`, `ContinueChoice`). The mobile code uses hard-coded inline literal unions that aren't validated at the type boundary.

### 🟡 `stage4Query.error` silently swallowed
**`mobile/src/screens/UnifiedSessionScreen.tsx:1465–1470`**

`stage4Query.error` and `stage4Query.isError` are never checked. If the stage 4 state fetch fails, the panel renders empty with no user feedback.

### 🟡 `useStage4State` has `staleTime: 0`
**`mobile/src/hooks/useStages.ts:2032`**

Every mount/focus triggers a full refetch of the large stage 4 payload. Consider a non-zero staleTime (10–30s) and rely on Ably events + `setQueryData` for live updates.

## Minor Notes

- `NEEDS_DISCUSSION` → `NOT_WILLING` enum migration is semantically lossy (schema migration `20260512000000`)
- `Stage4Phase` enum defined in `strategy.ts` instead of `enums.ts` (inconsistent with all other Stage 4 enums)
- `sendMessageStream` uses raw `{ error: string }` instead of the standard `errorResponse()` envelope
- `&ldquo;` / `&rdquo;` HTML entities in `Stage4RedesignPanel.tsx:258–259` will render literally in React Native `<Text>`
- `publishSessionEvent` not wrapped in try/catch in `submitTendingResponse` (tending.service.ts:342–346), unlike the identical call in `setIndividualEntryShare`

## Verdict

**Request changes** — The IDOR in `confirmNeeds` adjustments and the private-entry visibility bug in `TendingCheckinScreen` are security/privacy issues that need fixing before merge. The cache race condition should also be addressed to avoid flaky UI state. The rest are important but non-blocking improvements.

The overall architecture is solid, test coverage is thorough, and the Stage 4 state machine design is well-thought-out. Great work on the scope of this PR. 🤖

slam-paws (commented):
## PR #640 Review — Stage 3/4 Rework + Full Tending Flow

Massive PR (135 files, +16,820/−2,888). The architecture is sound overall — solid test coverage, clean commit history, and good use of tool-call/state-capture patterns for Stage 4. Below are the findings organized by severity.

---

### 🔴 Issues to Address Before Merge

**1. Permission gap — partner needId accessible in `requestSuggestions`**
`backend/src/controllers/stage4.ts` — The `targetNeed` lookup only checks `vessel.sessionId`, not `vessel.userId`. A user could pass their partner's `needId` to generate AI suggestions for partner's needs.
**Fix**: Add `userId: user.id` to the `vessel` where clause.

**2. Data exposure risk — Tending context loads all needs**
`backend/src/controllers/messages.ts` — `getTendingConversationContextForPrompt()` fetches `stage4NeedCoverage` data without filtering by `sourceUserId`. Partner's identified needs may leak into the current user's Tending conversation prompt.
**Fix**: Filter `stage4NeedCoverage` by `sourceUserId` or use the existing `getSourceLabel()` helper to scope.

**3. Race condition in Tending coordination cycle creation**
`backend/src/services/tending.service.ts` — The coordination cycle logic uses `findFirst()` + `update()` without preventing duplicate cycle creation. Two users submitting tending checks simultaneously could create separate coordination cycles.
**Fix**: Use a unique constraint on `(sessionId, sharedEntryIds)` or use `upsert` with explicit keys.

---

### 🟡 Warnings (Recommended Fixes)

**4. `refiningNeedId` validation incomplete** — `messages.ts` validates that the need exists in the user's vessel but doesn't check if it's still editable (`lockedAt`, `deletedAt`). User could refine a locked/deleted need.

**5. Missing FK constraints on `TendingNeedOutcome`** — `sourceUserId` and `needId` are nullable but have no foreign key constraints to `User` / `IdentifiedNeed`. Data integrity risk with orphaned references.

**6. `needRefinementThreads` stored in `useState` instead of React Query cache** — `UnifiedSessionScreen.tsx` stores refinement thread history in local state. This violates the "Cache-First" rule and will lose data on component unmount/re-render.

**7. React Query invalidation after `setQueryData`** — `useStages.ts` `useConfirmNeeds` hook calls `invalidateQueries` after `setQueryData`, which triggers a refetch that could overwrite the optimistic update.

**8. Missing `Keyboard.dismiss()` in modal transitions** — Several keyboard dismiss calls were removed from empathy validation flows and need refinement drawer transitions. Could cause keyboard overlapping modals.

**9. Missing index on `TendingEntryOutcome.responseId`** — FK constraint exists but no index for reverse queries (`TendingResponse` → `TendingEntryOutcome`). Will be slow for joined reads.

---

### 🔵 Notes (Non-blocking)

- **Enum overlap**: `ContinueChoice` and `TendingNextAction` share some values (`FULL_CLOSURE`, `EXTEND`, `NEW_PROCESS`, `PARTIAL_CLOSURE`) but `ANOTHER_ROUND` exists only in `ContinueChoice`. Document the mapping.
- **TendingCheckinScreen state explosion**: 15+ `useState` calls managing step navigation. Would benefit from `useReducer` for maintainability.
- **Default check-in date changed from 28 → 10 days** (`Stage4RedesignPanel.tsx:103`) — undocumented product decision.
- **Event publishing failures silently swallowed** — `messages.ts` catches realtime event failures with `.catch()` + warning log only. Partner could miss need update notifications.
- **Two needs modals could open simultaneously** — `showNeedsDrawer` and `showNeedsRevealModal` are independent booleans with no mutual exclusion guard.

---

### ✅ Positives

- **Clean NeedsDrawer → NeedRefinementDrawer migration** — 668-line bottom sheet removed, replaced with focused 164-line refinement drawer
- **Correct Ably event patterns** — New `need.captured/refined/locked/deleted` events use `setQueryData` (not `invalidateQueries`), following cache-first rules
- **Stage 4 walkthrough optimistic updates** — `useUpdateStage4WalkthroughNeed()` properly uses `onMutate` + `onError` rollback
- **Comprehensive test coverage** — 115+159+130+76 tests passing across backend, mobile, E2E, and moment-eval
- **All new routes have `requireAuth` + `requireSessionAccess`** middleware
- **All new POST endpoints validated with Zod schemas**
- **Migrations are additive-only** — safe for production, no data loss risk
- **ErrorBoundary added** — `SessionScreenErrorBoundary` wraps the full session screen

🤖 Reviewed by Slam Paws

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Major rework of Stages 3/4 and a complete Tending flow implementation. Impressive scope — 6 new migrations, new services, comprehensive test additions, mobile UI overhaul. The architecture is solid overall, but there are several authorization gaps and a couple of critical schema issues that need addressing before merge.

## Blocking Issues

### 🔴 CRITICAL: Schema/migration mismatch for `coordinationKey`
`backend/prisma/migrations/20260522121500_add_tending_coordination_key/migration.sql` adds a `coordinationKey` column and indexes to `TendingCoordinationCycle`, but `schema.prisma` has no corresponding field. After applying this migration, `prisma migrate dev` would try to drop the column. The migration file may also not be committed to the branch (appears in the diff but wasn't found in the tree).

### 🔴 Authorization bypass: need adjustments in `confirmNeeds`
`backend/src/controllers/stage3.ts:403-414` — The `adjustments` array passes `adj.needId` directly to `prisma.identifiedNeed.update` without checking that the need belongs to the requesting user's vessel. The outer `needIds` array is correctly validated against `userVessel.id`, but adjustments are not. An authenticated user could rewrite any need in the database.

**Fix:** Add `where: { id: adj.needId, vesselId: userVessel.id }` to the update.

### 🔴 Authorization bypass: `declineStage4Need` / `undeclineStage4Need`
`backend/src/controllers/stage4.ts:746-811` — `needId` URL param is used directly without verifying it references a real `stage4NeedCoverage` row in this session. Arbitrary UUIDs create phantom declinations that can influence the "all needs addressed" gate check in `shareStage4Selections`.

### 🔴 Thinking buffer leaked to client
`backend/src/controllers/messages.ts:1808-1812` — When the `<thinking>` block exceeds 2000 chars without closing, `sendCleanText(thinkingBuffer)` flushes raw AI reasoning (including internal checks like `FeelHeardCheck:Y`, proposed strategies, etc.) directly to the SSE stream. This should discard the buffer, not stream it.

### 🔴 Realtime channel collision
`shared/src/dto/realtime.ts:153-155` — `REALTIME_CHANNELS.session` and `REALTIME_CHANNELS.presence` both return `meetwithoutfear:session:${sessionId}`. Presence subscriptions will receive all session events and vice versa. Looks like a copy-paste bug.

## Important Issues

### 🟡 Missing stage-4 progress checks
`backend/src/controllers/stage4.ts` — `shareStage4Selections`, `declineStage4Need`, `undeclineStage4Need`, and `unshareStage4Selections` don't verify the user is currently in stage 4 (unlike `submitStage4SelectionsInternal` and `closeStage4` which do). Users in stage 3 or post-stage-4 can invoke these.

### 🟡 `addCustomNeed` skips enum validation
`backend/src/controllers/stage3.ts:895-902` — `category` is type-asserted but not runtime-validated against `NeedCategory` enum (unlike `captureNeeds` which calls `isNeedCategory()`).

### 🟡 `validateNeeds` default defeats guard
`backend/src/controllers/stage3.ts:677` — `const { validated = true }` means an empty body silently validates needs rather than returning 400.

### 🟡 Tending race conditions
- `backend/src/services/tending.service.ts:287-340` — TOCTOU between `getEntryForUser` status check and the transaction in `submitTendingResponse`. A cancelled entry could accept a response.
- `backend/src/services/tending.service.ts:326-338` — `responseCount` read inside transaction can be stale under `READ COMMITTED`, causing two simultaneous submissions to both mark `PARTIAL` instead of one marking `COMPLETED`.
- `backend/src/services/tending.service.ts:466-497` — `openEntries` fetched outside the transaction in `submitTendingCheckin`.

### 🟡 `TendingResponse.status` and `TendingReminder.status` are raw strings
`backend/prisma/schema.prisma:1317,1389` — Both use `String` instead of proper enums, losing DB-level constraint checking. The Zod schema also only validates `z.string().min(1).max(80)` — any string is accepted.

### 🟡 `closeStage4` uses raw selections instead of gate flag
`backend/src/controllers/stage4.ts:940-943` — `bothPartnersSubmitted` checks for any selection record rather than the `selectionSubmitted` gate, inconsistent with how the UI determines submission status.

### 🟡 Prompt injection via `intent` field
`backend/src/services/tending.service.ts:393-407` — User-supplied `intent` is interpolated into `buildReentrySummary` which feeds into AI prompts. Only length-limited to 1000 chars, no sanitization of control characters or prompt-injection tokens.

### 🟡 `as any` casts hiding real issues
- `backend/src/controllers/messages.ts:56` — `select: { privacyPreferences: true } as any` hides a likely schema mismatch
- `backend/src/controllers/messages.ts:730` — `(result.shareOffer as any).reason` masks a renamed field

### 🟡 `TendingBetweenPeriodNote.content` — private data without RLS
`backend/prisma/schema.prisma:1424-1440` — These notes have an explicit `consentToShareWithPartner` field but no RLS policy, unlike other sensitive vessel data.

## Minor / Nice-to-haves

- Schema: `TendingNeedOutcome.sourceUserId` (line 1363) has no FK relation or index
- Schema: `TendingNeedOutcome.needId` (line 1361) has no FK to `IdentifiedNeed`
- Schema: `TendingResponse` missing standalone `@@index([userId])`
- Schema: `TendingAdjustment.privacyScope` reuses `TendingReminderScope` enum (naming awkwardness)
- Schema: No compound `@@index([vesselId, deletedAt])` for `IdentifiedNeed` soft-delete queries
- `stage3.ts:570-582`: `consentToShareNeeds` uses `create` instead of `upsert`, producing duplicates on retry
- `tending.service.ts:396-420`: No cap on re-entry entries per session
- `tending.service.ts:688`: `openDueTendingEntries` only opens `SCHEDULED_SHARED_AGREEMENT_CHECKIN` — individual commitment entries never auto-opened
- `enums.ts:95,216`: `NeedCategory` vs `NeedsCategory` naming collision
- `micro-tag-parser.ts:138-150`: `targetProposalId` and `ownerUserId` from AI output not UUID-validated
- N+1 patterns in `openDueTendingEntries` (sequential updates) and scheduling loops (sequential creates)

## Verdict

**Request changes** — The authorization bypasses in `confirmNeeds` adjustments and `declineStage4Need`, the thinking-buffer leak, and the schema/migration mismatch are blocking. The realtime channel collision is also likely a bug. The race conditions and missing stage checks are important but could be addressed in a fast follow-up if needed.

Great work on the test coverage and the overall architecture of the Tending flow.

slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Massive PR combining Stage 3/4 rework + full Tending flow. The scope is large (~17k additions) but well-structured: new schema models, backend services, mobile screens, and solid test coverage (115 + 159 + 130 backend/mobile tests, 76 moment-eval, 7 E2E). The diff exceeded GitHub's 20k-line limit so I reviewed files individually by priority.

## High Priority

- **[Auth] `stage3.ts` ~line 404 — Missing `vesselId` check on need adjustments in `confirmNeeds`**: The `needIds` array is correctly validated against the user's vessel, but the `adjustments` array processes `adj.needId` without verifying it belongs to the same vessel. An authenticated user could supply arbitrary need IDs and overwrite another user's need text. **Fix:** add `vesselId: userVessel.id` to the `where` clause on the inner `prisma.identifiedNeed.update`.

- **[Logic] `tending.service.ts` ~line 531–561 — Single user can unilaterally reset Stage 4 via `ANOTHER_ROUND`**: The `submitTendingCheckin` handler for `continueChoice: ANOTHER_ROUND` deletes `stage4Closure`, all selections, coverage, and declinations, then sets session status back to `ACTIVE` — affecting both users. There's no check that both partners agree to another round. A single member can reset the entire Stage 4 lifecycle.

- **[Types] `shared/src/dto/strategy.ts` lines ~360, 508–511 — `continueChoice` typed as bare `string`**: Both `TendingResponseDTO.continueChoice` and `SubmitTendingResponseRequest.continueChoice` are typed as `string | null` / `string`, but the Prisma schema uses the `ContinueChoice` enum. This loses all compile-time safety across the mobile→backend boundary. Should be `ContinueChoice | null` and `ContinueChoice | undefined` respectively. Same issue with `TendingResponseDTO.status` and `TendingReminderDTO.status`.

## Medium Priority

- **[Race] `stage3.ts` ~line 776–806 — Both-validated advance not transactional**: When both users click "validate" simultaneously, both read `partnerValidated === true` and both execute the `updateMany + upsert` block. Wrapping lines 776–840 in a `$transaction` would prevent duplicate stage-4 progress rows.

- **[Atomicity] `stage3.ts` ~line 570–582 — Consent records created in non-atomic loop**: `prisma.consentRecord.create` in a `for` loop means partial consent state on failure. Should use `createMany` or `$transaction`.

- **[Logic] `stage4.ts` ~line 940–981 — `bothPartnersSubmitted` derived from selection rows, not gate flags**: The closure path infers submission from the presence of `Stage4ProposalSelection` rows rather than checking the `hasSubmittedSelectionGate` flag. A stale selection on a removed proposal would count. Also, `NO_SHARED_AGREEMENT` closure has no gate requiring both partners' input.

- **[Logic] `tending.service.ts` ~line 580–596 — New session from `NEW_PROCESS` has no StageProgress or invitation**: `session.create` with `status: 'CREATED'` but no `StageProgress` rows or invitation records. If initialisation isn't fully lazy, the partner can't join.

- **[Logic] `tending.service.ts` ~line 686–689 — Individual commitment check-ins never auto-open**: `openDueTendingEntries` only queries `SCHEDULED_SHARED_AGREEMENT_CHECKIN` entries. Individual commitment check-ins stay `SCHEDULED` indefinitely after their `scheduledFor` date.

- **[Schema] `TendingEntry.agreementId` FK uses `ON DELETE CASCADE` for a nullable relation**: Deleting an `Agreement` silently cascades through all associated `TendingEntry` rows (and their responses). For an optional FK, `ON DELETE SET NULL` would preserve tending history.

- **[Schema] `TendingResponse.status` is untyped `TEXT`**: Unlike `TendingEntry` which uses a proper enum, `TendingResponse.status` is a bare `String` column. Should be a `TendingResponseStatus` enum.

- **[Types] `shared/src/dto/need-edits.ts` — `NeedEditOperation` should be a discriminated union**: All payload fields are optional regardless of `type`, so `{ type: 'addNeed' }` (missing `text`) passes the type check. No Zod schemas exist for the two new endpoints, leading to `as any` casts in the controller.

- **[Mobile] `UnifiedSessionScreen.tsx` ~line 1569–1632 — `require()` inside `useCallback`**: Four callbacks use `require('@meet-without-fear/shared')` at runtime instead of the existing top-level import. Should add `Stage4SubChatAnchor` to the top-level import.

- **[Mobile] `TendingCheckinScreen.tsx` ~line 275 — Static `StyleSheet.create` without palette**: Uses raw `colors` import instead of the `(palette) => StyleSheet.create()` factory pattern used elsewhere. Won't respond to dark-mode changes.

- **[Mobile] `TendingPanel.tsx` ~line 149–163 — Form state not reset on entry change**: `reflection`, `statusChoice`, and `continueChoice` are independent `useState` values never reset when `selectedEntry` changes. Stale values persist across entry navigation.

- **[Mobile] `ChatBubble.tsx` ~line 110–117 — Animated.Value mutations during render**: `fadeAnim.setValue()` and ref mutations run in the render phase, not in a `useEffect`. In React 18 concurrent mode, this block could run multiple times per commit.

## Low Priority / Nice-to-haves

- **`UnifiedSessionScreen.tsx`**: ~40 `console.log` statements (including `[TIMING]` probes) left in production path.
- **`Stage4RedesignPanel.tsx` ~line 258**: `&ldquo;` / `&rdquo;` HTML entities won't render in React Native `Text` — use Unicode curly quotes.
- **`sharing.ts` ~line 865–892**: Concurrent `refine` calls on a share offer yield a 500 instead of 409; the `'Share offer already processed'` error should be caught and returned as a conflict.
- **`sharing.ts` ~line 56, 484**: `any` types on `findReconcilerResultWithRetry` and `generateShareSuggestion` — should use `ReconcilerResult`.
- **`messages.ts` ~line 851–879**: Duplicate message detection with nested loops runs on every `GET /messages` in production — should be gated behind a debug flag.
- **`chatUIState.ts` ~line 738**: `waitingBannerText` hardcodes `'Partner'` — comment says "will be replaced in the hook" but unclear if the hook actually does.
- **`ChatBubble.tsx` ~line 459**: Hardcoded `rgba(245, 158, 11, 0.15)` bypasses palette/dark-mode.
- **`useStages.ts` ~line 316**: Inline string array key instead of `stageKeys.empathyStatus()` helper.
- **`shared/src/dto/strategy.ts`**: `Stage4WalkthroughPhase`, `Stage4NeedWalkthroughStatus`, `Stage4ProposalSourceLabel` union types belong in `enums.ts`, not `strategy.ts`. `Stage4NeedWalkthroughStatus` uses `lower_snake_case` inconsistent with codebase `UPPER_SNAKE_CASE` convention.
- **Schema**: `Stage4ProposalRevision.sessionId` is indexed but has no FK; `TendingResponse` is missing a standalone `@@index([userId])`.
- **`NeedRefinementDrawer.tsx`**: Listed in the diff as a new 164-line file but not found on the branch — may have been renamed or dropped during rebase.

## Verdict

**Comment** — no blocking approval or change-request given the scope. The two highest-priority items (vessel ownership check on need adjustments, unilateral Stage 4 reset via ANOTHER_ROUND) are worth addressing before merge. The type safety issues (`continueChoice: string`) are straightforward fixes. Everything else is follow-up material. Impressive amount of work with thorough test coverage.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Large, ambitious PR combining Stage 3/4 rework with the full Tending flow. The architecture is sound — proper auth on all routes, Prisma parameterized queries throughout (no SQL injection risk), clean service separation, and good test coverage (404+ tests passing). Several issues need attention before merge, ranging from security-sensitive ownership checks to race conditions and cache-rule violations.

## High Severity

### 1. **[Security]** Adjustments loop skips ownership check — `controllers/stage3.ts:402–415`
`confirmNeeds` validates that `needIds` belong to the user's vessel, but the `adjustments` array is a separate field whose `adj.needId` values are **never cross-checked** against `userVessel.id`. An authenticated user could include need IDs from another user's vessel in the `adjustments` array to overwrite their partner's needs. Fix: add `vesselId: userVessel.id` to the update `where` clause.

### 2. **[Correctness]** `closeStage4` race condition — `controllers/stage4.ts:905–1099`
The check-then-create pattern (`findUnique` → create inside `$transaction`) is not atomic. Two concurrent requests can both pass the `findUnique` guard before either writes. The second writer gets an unhandled P2002 unique violation (500 instead of 409). Use `upsert` or catch the unique constraint error. Same pattern in `shareStage4Selections`, `unshareStage4Selections`, `declineStage4Need`, `undeclineStage4Need`.

### 3. **[Correctness]** `bothPartnersSubmitted` uses wrong gate — `controllers/stage4.ts:940–943`
Checks if each partner has *any* selection (including private/unshared ones). Should use the explicit `hasSubmittedSelectionGate` from progress rows, which is used everywhere else. Current code can allow closing Stage 4 into `SHARED_AGREEMENT` based on a partner's private stances before they explicitly share.

### 4. **[Privacy]** Thinking buffer leak — `controllers/messages.ts:1807–1813`
When the thinking block exceeds 2000 chars before `</thinking>` arrives, the raw AI internal reasoning is flushed to `sendCleanText` and sent to the user. The overflow handler should **discard** the buffer, not forward it.

### 5. **[Security]** `ownerUserId` from AI output not validated — `utils/micro-tag-parser.ts:148–150`
The `ownerUserId` field is taken verbatim from AI-generated JSON and stored in `strategyProposal.createdByUserId`. Nothing validates it's a member of the current session. A prompt injection could attribute proposals to arbitrary users. Enforce membership check before storing.

### 6. **[Correctness]** `ANOTHER_ROUND` doesn't reset stage progress — `services/tending.service.ts:531–561`
When `ANOTHER_ROUND` is chosen, Stage 4 closure/selections/coverage are deleted and session set to `ACTIVE`, but `stageProgress` rows keep `status: 'COMPLETED'`. Users returning to Stage 4 will be blocked because helpers check for `status: 'IN_PROGRESS'`.

## Medium Severity

### 7. **[DB]** `TendingCoordinationCycle.createdByUserId` uses `ON DELETE CASCADE` — `migration 20260522093000`
If a user account is deleted, all their coordination cycles are cascade-deleted, orphaning the session's coordination history. Other user-created content uses `SET NULL`. This should be `SET NULL` for consistency and data preservation.

### 8. **[Cache Rule]** `invalidateQueries` on message keys with optimistic updates in-flight — `hooks/useStages.ts:321,367`
`useRespondToShareOffer.onSuccess` and `useGenerateShareDraft.onSuccess` call `invalidateQueries({ queryKey: messageKeys.infinite(sessionId) })`. Per project rules: "Never invalidateQueries on a key with optimistic updates in-flight." Use `setQueryData` from the server response instead.

### 9. **[Correctness]** `isRespondable` called without `currentUserId` — `TendingCheckinScreen.tsx:104`
`isRespondable(entry)` never receives the current user's ID, so `INDIVIDUAL`-scope entries will pass the ownership check for the wrong user (since `currentUserId` is `undefined`). Thread `currentUserId` through props.

### 10. **[Performance]** N+1 consent creation + no upper bound — `controllers/stage3.ts:568–582, 219`
`consentToShareNeeds` creates one DB record per need in a sequential loop outside a transaction. No upper bound on `needIds` length — an authenticated user could submit hundreds. Use `createMany` inside a transaction and cap the array length (e.g., ≤ 20).

### 11. **[DB]** `TendingReminder.status` is plain TEXT — `migration 20260522000000`
Uses `String @default("SCHEDULED")` with no enum constraint. Invalid status strings can be inserted without error. Use an enum for type safety.

### 12. **[Mobile]** HTML entities in React Native — `Stage4RedesignPanel.tsx:258`
`&ldquo;` / `&rdquo;` are not decoded in React Native `<Text>` — they render as literal text. Use Unicode: `\u201cwilling\u201d`.

### 13. **[Date Validation]** `checkInDate` / `followUpDate` not range-validated — `controllers/stage4.ts:1007,1761`
Validated as syntactically valid ISO strings but no range check. A caller can submit `1970-01-01`, causing tending entries to open immediately.

## Low Severity / Suggestions

- **Duplicate index on `coordinationKey`** (`migration 20260522121500`): The plain index is redundant given the partial unique index. Drop one.
- **Pervasive `any` casts in `setQueryData`** (`UnifiedSessionScreen.tsx`): 15+ occurrences of `(old: any) =>`. Use `setQueryData<SessionStateResponse>(...)` for type safety.
- **Local-only enum types** (`TendingPanel.tsx`): `TendingStatusChoice` / `TendingContinueChoice` should be promoted to `shared/src/enums.ts` for compile-time backend/mobile agreement.
- **`ChatBubble.tsx:82–85`**: Uses `(message.role as string) === 'SHARED_CONTEXT'` instead of `MessageRole.SHARED_CONTEXT` enum values.
- **Missing index on `TendingEntryOutcome.responseId`** (FK column with no index).
- **`TendingAdjustment.privacyScope`** reuses `TendingReminderScope` enum — naming mismatch.
- **E2E auth bypass** (`middleware/auth.ts:77–81`) relies solely on `NODE_ENV !== 'production'`. Consider requiring an explicit `ALLOW_E2E_BYPASS=true` flag.

## What's Good

- Auth is consistently applied — every new route uses `requireAuth` + `requireSessionAccess`. No unauthenticated endpoints.
- Zero raw SQL — all queries use Prisma's parameterized builder.
- `useSubmitStage4ProposalSelection.onMutate` is a textbook optimistic update with proper rollback.
- `chatUIState.ts` is a clean, testable pure-derivation function with no side effects.
- `micro-tag-parser.ts` has thorough type-narrowing on all AI-sourced fields — no `eval`, no `Function` constructors.
- Migration ordering is correct and schema consistency between SQL and Prisma is verified.
- Comprehensive test coverage: 404+ tests across backend, mobile, E2E, and moment-eval.

## Verdict

**Request changes.** Items 1–6 should be addressed before merge — the ownership check bypass (#1) and thinking buffer leak (#4) are the most urgent. The rest are medium-priority improvements that would strengthen the PR but could be tracked as follow-ups if needed.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Massive PR combining Stage 3/4 rework with the full Tending flow (~17k additions across 140+ files). The architecture is well-structured — clean separation of TendingEntry/TendingResponse, proper use of tool-call/state capture for Stage 4, and good test coverage (400+ tests passing). However, there are several security, atomicity, and cache-consistency issues that should be addressed before merge.

## Critical Issues

### 1. **[Security] IDOR on needs adjustments** — `backend/src/controllers/stage3.ts:403`
`confirmNeeds` verifies `needIds` belong to the user's vessel (lines 385–399), but `adjustments[].needId` values are **not** checked against the same set. An attacker who knows another user's need ID can modify any `identifiedNeed` row by placing it in `adjustments`. Fix: filter `adjustments` to only those where `adj.needId` is in the already-verified `userNeeds` set.

### 2. **[Race Condition] TOCTOU on Stage 4 closure** — `backend/src/controllers/stage4.ts:905`
The `closeStage4` handler checks for an existing closure outside the transaction, then creates one inside `$transaction`. Double-tap = two requests pass the guard → second hits a unique-constraint error surfaced as 500. Same pattern in `shareStage4Selections` (line 561) and `unshareStage4Selections` (line 837). Fix: move the check inside the transaction or catch `P2002` and return 409.

### 3. **[Atomicity] Consent records not transactional** — `backend/src/controllers/stage3.ts:569`
`consentToShareNeeds` creates consent records in a plain `for` loop and updates `stageProgress` separately. Partial write on crash = inconsistent state. Fix: wrap in `prisma.$transaction`.

### 4. **[Atomicity] Stage 3→4 transition not atomic** — `backend/src/controllers/stage3.ts:775`
Stage completion marking, Stage 4 progress upserts for both users, and transition messages are all individual Prisma calls. Crash mid-sequence = one partner in Stage 4, the other stuck in Stage 3. Fix: wrap in `$transaction`.

### 5. **[Atomicity] Agreement confirmation not atomic** — `backend/src/controllers/stage4.ts:1921`
`agreement.update` + `session.update` + `stageProgress.updateMany` are three separate DB calls. Fix: wrap in `$transaction`.

### 6. **[Authorization] Unilateral session-wide destructive ops** — `backend/src/services/tending.service.ts:530`
For `ContinueChoice.ANOTHER_ROUND`, a single user can delete all `stage4Closure`, `stage4ProposalSelection`, `stage4NeedCoverage`, `stage4NeedDeclination` records and reset the session to `ACTIVE` — without partner consent. Is this intentional? If not, require both partners to have submitted check-ins before applying destructive transitions.

### 7. **[Validation] `status` accepts arbitrary strings** — `backend/src/controllers/tending.ts:17`
`submitTendingResponseSchema` validates `status` as `z.string().min(1).max(80)` instead of `z.nativeEnum(TendingEntryStatus)`. Same issue with `continueChoice`. Invalid strings will either cause a 500 (if DB is enum) or silently store garbage (if TEXT).

### 8. **[Validation] No upper bound on `needs` array** — `backend/src/controllers/stage3.ts:219`
`captureNeeds` checks `needs.length > 0` but has no cap. An attacker could submit 10k needs, causing a large delete-and-recreate transaction. Add a reasonable limit (e.g. 50).

## High Severity

### 9. **[Data Integrity] Cross-session proposal leak** — `backend/src/services/tending.service.ts:372`
`buildReentrySummary` queries `strategyProposal` by `id: { in: individualIds }` without scoping to `sessionId`. If `closure.individualProposalIds` contains stale/incorrect IDs, proposals from another session leak into the re-entry summary. Fix: add `sessionId` to the `where` clause.

### 10. **[Data Integrity] Individual commitment entries never auto-opened** — `backend/src/services/tending.service.ts:685`
`openDueTendingEntries` only queries `SCHEDULED_SHARED_AGREEMENT_CHECKIN`. Individual commitment check-in entries created via `scheduleIndividualCommitmentTendingEntries` will stay `SCHEDULED` forever.

### 11. **[Migration] Lossy enum remap** — `NEEDS_DISCUSSION` → `NOT_WILLING`
This is semantically destructive and irreversible. If any production rows existed with `NEEDS_DISCUSSION`, their meaning is permanently lost. Consider preserving the enum value or adding an audit trail before remapping.

### 12. **[Mobile] Cache-first violations** — `mobile/src/hooks/useStages.ts:2395-2436`
- `useSubmitTendingResponse` calls `refetchQueries` after `setQueryData` (violates CLAUDE.md Rule 7 — risks overwriting optimistic data with stale refetch)
- `useSubmitTendingResponse` and `useSetTendingEntryShare` lack `onMutate` optimistic updates (violates Rule 2)
- `useSubmitTendingCheckin` has no `setQueryData`, only `refetchQueries`

### 13. **[Mobile] TendingPanel form state not reset on entry change** — `mobile/src/components/TendingPanel.tsx:149`
`statusChoice`, `continueChoice`, and `reflection` are initialized once. If `selectedEntry` changes, these retain values from the previous entry. User could accidentally submit stale data.

### 14. **[Mobile] Duplicated footer logic with label mismatch** — `mobile/src/components/Stage4RedesignPanel.tsx:453,710`
Share/revise/close button logic is duplicated between `Stage4RedesignFooter` and the inline IIFE. The two copies use different labels ("Pull my stances back to revise" vs "I'm not ready yet"), causing inconsistent UX.

### 15. **[Types] `SharedContentDeliveryStatus` casing mismatch** — `shared/src/dto/empathy.ts` vs `backend/src/services/reconciler/sharing.ts:948`
DTO defines lowercase values (`'sending'`, `'delivered'`), but the backend writes uppercase `'DELIVERED'` to the DB. This is a runtime type lie.

### 16. **[Types] `TendingResponseDTO.status` and `continueChoice` are bare `string`** — `shared/src/dto/strategy.ts:302,304`
Both have corresponding enums defined in the same file. These should use `TendingEntryStatus` and `ContinueChoice | null` respectively.

## Medium Severity

- **`addCustomNeed` lacks Zod validation** (`stage3.ts:895`) — `category` not validated against `NeedCategory` enum
- **`consentToShareNeeds` doesn't validate `needIds` elements** (`stage3.ts:488`) — no CUID validation
- **`TendingCheckinScreen` uses static `colors` import** (`TendingCheckinScreen.tsx:8`) — no dark mode support
- **`NeedsDrawer` competing `useEffect`s** (`NeedsDrawer.tsx:183,193`) — two effects both animate on `visible` change
- **`TendingResponse.status` is untyped TEXT in DB** — no enum constraint at the database level
- **`ownerUserId` in `TendingEntry` has no FK to `User`** — dangling IDs after user deletion

## Good Patterns Observed

- Clean separation of `TendingEntry` (scheduled event) from `TendingResponse` (per-user answer)
- Tool-call/state capture paths for Stage 4 instead of leaking untagged reasoning
- Composite index `(sessionId, scope, ownerUserId)` on `TendingEntry` — well-chosen
- Proper `ON DELETE CASCADE` on most FK constraints
- Comprehensive test coverage across backend, mobile, E2E, and moment-eval
- No hardcoded secrets detected

## Verdict

**Request changes** — the IDOR on needs adjustments (#1) is a security vulnerability that must be fixed before merge. The atomicity issues (#2–5) are high-risk for production state corruption. The validation gaps (#7–8) should also be addressed. The remaining items are important but could be tracked as follow-ups if needed.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Large PR combining Stage 3/4 rework (need editing, refinement, sharing redesign) and the full Tending flow (structured check-ins, coordination cycles, between-period notes, adjustments, reminders). Well-tested with 115+159+130+76 passing tests and E2E coverage. The architecture is solid overall, but there are two privilege-escalation issues and several data-integrity gaps that should be fixed before merge.

## Critical — Must Fix

### 1. `confirmNeeds` adjustment `needId` bypasses vessel ownership
**`backend/src/controllers/stage3.ts:403-414`**
In `confirmNeeds`, the `adjustments` loop updates needs using `adj.needId` directly without verifying the need belongs to the calling user's vessel:
```ts
await prisma.identifiedNeed.update({
  where: { id: adj.needId },  // ← no vesselId constraint
  data: { need: adj.correction, confirmed: adj.confirmed },
});
```
The `needIds` array *is* ownership-checked (lines 385-400), but `adjustments` is not. A user can overwrite any need in the DB, including their partner's private needs.
**Fix:** Add `vesselId: userVessel.id` to the `where` clause.

### 2. `interpretNeedEditRequest` creates vessel without session membership check
**`backend/src/services/needs-edit-interpreter.service.ts:108`**
`getOrCreateUserVessel(sessionId, userId)` will **create a new vessel** for any `(userId, sessionId)` pair without verifying the user is a session member. An authenticated user who knows a session ID they don't belong to can trigger vessel creation and receive a probe response.
**Fix:** Add a session membership check before calling `getOrCreateUserVessel`.

## High

### 3. Post-transaction reads return deleted/superseded needs
**`backend/src/services/needs-edit-applier.service.ts:228-233, 318-323`**
After `applyNeedEdits` and `applyNeedAction` complete their transactions, the final `findMany` has no `deletedAt: null` or `supersededByNeedId: null` filter. Deleted/superseded needs are returned to the client. Same issue in the interpreter (line 112-115) — deleted needs are sent to the LLM prompt.
**Fix:** Add `deletedAt: null, supersededByNeedId: null` to all post-transaction need reads.

### 4. `removeNeed` operation lacks vessel ownership in WHERE clause
**`backend/src/services/needs-edit-applier.service.ts:219-222`**
The `updateNeedText` path correctly verifies vessel ownership via `currentNeeds.find()`, but `removeNeed` passes only `{ id: operation.needId }`. Safety currently relies on `previewNeedEditOperations` throwing first, but this is fragile.
**Fix:** Change to `where: { id: operation.needId, vesselId: vessel.id }`.

### 5. `lock`/`delete` actions accept already-deleted needs
**`backend/src/services/needs-edit-applier.service.ts:297-300`**
`applyNeedAction`'s `lock` and `delete` branches look up needs without `deletedAt: null`, allowing locking a deleted need or deleting a locked need — creating inconsistent record states.
**Fix:** Add `deletedAt: null` to both lookup queries.

### 6. `declineStage4Need` accepts arbitrary `needId` from other sessions
**`backend/src/controllers/stage4.ts:761-765, 802-804`**
Both `declineStage4Need` and `undeclineStage4Need` accept `needId` from params without verifying it belongs to the current session's coverage rows.
**Fix:** Validate needId exists in session's `stage4NeedCoverage` before upserting.

### 7. `ContinueChoice` enum mismatch between mobile and shared
**`mobile/src/screens/UnifiedSessionScreen.tsx:1733-1735` vs `shared/src/dto/strategy.ts:312-316`**
Mobile uses a locally-typed inline union (`'CONTINUE' | 'ADJUST' | 'CLOSE' | 'NEW_PROCESS' | 'OTHER_TRACK'`) that does **not** match the `ContinueChoice` enum in shared (`ANOTHER_ROUND | EXTEND | NEW_PROCESS | PARTIAL_CLOSURE | FULL_CLOSURE`). These are entirely different vocabularies — the mobile client will send values the backend doesn't expect.
**Fix:** Use the shared `ContinueChoice` enum in mobile.

## Medium

### 8. `FULL_CLOSURE` resolves session while shared entries are still `PARTIAL`
**`backend/src/services/tending.service.ts:~1450`**
When a user submits `FULL_CLOSURE` with both shared and individual entries, the code resolves the session immediately. But shared entries are set to `PARTIAL` awaiting the partner's coordination submission. The coordination cycle resolver *also* resolves the session, creating a race.
**Fix:** Gate session `RESOLVED` in `FULL_CLOSURE` on `!hasSharedEntry`.

### 9. `missingUserIds` leaked in coordination timeout event
**`backend/src/services/tending.service.ts:~1637`**
The `tending_coordination_timed_out` Ably event publishes `submittedUserIds` and `missingUserIds` to the session channel. This reveals which specific user failed to submit their check-in — a privacy concern.
**Fix:** Replace with aggregate: `{ missingCount: missingUserIds.length }`.

### 10. No idempotency guard on `submitTendingCheckin`
**`backend/src/services/tending.service.ts:~1102`**
Double-tap or client retry creates duplicate `TendingCheckin` records. The `TendingResponse` upsert is idempotent, but `tx.tendingCheckin.create` is not.
**Fix:** Add unique constraint on `(sessionId, userId, coordinationCycleId)` or guard with an existence check.

### 11. Consent records created outside a transaction
**`backend/src/controllers/stage3.ts:569-603`**
`consentToShareNeeds` creates consent records in a sequential loop with no enclosing transaction. A crash mid-loop leaves partial consent state.
**Fix:** Wrap in `prisma.$transaction`.

### 12. `isRespondable` guard silently bypassed
**`mobile/src/screens/TendingCheckinScreen.tsx:88-104`**
`isRespondable(entry, currentUserId)` checks `INDIVIDUAL` scope ownership, but the call at line 104 never passes `currentUserId`. Partner's individual entries are always included in the respondable list.
**Fix:** Pass `currentUserId` from auth context.

### 13. `TendingNeedOutcome.needId` has no FK constraint
**Migration `20260522000000`**
`needId` has an index but no foreign key to `IdentifiedNeed`. Dangling references will silently accumulate when needs are deleted.

### 14. `TendingReminder.status` is raw `TEXT` instead of enum
**Migration `20260522000000`**
All other status fields use enums; this one accepts any string with no DB-level validation.

## Low

- **`shared/src/dto/realtime.ts:153-155`**: `REALTIME_CHANNELS.session` and `.presence` return identical strings — intentional or bug?
- **`UnifiedSessionScreen.tsx:2664-2684`**: `renderInlineCard` `useCallback` deps include ~12 dead references from removed switch cases, causing unnecessary FlatList re-renders.
- **`UnifiedSessionScreen.tsx:1563-1630`**: Dynamic `require('@meet-without-fear/shared')` inside callbacks instead of using top-level imports — breaks tree-shaking and fast refresh.
- **`stage4.ts:117`**: `followUpDate` is `z.string().optional()` with no date-format validation; `new Date("not-a-date")` produces Invalid Date.
- **`stage3.ts:901-904`**: `addCustomNeed` doesn't validate `category` against `NeedCategory` enum (unlike `captureNeeds`).
- **`TendingCoordinationCycle.createdByUserId`** FK uses `ON DELETE CASCADE` — deleting a user destroys shared coordination records. Consider `SET NULL`.

## What Looks Good

- Proper use of Prisma transactions for atomic need replacement in `captureProposedNeedsForUser`
- `validateOperationOwnerUserId` added to `stage4-capture.service.ts` — good defense-in-depth
- Coordination cycle unique constraint race handled with try/catch + retry
- Partial unique index on `coordinationKey` is well-designed
- Between-period note content length validated to 4000 chars
- Comprehensive test coverage across backend, mobile, E2E, and moment-eval
- All migrations are purely additive — no destructive operations

## Verdict

**Request changes** — The two critical privilege-escalation issues (#1, #2) and the enum mismatch (#7) need to be fixed before merge. The other high/medium items are strongly recommended but could be addressed in a fast-follow if needed given the PR size.

slam-paws:
On it — reviewing now.

slam-paws (commented):
## Summary

Massive PR combining two major features: Stage 3/4 rework (need editing, refinement, walkthrough-based Stage 4) and full Tending follow-up flow (structured check-ins, outcomes, coordination cycles, reminders, between-period notes). 140 files changed, 17k additions, 3k deletions. Test coverage is strong (3,700+ lines of test code added across backend, mobile, E2E, and moment-eval).

## Feedback

### Critical Bug Fix (Nice catch)
- **Realtime channel collision fixed** (`shared/src/dto/realtime.ts:159`): Presence channel was using the same `meetwithoutfear:session:` prefix as the main session channel. Now correctly uses `meetwithoutfear:presence:`. This would have caused cross-talk between presence and session events.

### Database & Schema — Clean
- 6 proper migrations (not db push), well-indexed, appropriate cascade strategies, nullable columns make sense. Self-referential need supersession pattern is clean.

### Items to Consider (Non-blocking)

- **[Type Safety]** `backend/src/services/tending.service.ts` — Several `any` casts for enum fields from the database (`followThroughStatus: any`, `helpfulnessStatus: any`, `blockerCategories: any[]`, `scope: any`, `privacyScope: any`). Consider adding type guards to validate these at the boundary rather than casting.

- **[Race Condition]** `backend/src/services/tending.service.ts` `submitTendingCheckin` — Multi-step state mutation (create checkin → create outcomes → update gates → publish events) doesn't appear to be fully wrapped in a single transaction. If a failure occurs mid-way, state could become inconsistent. Consider expanding the transaction scope.

- **[Unused Utility]** `backend/src/services/tending.service.ts` — `isUniqueConstraintError()` is defined but never called. If P2002 errors are possible in tending operations, they'll surface as 500s instead of 409s.

- **[Mobile Architecture]** `mobile/src/screens/UnifiedSessionScreen.tsx` is now ~5,000 lines with 34+ useState calls. While this PR isn't the right place to refactor it, consider decomposing into a container/presenter pattern in a follow-up. Same observation for `mobile/src/hooks/useStages.ts` (~2,800 lines).

- **[Realtime Version Guards]** `mobile/src/screens/UnifiedSessionScreen.tsx` — The `empathy.status_updated` handler has version checking, but other realtime events like `partner.advanced` and `partner.signed_compact` use `setQueryData` without version validation. Stale events could theoretically overwrite newer state.

- **[Recursive Retry]** `mobile/src/hooks/useRealtime.ts` (~line 438) — `setupSubscription()` calls itself recursively after token refresh. If cleanup races with the retry, this could create duplicate subscriptions. Consider a retry counter or extracting to a separate function.

### Positive Patterns

- Shared types are well-structured with proper discriminated unions and consistent DTO naming
- New need lifecycle realtime events (`need.captured`, `need.refined`, `need.locked`, `need.deleted`) provide granular state tracking
- Need edit flow uses proper validation errors (`NeedEditValidationError`, `NeedEditForbiddenError`) with clean error hierarchy
- `stripUntaggedReasoningPreamble()` prevents internal AI reasoning from leaking into chat
- Proper `setQueryData` usage for realtime events instead of `invalidateQueries` (prevents refetch races)
- Comprehensive test coverage: backend service tests (1,361 lines for tending alone), E2E stage 4 redesign tests, mobile component tests, moment-eval tests

## Verdict

**Comment** — This is a well-executed, large feature PR. The architecture is sound, test coverage is thorough, and the critical presence channel bug fix is a bonus. The items above are suggestions for hardening, not blockers. The `any` casts and transaction scope are the most actionable near-term improvements.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Large PR combining Stage 3/4 rework with the full Tending flow: 6 new migrations, new service layer for need editing, structured tending check-ins/coordination/reminders, and significant mobile UI changes. Well-tested (115+159+130+76 tests passing). The architecture is solid — the issues below are focused on race conditions, data integrity gaps, and cache consistency violations.

## Blocking Issues

### Database

1. **`TendingReminder.status` is `TEXT`, not an enum** — Every other status field in the schema uses a Postgres enum. A plain `TEXT DEFAULT 'SCHEDULED'` accepts any string, risking silent data corruption. Should be an enum like `TendingReminderStatus`.

2. **Partial unique index on `coordinationKey` is invisible to Prisma** — Migration 6 creates a filtered `WHERE status IN ('WAITING_FOR_PARTNER','READY_TO_RESOLVE')` unique index, but Prisma's schema has only `@@index([coordinationKey])`. Future `prisma migrate` runs could silently drop this constraint. Needs a schema comment + CI assertion to protect it.

3. **`TendingEntryOutcome` unique key `(checkinId, tendingEntryId)` may be too narrow** — In a two-partner flow, each user records their own outcome per entry. The current unique constraint blocks the second partner. If one outcome per partner is intended, the key should be `(checkinId, tendingEntryId, userId)`.

### Backend

4. **`closeStage4` transaction lacks `Serializable` isolation** (`stage4.ts`) — Other Stage 4 mutations (`shareStage4Selections`, `declineStage4Need`) correctly use `Serializable`, but the closure transaction does not. Concurrent partner closure calls can read stale state. The P2002 handler is a last-resort guard but shouldn't be the primary safety net. Add `{ isolationLevel: Prisma.TransactionIsolationLevel.Serializable }` and move reads inside the transaction.

5. **`ANOTHER_ROUND` path in `submitTendingCheckin` is a no-op** (`tending.service.ts`) — The reset logic was removed but no coordination mechanism replaces it. The session stays `RESOLVED` after a successful 200 response. Either return `501 Not Implemented`, gate behind a flag, or wire the coordination path.

6. **`declineStage4Need` / `undeclineStage4Need` don't validate `needId` against session** (`stage4.ts`) — `req.params.needId` is used directly in an upsert without checking it exists in `stage4NeedCoverage` for this session. An authenticated member can create junk rows with arbitrary IDs.

### Mobile / Shared Types

7. **`invalidateQueries(messageKeys.infinite)` in `onSuccess` while optimistic update is in-flight** (`useStages.ts:~321`) — Violates CLAUDE.md Rule 6. The `onMutate` writes an optimistic message; `onSuccess` then invalidates the same key, causing a refetch race. Use `setQueryData` instead.

8. **`isRespondable()` called without `currentUserId`** (`TendingCheckinScreen.tsx:~104`) — Individual-scope tending entries owned by the partner will be shown to the wrong user. `currentUserId` needs to be passed through props and forwarded to the filter.

9. **`TendingResponseDTO.status` is untyped `string`** (`shared/src/dto/strategy.ts`) — The mobile side defines its own local `TendingStatusChoice` union type. No shared enum exists, so a typo on either side silently produces invalid data. Move the enum to `shared/src/enums.ts` and reference it in the DTO.

## Non-Blocking Issues (should fix before ship)

10. **`checkInDate` schema contradiction** (`stage4.ts`) — Zod schema has `.min(1, 'checkInDate is required').optional()`. The `min` message says required, but `.optional()` allows omission (silently defaults to now+10 days). Pick one.

11. **`confirmNeeds` adjustment loop not transactional** (`stage3.ts`) — Sequential `updateMany` calls outside a `$transaction`. Crash between updates = partial write.

12. **`refreshStage4Caches` redundantly refetches after `setQueryData`** (`useStages.ts`) — Every Stage 4 mutation writes server data into cache via `setQueryData`, then immediately calls `refetchQueries` on the same key. This doubles the network load for every Stage 4 interaction.

13. **Duplicate `empathy.status_updated` handler** (`UnifiedSessionScreen.tsx:~818 and ~1043`) — Same event handled twice in `onSessionEvent`. Idempotent in the best case, but confusing and wasteful.

14. **`void partnerId` wastes a DB call** (`stage4.ts:~535`) — `getPartnerUserId` is called then immediately discarded. Remove the call or defer it.

15. **User `intent` embedded in AI context without sanitization** (`tending.service.ts:buildReentrySummary`) — Low risk currently (internal summary), but is prompt injection surface area.

16. **Missing `updatedAt` on `TendingCheckin` and `TendingEntryOutcome`** — Inconsistent with the rest of the schema; makes production debugging harder.

17. **12+ `setQueryData` calls use `(old: any)`** (`UnifiedSessionScreen.tsx`) — Untyped cache writes won't catch shape mismatches at compile time. Type the most critical updaters.

## What Looks Good

- Comprehensive test coverage across backend, mobile, E2E, and moment-eval
- Stage 4 `P2002` guard as defense-in-depth for closure races
- Proper use of `Serializable` isolation on `shareStage4Selections`
- Clean separation of `needs-edit-interpreter` and `needs-edit-applier` services
- Good use of Zod validation on all new endpoints
- Tending coordination model is well-designed for the two-partner async flow

## Verdict

**Request changes** — The 9 blocking items above are a mix of race conditions (#4), data integrity gaps (#1-3, #6), behavioral bugs (#5, #8), and cache consistency violations (#7, #9). Most are focused fixes. Happy to re-review once addressed.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary

Massive PR combining Stage 3/4 rework (need editing, refinement, sharing, walkthrough) and full Tending follow-up flow (structured check-ins, outcomes, adjustments, reminders, coordination cycles, between-period notes). 140 files, ~17K additions. Migrations are clean, test coverage is solid, and several important fixes land here (IDOR on need adjustments, presence channel collision, atomic consent transactions). However, there are a few issues that should be addressed before merge.

## Blocking Issues

### 1. **[Bug]** `TendingCheckinScreen.tsx` — `isRespondable` regression drops `currentUserId` guard

The old code had `&& currentUserId` before checking `entry.ownerUserId !== currentUserId`. The new code removed that guard, so when `currentUserId` is `undefined`, *all* individual entries are excluded (since `entry.ownerUserId !== undefined` is always `true`). This silently hides respondable entries.

```ts
// Old (correct):
if (entry.scope === TendingEntryScope.INDIVIDUAL && currentUserId && entry.ownerUserId !== currentUserId) return false;

// New (regression):
if (entry.scope === TendingEntryScope.INDIVIDUAL && entry.ownerUserId !== currentUserId) return false;
```

### 2. **[Cache Rules]** `useStages.ts:useConfirmNeeds` — `invalidateQueries` after `setQueryData` violates Rule 6

Per CLAUDE.md: "Never `invalidateQueries` on a key with optimistic updates in-flight." The code does `setQueryData` on `stageKeys.needs(sessionId)` then immediately calls `invalidateQueries` on the same key, racing against the optimistic write.

### 3. **[Security]** `middleware/auth.ts` — Silent Clerk ID overwrite on email-match merge

When a P2002 unique constraint fires and the fallback finds a user by email with a *different* `clerkId`, it silently overwrites the Clerk identity. This is an account-takeover vector: an attacker who controls a Clerk account with a known victim's email claims the victim's DB record. At minimum, this needs audit logging (old + new Clerk ID). Ideally, gate behind an explicit migration flag rather than auto-merging in the auth hot path.

### 4. **[Validation]** `controllers/stage3.ts:applyNeedEditPlan` — `operations` cast as `any` without schema validation

The controller checks `Array.isArray(operations)` but individual operation shapes (type, needId, newText, newCategory) are not validated with Zod before being passed to the service. Malformed or oversized payloads pass through.

## Should Fix

### 5. **[Stale Closure]** `UnifiedSessionScreen.tsx:handleSendNeedRefinementMessage` — `needRefinementThreads` captured by value

If two messages are sent in rapid succession, the second sees the stale pre-first-message thread when building `conversationHistory`. The AI loses context. Fix: use a `useRef` for the threads map or a functional update.

### 6. **[Logic]** `stage4-prompts.ts` — `whatHappened` and `whatWorked` phases share the same persona

```ts
whatHappened: TENDING_WHAT_WORKED_PERSONA,
whatWorked: TENDING_WHAT_WORKED_PERSONA,
```

If these are distinct phases, they should have distinct personas. The "what worked" persona will skip to wins before asking what actually occurred.

### 7. **[Data Integrity]** `tending.service.ts` — Transaction closure mutation pattern

`checkinDTO`, `coordinationCycleDTO`, `nextScheduledFor`, and `resultSummary` are mutated inside transactions via outer scope closure. If the transaction rolls back partway, these outer variables retain stale values that are then used post-transaction (e.g., in `publishSessionEvent`). Safer: return the values from `prisma.$transaction(...)`.

### 8. **[Two-Pass AI]** `ai-orchestrator.ts` — State-capture pass prose leaks as user response

The visible-response pass only fires when `!modelResponse && sessionStateTool`. If the state-capture pass returns even a fragment of prose text, the second pass is skipped and that fragment becomes the user-facing message. This is a soft instruction-following dependency with no fallback.

### 9. **[UX Migration]** `chatUIState.ts` — `computeShowNeedsReviewPanel` removed

Sessions in the old `needs-review` state (where `needsAvailable = true` but `allNeedsConfirmed = false`) will see no CTA after deploy. The new `needs-send` panel requires `allNeedsConfirmed`. Consider a transition gate for in-progress sessions.

### 10. **[DTO]** `shared/src/dto/strategy.ts` — `GetStage4StateResponse.walkthrough` is required but optional in practice

The field is required in the DTO, but `withStage4Defaults()` in the component handles the `undefined` case. Making it `walkthrough?: Stage4WalkthroughDTO` would be more honest and prevent issues during rollout.

### 11. **[Prompt Safety]** `stage-prompts.ts` — `tendingConversationPrompt` injected without stage guard

The tending prompt is appended to `dynamicBlock` for any stage where `promptContext.tendingConversationPrompt` is set. If a context assembler accidentally sets it on a non-Tending session, the prompt leaks into unrelated stages.

## Low Priority / Follow-up

- `bedrock.ts`: `as any` cast on `anthropic.messages.create` hides tool schema type errors
- `needs-edit-interpreter.service.ts`: AI-returned `NeedCategory` values cast without re-validation
- `needs.ts`: `description: need.need` duplicates the `need` field semantically
- `needs-edit-applier.service.ts`: Post-transaction re-fetch creates TOCTOU window; `deleteNeed` doesn't guard against re-deleting soft-deleted records
- `stage4-capture.service.ts`: Silent `ownerUserId` fallback should have a metric counter
- `context-assembler.ts`: Verify `currentMember?.nickname` is "what this member calls their partner" (not self-referential)
- `reconciler/sharing.ts`: `guesserName` parameter is now dead code after hardcoded static message
- `NeedRefinementDrawer.tsx`: `planByText` map can be overwritten by same-text proposals from different turns
- `useStages.ts`: `staleTime: 30_000` on `useNeeds` without reconnect sync; needs with no `createdAt` sort to epoch
- `TendingCheckinScreen.tsx`: `'TWO_WEEKS'` preset has no matching branch in reminderDate ternary
- `shared/src/dto/realtime.ts`: Presence channel rename is a coordinated-deploy breaking change — document in release notes
- `UnifiedSessionScreen.tsx`: `SafeAreaProvider` nested inside a Modal may produce incorrect inset calculations
- Schema: `TendingReminder.status` is a plain `String` instead of an enum; `TendingCheckin` lacks `@@unique([userId, coordinationCycleId])` constraint

## Positives

- Migrations are all proper migration files (no `db push`), additive, and safe for live data
- IDOR fix on `confirmNeeds` (adjustments now scoped to user's vessel) — important security fix
- Serializable transactions with closure guard on Stage 4 mutations — excellent
- Consent + stage transition now atomic in `$transaction`
- Presence channel collision fix (`session` → `presence`) — real bug fix
- `captureNeeds` 50-item upper bound — good DoS prevention
- Comprehensive test coverage across backend, mobile, E2E, and moment-eval
- `allNeedsConfirmed` now filters out deleted/superseded needs — correct semantics

## Verdict

**Request changes** — Items 1–4 should be addressed before merge. Item 1 is a functional regression, item 2 violates the project's cache-first rules, item 3 is a security concern in the auth path, and item 4 is missing input validation on a new endpoint. Items 5–11 are important but could be addressed in a fast follow-up if needed.

PR #641, opened 2026-05-22
[codex] Guard Stage 4 streaming leftovers

shantamg:
## Summary
- keep hidden Stage 4 thinking trapped until the closing tag, and keep malformed hidden buffers out of visible/persisted AI text
- prevent Stage 4 walkthrough/proposal state capture when the visible response is asking the user for clarification
- refresh Stage 4 caches from streaming metadata and use soft/hard streaming recovery timeouts on mobile
- keep the redesigned Stage 4 drawer reachable while hiding freeform input outside active needs walkthrough states

## Validation
- `npm test -- --runInBand src/routes/__tests__/messages.test.ts` from `backend`
- `npm test -- --runInBand src/hooks/__tests__/useStreamingMessage.test.ts` from `mobile`
- `npm run check` from `backend`
- `npm run check` from `mobile`

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -585,6 +585,13 @@ export function isReadyForStage3RevealText(content: string): boolean {
     /(list|lists|needs|side by side|reveal|see them|see it|show)/.test(normalized);
 }
 
+function isClarifyingStage4Response(content: string): boolean {
+  const trimmed = content.trim();
+  if (!trimmed.endsWith('?')) return false;
+
+  return /\b(?:do you mean|what do you mean|which|or something else|clarify|more specific|can you say more|are you saying)\b/i.test(trimmed);
+}
+
 async function getStage3GateResponse(sessionId: string, userId: string): Promise<string | null> {
   const partnerId = await getPartnerUserId(sessionId, userId);
   const progress = await prisma.stageProgress.findUnique({
@@ -2230,12 +2237,11 @@ Write only the user-facing conversational response. Do not include tool JSON, XM
               tagTrapBuffer = thinkingBuffer.substring(closingTagIndex + 11); // 11 = '</thinking>'.length
               thinkingBuffer = '';
             }
-            // Safety: If buffer grows without a closing tag, drop it rather
-            // than risk streaming hidden reasoning to the client.
+            // Safety: if the hidden preamble is long, keep waiting for the
+            // closing tag. Flushing here can expose chain-of-thought or tool
+            // planning to the user and persist it as a real AI message.
             else if (thinkingBuffer.length > 2000) {
-              logger.warn(`[sendMessageStream:${requestId}] Thinking buffer exceeded 2000 chars without closing tag; discarding hidden buffer`);
-              isInsideThinking = false;
-              thinkingBuffer = '';
+              logger.warn(`[sendMessageStream:${requestId}] Thinking buffer exceeded 2000 chars without closing tag; continuing to trap hidden text`);
             }
           }
           // PHASE 2: TAG TRAP - Buffer to catch hidden semantic tags before streaming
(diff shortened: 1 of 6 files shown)


slam-paws (approved):
## Review: [codex] Guard Stage 4 streaming leftovers

**Verdict: Approve** — well-structured defensive fixes addressing real production issues around Stage 4 streaming.

### Summary

Four cohesive changes:

1. **Thinking-buffer trap hardened** — instead of discarding the buffer at 2000 chars (which could cause the visible response to start from a blank slate if the closing tag arrives later), it now keeps trapping until `</thinking>` arrives or the stream ends. Prevents chain-of-thought leakage.

2. **Clarification guard** — `isClarifyingStage4Response()` detects when the AI asks for clarification (question mark + trigger phrases) and resets `stage4WalkthroughAction` to NONE / clears proposals. Prevents premature state capture when the user's intent is ambiguous.

3. **Soft/hard streaming timeout** — the old 15s timeout was closing the SSE connection, which killed responses that were still being generated. Now: 15s soft-recovers (refetch persisted messages, keep connection open) and 90s hard-closes. Solid approach.

4. **Stage 4 UI input gating** — freeform input now only shows during active needs walkthrough phases (MY_NEEDS/PARTNER_NEEDS with a currentNeed). Drawer entry point stays reachable regardless.

### Notes

- The thinking buffer can now grow unbounded if a closing tag never arrives — acceptable since streams are finite and the stream-end handler catches this case.
- `isClarifyingStage4Response` is heuristic-based — a question that doesn't match the regex won't be caught. Worth monitoring whether false negatives happen in practice, but conservative (false negatives just preserve current behavior).
- Good test coverage for both the buffer trap and clarification guard scenarios.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Guards Stage 4 streaming by: (1) keeping hidden thinking buffers trapped instead of discarding them mid-stream, (2) preventing Stage 4 state capture when the visible response is a clarification question, (3) splitting the mobile streaming timeout into soft (15s refetch) and hard (90s close) recovery, and (4) adjusting input visibility to only show during active needs walkthrough.

## Feedback

**[Correctness — minor]** `isClarifyingStage4Response` (`messages.ts:588-593`) relies on the response ending with `?` *and* matching a keyword list. This is a reasonable heuristic, but edge cases exist — e.g., an AI response like "Would you like to explore that further? Let me know." would not match because it doesn't end with `?`. Given the guard only suppresses premature state capture (not user-visible behavior), this is acceptable, but worth noting if false negatives surface in production.

**[Security — good]** The thinking buffer change (keeping it trapped rather than discarding at >2000 chars) is the right call. The old behavior risked a race where the buffer would be discarded and subsequent text (which might still be hidden reasoning) would flow to the client as visible content. The new approach is strictly safer.

**[Architecture — good]** The soft/hard timeout split in `useStreamingMessage.ts` is a solid improvement. Stage 4 model calls can be slow (tool use + streaming), and the old 15s hard close was causing lost responses. The 15s soft recovery + 90s hard timeout is a pragmatic solution.

**[Tests — good]** Both new test cases are well-constructed:
- The hidden reasoning test (`messages.test.ts:628`) correctly simulates a >2000 char thinking buffer and verifies it doesn't leak.
- The clarification guard test (`messages.test.ts:813`) covers the full flow including `getModelCompletionWithTools` mock and SSE output assertions.

**[Performance — minor]** The Stage 4 cache invalidation (`useStreamingMessage.ts:582-594`) invalidates 4 query keys whenever any stage4 metadata is present. This is fine for correctness, but if Stage 4 produces many metadata events per turn, it could trigger redundant refetches. Not blocking since React Query deduplicates in-flight requests.

**[Prompt — good]** The additions to `stage-prompts.ts` (lines 1065-1070) add clear phase-specific guidance and the "walkthrough can only reach QUALITY_REVIEW once each need has been covered or skipped" constraint is important for preventing premature completion signals.

## Verdict
Approve — well-scoped defensive changes that address real production issues (leaked reasoning, premature state capture, lost responses from slow streams). Tests cover the key scenarios. No blocking concerns.

PR #645, opened 2026-05-24
Fix duplicate needs capture in Stage 3

slam-paws:
## Summary

- **Bug 1 fix**: `captureSingleNeedForUser` now deletes existing unconfirmed needs before creating a new one, preventing duplicates from accumulating when the AI proposes on consecutive turns
- **Bug 2 fix**: Stage 3 dynamic prompt now injects the current captured needs list so the AI knows what's already been captured and avoids re-proposing
- Adds `capturedNeeds` field to `PromptContext` and fetches it in parallel during message streaming

## Provenance

- Channel: DM (Shantam)
- Requester: Shantam (U0AD3TF2U7L)
- Context: Duplicate needs appeared in session `cmpc3ra1g0097nw1yu40bb1q6` — three needs captured across three consecutive turns despite user confirming first was enough

## Test plan

- [ ] Verify type-checking passes (`npm run check`)
- [ ] Verify existing needs tests pass
- [ ] Manual test: in Stage 3, confirm a single need → verify AI doesn't re-propose on the next turn
- [ ] Manual test: confirm that confirmed needs are preserved (only unconfirmed are replaced)

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -1783,7 +1783,7 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
 
     // Assemble full context including notable facts from UserVessel
     // Also fetch latest emotional reading for intensity-dependent prompt behavior
-    const [contextBundle, sharedContentHistory, milestoneContext, emotionalIntensity] = await Promise.all([
+    const [contextBundle, sharedContentHistory, milestoneContext, emotionalIntensity, capturedNeeds] = await Promise.all([
       assembleContextBundle(
         sessionId,
         user.id,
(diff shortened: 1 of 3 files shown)


slam-paws:
## Bot review

**LGTM** ✅ — solid two-pronged fix for the duplicate-needs accumulation bug.

### What I checked

**`needs.ts` — `deleteMany` before create**
The `confirmed: false` filter is the key safety guard. Confirmed needs are never touched; only the transient unconfirmed draft is replaced. In Stage 3's one-need-at-a-time flow this is the right call — the old unconfirmed proposal is superseded by the new one. No regression risk for confirmed needs.

**`messages.ts` — captured needs prefetch**
Two sequential DB queries (vessel → needs) inside the Stage 3 conditional IIFE. Works correctly. Could be a single query via `findFirst` with `include: { identifiedNeed: ... }` but not a bug — the perf delta is negligible here.

**`stage-prompts.ts` — prompt injection**
Goes into `dynamicParts`, so no prompt-cache invalidation. The phrasing is clear: lists needs by index with confirmed/draft status and blocks re-proposals. This is the root-cause fix.

**Doc Impact Check**
The flagged docs (`stage-1.md`, `stage-1-witness.md`) are stale mappings — this is Stage 3 code. `prompting-architecture.md` doesn't need updating for a dynamic-section addition. No doc changes needed.

### Minor note

`captureSingleNeedForUser` has no unit test in the codebase, so the new dedup behavior is untested. Not blocking for a prod bug fix, but worth a follow-up test (verify: two consecutive calls leave exactly one unconfirmed need, confirmed needs survive).

slam-paws:

Issue #648, opened 2026-05-24
Internal ID shown in proposal 'Addresses' field instead of need label


slam-paws:
## Bug Report

The "Addresses" field on shared proposal cards displays a raw internal ID (e.g., `cmpk4efu200kpp81yrr5q4bye`) instead of the human-readable need label.

<img src="https://github.com/shantamg/meet-without-fear/blob/temp-image-upload/.github/images/issue-id-shown-in-proposal.png?raw=true" alt="Screenshot showing internal ID in Addresses field" width="300" />

### Steps to reproduce
1. Progress through a session to Stage 4 (proposals)
2. View a shared proposal card
3. The "Addresses:" line shows a database ID instead of a label like "Partnership" or "Understanding"

## What We Know So Far

- The mobile component (`Stage4RedesignPanel.tsx:363-370`) correctly renders `proposal.needsAddressed[].label` — the display code is fine
- The fallback path in `stage4-state.ts:276-279` uses `proposal.needsAddressed` strings directly as labels. When the AI stores need IDs instead of labels, those IDs pass through unresolved
- The AI tool schema (`stage-tools.ts:100`) defines `needsAddressed` as `string[]` with no guidance on whether to use labels or IDs — the AI sometimes returns IDs

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Shantam
- **Original message**: It looks like there's an id or something being shown to the user.
- **Timestamp**: 1779656754.046079

slam-paws:
## Investigation Report

**Root cause**: The AI sometimes fills `needsAddressed` with internal CUID IDs (e.g. `cmpk4efu200kpp81yrr5q4bye`) instead of human-readable labels. Two things conspire:

1. **Ambiguous AI context**: The prompt formats needs as `id=<cuid> | status=... | label="<label>"` (`messages.ts:156`), giving the AI both IDs and labels with no guidance on which to use for proposals.
2. **No ID-based matching on write**: `linkProposalToIdentifiedNeeds` (`stage4-capture.service.ts:70-108`) only matches via `needLabelMatches` (fuzzy label comparison). A CUID string never matches a label like "Partnership", so no `StrategyProposalNeed` link rows are created.
3. **Fallback displays raw strings**: `buildProposalCard` (`stage4-state.ts:276`) wraps whatever is in the `needsAddressed` column as `{ label: <raw_string> }` when no links exist — so the CUID passes straight through to the mobile UI.

**Evidence**:
[...]

PR #651, opened 2026-05-24
Fix internal ID shown in proposal Addresses field

slam-paws:
## Changes
- Fix: Add description to AI tool schema for `needsAddressed` instructing use of human-readable labels, not IDs
- Fix: `linkProposalToIdentifiedNeeds` now matches by need ID in addition to fuzzy label matching, so junction records are created even when AI passes IDs
- Fix: `buildProposalCard` fallback resolves ID-like strings to labels via coverage row lookup, fixing display for existing data

Related to #648

## Provenance
- **Requested by:** Shantam
- **Original message:** It looks like there's an id or something being shown to the user.

## Docs updated
No doc changes needed — backend-only bug fix with no doc mappings for affected files.

### DIFF
--- a/backend/src/services/stage-tools.ts
+++ b/backend/src/services/stage-tools.ts
@@ -97,7 +97,7 @@ Call this tool before your visible conversational response whenever the latest t
             description: { type: 'string' },
             kind: { type: ['string', 'null'], enum: ['SHARED_PROPOSAL', 'INDIVIDUAL_COMMITMENT', null] },
             ownerUserId: { type: ['string', 'null'] },
-            needsAddressed: { type: 'array', items: { type: 'string' } },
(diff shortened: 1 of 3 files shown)


slam-paws:
Plan-alignment review: **ALIGNED**

Implementation matches requirements from #648:
- [x] AI tool schema description added to `needsAddressed` instructing use of labels, not IDs (`stage-tools.ts`)
- [x] Direct ID matching in `linkProposalToIdentifiedNeeds` — write path now creates links even when AI passes CUIDs (`stage4-capture.service.ts`)
- [x] Fallback display resolves IDs to labels via coverage row lookup — fixes existing stored data (`stage4-state.ts`)

No scope deviations detected. Three-pronged fix covers prevention (schema), write-path (link creation), and read-path (display fallback). Ready for verification.

PR #653, opened 2026-05-25
[codex] Fix empty Stage 4 review state

shantamg:
## Summary

- Handle the empty Stage 4 `QUALITY_REVIEW` state as a brainstorming recovery state instead of an agreement review.
- Add a `Back to chat` affordance for the Stage 4 drawer when no proposals have been captured.
- Add regression coverage for the empty review state.

## Why

Stephanie hit a drawer that said `Review agreements` even though there were no proposals and no composer in the drawer. That made the app feel like it wanted input but offered nowhere to type.

## Root Cause

The Stage 4 panel treated `walkthrough.phase === QUALITY_REVIEW` as agreement review unconditionally. When `allProposals.length === 0`, it still rendered `Options to consider` / `No proposals captured yet`, and the drawer footer only said proposals would appear later.

## Validation

- `npm --prefix mobile test -- Stage4RedesignPanel.test.tsx --runInBand`
- `npm --prefix mobile run check`
- `git diff --check`

### DIFF
--- a/mobile/src/components/Stage4RedesignPanel.tsx
+++ b/mobile/src/components/Stage4RedesignPanel.tsx
@@ -33,6 +33,7 @@ interface Stage4RedesignPanelProps {
   onUndeclineNeed?: (needId: string) => void;
   onMarkNeedCovered?: (needId: string) => void;
   onSkipNeed?: (needId: string) => void;
+  onBackToChat?: () => void;
   onCloseStage4: (kind: Stage4ClosureKind, reason: Stage4ClosureReason, checkInDate: string) => void;
 }
 
@@ -45,6 +46,7 @@ interface Stage4RedesignFooterProps {
   onShareSelections?: () => void;
   onReviseSelections?: () => void;
   onKeepRefiningNoOverlap?: () => void;
+  onBackToChat?: () => void;
   onCloseStage4: (kind: Stage4ClosureKind, reason: Stage4ClosureReason, checkInDate: string) => void;
 }
 
@@ -535,6 +537,7 @@ export function Stage4RedesignFooter({
   onReviseSelections,
   onCloseStage4,
   onKeepRefiningNoOverlap,
+  onBackToChat,
 }: Stage4RedesignFooterProps) {
   const { palette } = useAppAppearance();
   const styles = useMemo(() => makeStyles(palette), [palette]);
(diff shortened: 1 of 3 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Handles the empty `QUALITY_REVIEW` state in Stage 4 by showing a "Keep brainstorming" recovery view instead of the misleading "Review agreements" screen. Adds a "Back to chat" button to dismiss the drawer.

## Feedback

- **Good pattern**: The early return for `allProposals.length === 0` inside the `QUALITY_REVIEW` branch (`Stage4RedesignPanel.tsx:845`) cleanly separates the empty state from the review state, avoiding the confusing "Options to consider / No proposals captured yet" dead end.
- **Good pattern**: The `GuidedActionPanel` in `UnifiedSessionScreen.tsx:3648` now contextualizes its copy based on `isEmptyQualityReview`, giving the user appropriate guidance rather than "Start with the first need" when there are no proposals.
- **Tests**: Good regression test covering both the presence of the recovery copy and the absence of the agreement review copy, plus verifying the `onBackToChat` callback fires.
- **Minor observation**: The existing `QUALITY_REVIEW` branch with proposals (line ~863) doesn't pass `onBackToChat` to the footer — that's correct since the button only makes sense when there are zero proposals.

## Verdict

Approve — well-scoped fix with clear regression coverage. The root cause is correctly identified and the fix addresses both the drawer panel and the chat surface CTA.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes the empty Stage 4 `QUALITY_REVIEW` state so it shows a "Keep brainstorming" recovery UI instead of a misleading "Review agreements" panel when no proposals exist.

## Feedback

- **Good pattern**: The early return in `Stage4RedesignPanel.tsx:845` for `allProposals.length === 0` within the `QUALITY_REVIEW` phase is the right approach — it intercepts the empty state before the existing review UI renders, keeping the existing code path untouched.
- **Good UX copy**: The guidance text ("Close this and keep talking in the chat. When something feels workable, it will appear here for review.") sets clear expectations without jargon.
- **Test coverage**: The regression test verifies both positive assertions (correct text) and negative assertions (`queryByText` for text that should *not* appear), which is solid.
- **Minor note**: The `onBackToChat` prop is optional in both `Stage4RedesignPanelProps` and `Stage4RedesignFooterProps`, which is correct since not all callers need it. The conditional rendering with `{onBackToChat && (...)}` handles this cleanly.
- **CTA adjustment** in `UnifiedSessionScreen.tsx:3648-3666`: Nice touch adapting the inline CTA card copy for the empty quality review state — "Open guidance" instead of "Open Stage 4" makes the action clearer.

## Verdict

Approve — clean fix with good test coverage. No issues found.

PR #655, opened 2026-05-25
[codex] Add Stage 4 walkthrough reset recovery

### LINKED ISSUE #654: Add Stage 4 walkthrough reset/unskip recovery
## Problem

A user can reach empty Stage 4 `QUALITY_REVIEW` with no proposals captured. PR #653 makes that state recoverable by sending the user back to chat, but there is still no structured way to return to a specific need if the walkthrough reached review because needs were skipped or advanced too aggressively.

## Desired behavior

Provide a backend-supported recovery action that can move the Stage 4 walkthrough back to an own or partner need, clearing the relevant skipped/covered marker as needed.

## Notes

Current endpoint only supports marking the current need as `covered` or `skip`:

`POST /sessions/:id/stage4/walkthrough/needs/:needId`

Once `currentNeed` is null in `QUALITY_REVIEW`, the client has no clean action to reset/unskip a need.

## Acceptance criteria

- Backend exposes a safe reset/unskip action for Stage 4 walkthrough needs.
- Client can offer a secondary recovery CTA from empty review state, e.g. `Go back to a need`.
- Recovery preserves existing proposals and does not fake state client-side.
- Tests cover resetting from empty `QUALITY_REVIEW` back to a need.

shantamg:
## Summary

- Add a backend-supported `reset` action to the Stage 4 walkthrough need endpoint.
- Clear covered/skipped walkthrough markers for the selected need and restore the correct own/partner walkthrough phase from server state.
- Add an empty-review recovery UI so users can go back to a reviewed need instead of only returning to chat.

## Why

Issue #654 identified a recoverable but awkward state: Stage 4 can reach `QUALITY_REVIEW` with no proposals and no `currentNeed`, leaving the client without a structured backend action to unskip/reset a need.

Closes #654.

## Validation

- `cd backend && npm test -- --runTestsByPath src/routes/__tests__/stage4.test.ts --runInBand`
- `cd backend && npm run check`
- `cd mobile && npm test -- --runTestsByPath src/components/__tests__/Stage4RedesignPanel.test.tsx --runInBand`
- `cd mobile && npm run check`

### DIFF
--- a/backend/src/controllers/stage4.ts
+++ b/backend/src/controllers/stage4.ts
@@ -243,7 +243,7 @@ const submitStage4SelectionsRequestSchema = z.object({
 });
 
 const stage4NeedWalkthroughActionSchema = z.object({
-  action: z.enum(['covered', 'skip']),
+  action: z.enum(['covered', 'skip', 'reset']),
 });
 
 const requestSuggestionsRequestSchema = z.object({
@@ -303,6 +303,13 @@ class Stage4ClosedConflictError extends Error {
   constructor() {
     super('Stage 4 is already closed');
     this.name = 'Stage4ClosedConflictError';
+  }
+}
+
+class Stage4WalkthroughNeedNotFoundError extends Error {
+  constructor() {
+    super('Stage 4 walkthrough need not found');
+    this.name = 'Stage4WalkthroughNeedNotFoundError';
   }
 }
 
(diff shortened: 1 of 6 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Adds a `reset` action to the Stage 4 walkthrough need endpoint, allowing users to un-skip/un-cover a need and return to it. Also adds an empty-review recovery UI showing previously reviewed needs with reset buttons.

## Feedback
- **Good**: The phase logic refactor from nested ternaries to explicit if/else is much more readable.
- **Good**: Stage guard (`mutable.progress?.stage !== 4`) prevents misuse from wrong stages — nice hardening.
- **Good**: The `Stage4WalkthroughNeedNotFoundError` with proper 404 handling is clean error boundary work.
- **Minor (non-blocking)**: `backend/src/controllers/stage4.ts:425` — the `targetOwnNeed` / `targetPartnerNeed` lookup happens before the DB call for `stageProgress`, so if the need ID is valid but the user has no stage progress record, you'd get a less specific error ("Stage 4 walkthrough state not found") rather than the 404. Not a problem in practice since the stage guard above catches non-stage-4 users first.
- **Style (non-blocking)**: Trailing comma formatting changes (e.g., lines 433, 437, 449, 453) are fine but unrelated to the feature — presumably auto-formatter. No issue.

## Verdict
Approve — clean implementation with good test coverage for the new reset path. The backend validates correctly and the mobile UI conditionally renders only when `onResetNeed` is provided.

PR #652, opened 2026-05-24
Exclude partial needs from Stage 4 sharing gate

slam-paws:
## Changes
- Fix: Remove PARTIAL coverage rows from the `allOpenNeedsAddressedOrDeclined()` gate check — only OPEN needs now block sharing
- The gate previously included both OPEN and PARTIAL rows, but only OPEN needs have user-actionable controls (brainstorm in chat / leave for now). PARTIAL needs blocked sharing with no escape hatch.
- The CTA subtitle text ("brainstorm or set them aside first") is accurate for OPEN needs, so no text change needed.

Related to #649

## Provenance
- **Channel:** #bugs-and-requests
- **Requested by:** Shantam
- **Original message:** "It says there are needs still unaddressed but that doesn't seem true. I think the brainstorm is not even an option anymore."

## Docs updated
No doc changes needed — one-line logic fix in mobile component.

Issue #649, opened 2026-05-24
CTA subtitle shows stale 'needs unaddressed' gate text


slam-paws:
## Bug Report

The CTA footer in Stage 4 shows the text: "There are still needs unaddressed — brainstorm or set them aside first." even when the user believes all needs have been addressed. The gate logic in `allOpenNeedsAddressedOrDeclined()` may be evaluating incorrectly, or the coverage audit data may be stale.

Two sub-issues:
• The gate may be falsely blocking — reporting needs as unaddressed when they are actually covered
• The text mentions "brainstorm" as an action — need to verify this is still accurate for the current flow

Appears at two locations in `mobile/src/components/Stage4RedesignPanel.tsx` (lines 616, 999), gated by `allMyDecisionsMade && !needsGated`.

## What We Know So Far

• `allOpenNeedsAddressedOrDeclined()` checks `coverageAudit.open` and `coverageAudit.partial` rows — each row must either have `userDeclinedToAddress === true` or a covering proposal marked WILLING
• "Brainstorm in chat" buttons do exist in the code, so brainstorm is still a feature — but the gate text should be verified against current product state
• The raw need ID (`cmpk4iod800ljp81y0xrmb6xk`) is also leaking into the "Addresses:" line in the same screen (separate issue)

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Shantam
- **Original message**: "It says there are needs still unaddressed but that doesn't seem true. I think the brainstorm is not even an option anymore. Can you see if that text at the bottom is out of date or wrong?"
- **Timestamp**: 1779656857.975389

slam-paws:
## Investigation Report

**Root cause**: The gate function `allOpenNeedsAddressedOrDeclined()` in `Stage4RedesignPanel.tsx:62` includes **both** `open` AND `partial` coverage rows in its check:

```typescript
const rows = [...safeState.coverageAudit.open, ...safeState.coverageAudit.partial];
```

However, the brainstorm and "Leave for now" (decline) buttons only appear for **OPEN** needs (`NeedRows` line 470: `showActions = tone === 'open'`). **PARTIAL needs block the gate but provide no user-actionable controls** — the user literally cannot "brainstorm or set them aside" as the text instructs.

[...]

PR #656, opened 2026-05-25
Fix Stage 4 backend share gate

shantamg:
## Changes
- Align backend Stage 4 share validation with the frontend rule from #652: only OPEN needs block sharing.
- Keep PARTIAL coverage rows from creating a second backend-only dead end after the Share button is enabled.
- Add a regression test proving partial-only coverage can share while open needs still block.

## Verification
- npm --workspace backend test -- --runTestsByPath src/routes/__tests__/stage4.test.ts --runInBand

Related to #649

### DIFF
--- a/backend/src/controllers/stage4.ts
+++ b/backend/src/controllers/stage4.ts
@@ -917,13 +917,13 @@ export async function shareStage4Selections(req: Request, res: Response): Promis
       return;
     }
 
-    // Gate: every open/partial need must be addressed by a willing-active
+    // Gate: every open need must be addressed by a willing-active
     // proposal or explicitly declined ("leave for now") by this user.
-    const [openOrPartialNeeds, willingSelections, declinations] = await Promise.all([
+    const [openNeeds, willingSelections, declinations] = await Promise.all([
       prisma.stage4NeedCoverage.findMany({
         where: {
           sessionId,
-          coverageStatus: { in: ['OPEN', 'PARTIAL'] },
+          coverageStatus: 'OPEN',
         },
         select: {
           id: true,
@@ -957,7 +957,7 @@ export async function shareStage4Selections(req: Request, res: Response): Promis
     );
     const declinedNeedIds = new Set(declinations.map((d) => d.needId));
 
-    const ungated = openOrPartialNeeds.filter((row) => {
+    const ungated = openNeeds.filter((row) => {
       const needIdentifier = row.needId ?? row.id;
       if (declinedNeedIds.has(needIdentifier)) return false;
       const hasWillingCover = row.coveringProposalIds.some(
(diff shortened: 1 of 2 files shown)


shantamg:
@slam-paws please review this PR. It aligns the backend Stage 4 share gate with the already-merged frontend gate from #652 and includes targeted regression coverage.

slam-paws:
Looking into this — reviewing now.

slam-paws (approved):
Looks good — clean, well-scoped change that correctly aligns the backend share gate with the frontend.

**What I verified:**
- The frontend gate (`Stage4RedesignPanel.tsx:65`) already only checks `coverageAudit.open` (OPEN needs) — this backend change now matches that by querying `coverageStatus: 'OPEN'` instead of `{ in: ['OPEN', 'PARTIAL'] }`.
- Variable rename `openOrPartialNeeds` → `openNeeds` and comment update are consistent.
- The regression test covers the right scenario and asserts the query parameter change.

**Minor observation (non-blocking):** The new test arranges a PARTIAL row with willing+active coverage on it, so it would pass the `ungated` filter even under the old code. The `toHaveBeenCalledWith` assertion on the query parameter is what actually validates the behavior change, which is fine — just noting it for clarity.

LGTM ✅

slam-paws:
## Bot review — PR #656

**Related to**: #649
**Verdict**: LGTM ✅

### Checklist

**Correctness** ✅
- Aligns the backend share gate with the frontend behavior established in #652: only `OPEN` needs block sharing, not `PARTIAL`
- `Stage4RedesignPanel.tsx` already checks `coverageAudit.open` (OPEN only) — this makes the backend match
- Logic is straightforward: single `coverageStatus: 'OPEN'` query param replaces the `{ in: ['OPEN', 'PARTIAL'] }` filter
- Variable rename `openOrPartialNeeds` → `openNeeds` and comment update are consistent with the intent

**Completeness** ✅
- All three touch points updated together: query param, variable name, comment
- PR body correctly describes the change and its relationship to #649

**Conventions** ✅
- Follows existing patterns in stage4.ts; no unnecessary refactoring

**Security** ✅
- No secrets, no injection surface, no auth bypass

**Tests** ✅
- New regression test `allows share when remaining coverage rows are only partial` covers the key scenario
- Test assertion `toHaveBeenCalledWith(expect.objectContaining({ where: expect.objectContaining({ coverageStatus: 'OPEN' }) }))` directly validates the query parameter change — this is the right approach
- Note: the test arrangement (PARTIAL row with willing+active coverage) would also pass the `ungated` filter under the old code, so the behavioral assertion (`status 200`) doesn't distinguish old vs new; the query-param assertion is what counts. This is fine and consistent with how mock-based tests validate DB layer changes.

**Doc impact** ✅
- No doc update needed — the share gate's internal query logic isn't documented at the API spec level; the public API contract (POST /share-selections) is unchanged

slam-paws:

PR #657, opened 2026-05-25
fix: resolve partner name on existing-person invitation path

shantamg:
## Changes
- Fix: existing-person invitation path now resolves and stores the selected person's display name on the invitation (`invitations.ts`)
- Fix: recover partner name from relationship data for older sessions where `invitation.name` is blank (`messages.ts`)
- Fix: honor `partnerId` when starting a session from a person detail screen (`new.tsx`)
- Add: backend regression test for the existing-person name resolution (`invitations.test.ts`)

**Root cause:** the existing-person path only sent `personId`, so the backend stored `invitation.name` as `undefined` and the first-message AI prompt saw "someone" instead of the partner's name. The session header still showed the correct name from relationship data, which masked the bug.

## Provenance
- **Original message:** Work performed by Codex at the user's request; user asked Claude to open a PR for the resulting changes.
- **Prompt(s) used:** `/pr` skill

## Docs updated
- No doc changes needed (bug fix; no mapped docs match the changed files)

### DIFF
--- a/backend/src/controllers/invitations.ts
+++ b/backend/src/controllers/invitations.ts
@@ -315,6 +315,12 @@ export async function createSession(req: Request, res: Response): Promise<void> 
     }
 
     const { personId, inviteName, innerThoughtsId, linkedAtMessageId, context } = parseResult.data;
+    let inviteeDisplayName = inviteName?.trim() || undefined;
+
+    if (personId === user.id) {
+      errorResponse(res, 'VALIDATION_ERROR', 'Cannot start a session with yourself', 400);
+      return;
+    }
 
     // Create or find relationship
     let relationship;
@@ -329,11 +335,51 @@ export async function createSession(req: Request, res: Response): Promise<void> 
             },
           },
         },
-        include: { relationship: true },
+        include: {
+          user: {
+            select: {
+              firstName: true,
+              name: true,
+            },
+          },
+          relationship: {
+            include: {
+              members: {
+                where: { userId: user.id },
+                select: { nickname: true },
+              },
+            },
+          },
+        },
       });
 
       if (existingMember) {
         relationship = existingMember.relationship;
+        inviteeDisplayName =
+          inviteeDisplayName ||
+          existingMember.relationship.members[0]?.nickname ||
+          existingMember.user.firstName ||
+          existingMember.user.name ||
+          undefined;
+      } else {
+        const existingUser = await prisma.user.findUnique({
+          where: { id: personId },
+          select: {
+            firstName: true,
+            name: true,
+          },
+        });
+
+        if (!existingUser) {
+          errorResponse(res, 'NOT_FOUND', 'Person not found', 404);
+          return;
+        }
+
+        inviteeDisplayName =
+          inviteeDisplayName ||
+          existingUser.firstName ||
+          existingUser.name ||
+          undefined;
       }
     }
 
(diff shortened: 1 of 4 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the existing-person invitation path so the partner's display name is properly resolved and stored on the invitation, preventing the AI from seeing "someone" instead of the actual name. Also adds a fallback in `getInitialMessage` for older sessions with blank invitation names, and pre-selects the person on the mobile new-session screen when launched from a person detail view.

## Feedback

- **Correctness (good)**: The name resolution priority chain (`inviteName > nickname > firstName > name`) is sensible and consistent across both the creation and recovery paths.

- **Correctness (good)**: The self-session guard (`personId === user.id`) is a nice defensive addition.

- **Safety (good)**: The dual-membership creation when `personId` is provided and no existing relationship exists is safe — `acceptInvitation` already guards against duplicates with `findUnique` before creating (invitations.ts:638-654).

- **Performance (minor)**: The recovery path in `messages.ts:1306-1335` makes two sequential Prisma queries (`findUnique` + `findFirst`). These could be a single query joining both membership records, but since this is a backward-compat fallback for older sessions, it's fine as-is.

- **Test coverage (good)**: The regression test covers the happy path for existing-person name resolution and asserts both the invitation name and the response partner object. Nice.

- **Mobile (good)**: The `partnerId` effect in `new.tsx` correctly guards against re-selection with `!selectedPerson` and waits for the people list to load.

## Verdict
**Approve** — Clean, well-scoped bug fix with good test coverage. The root cause analysis in the PR description is clear and the fix addresses both new sessions and older sessions missing the name.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the existing-person invitation path so the partner's display name is resolved and stored on the invitation, preventing the AI prompt from seeing "someone" instead of the actual name. Also adds a recovery path in `messages.ts` for older sessions with blank `invitation.name`, and wires up `partnerId` in the mobile new-session screen.

## Feedback

- **Correctness (good)**: The name resolution priority chain (`inviteName > nickname > firstName > name`) is sensible and consistent across both the creation path (`invitations.ts`) and the recovery path (`messages.ts`).

- **Correctness (good)**: The self-session guard (`personId === user.id`) is a nice defensive addition at `invitations.ts:319`.

- **Architecture (good)**: Creating both relationship members upfront when `personId` is known (`invitations.ts:391-401`) is the right call — it avoids the dangling single-member relationship that caused the original bug.

- **Performance (minor nit, non-blocking)**: In `messages.ts:1304-1332`, the two sequential queries (`findUnique` for inviter membership + `findFirst` for partner member) could be parallelized with `Promise.all` since they're independent. Not blocking since this is a fallback path for older sessions only.

- **Tests (good)**: The regression test in `invitations.test.ts` covers the core fix well — verifying that `invitation.name` is set to `'Darryl'` and the response includes the correct `partner.id` and `partner.name`.

- **Mobile (good)**: The `useEffect` in `new.tsx` correctly gates on `!selectedPerson` to avoid overwriting a user's manual selection, and the dependency array is appropriate.

## Verdict
**Approve** — Clean, well-scoped bug fix with a clear root cause explanation. The only suggestion is the minor `Promise.all` optimization in the recovery path, which is non-blocking.

PR #626, opened 2026-05-18
fix: move WebSocket auth to first-message pattern

slam-paws:
## Changes
- Fix: Removed JWT token from WebSocket upgrade URL query param (leaked to server/proxy logs)
- Add: First-message auth pattern — client sends `{ type: "auth", token }` as first WebSocket message
- Add: 5-second auth timeout — connections that don't authenticate are closed with `1008 Policy Violation`
- Add: 9 tests covering auth success, rejection (bad token, missing token, non-auth message, invalid JSON), and timeout behavior

Related to #620

## Provenance
- **Channel:** #security-audit
- **Requested by:** automated security audit (weekly)
- **Original message:** Weekly security audit 2026-05-18
- **Prompt(s) used:** 7-domain parallel scan — asset inventory agent

## Docs updated
No doc changes needed — `realtime-transcription.ts` has no entry in `docs/code-to-docs-mapping.json` and no corresponding living doc. The file's own docblock was updated to describe the new auth flow.

PR #547, opened 2026-05-11
fix: enforce FIELD_ENCRYPTION_KEY in production

slam-paws:
## Changes
- Add: fail-fast startup guard in `server.ts` — throws if `FIELD_ENCRYPTION_KEY` is missing in production (matches existing `E2E_AUTH_BYPASS` pattern)
- Fix: `getKey()` in `field-encryption.ts` now throws in production instead of silently degrading to plaintext passthrough
- Add: production enforcement tests (encrypt/decrypt throw when key missing in production)
- Update: `docs/deployment/environment-variables.md` — `FIELD_ENCRYPTION_KEY` listed as required production variable
- Update: `docs/backend/security/index.md` — reflects mandatory key requirement in production

Related to #543

## Provenance
- **Channel:** #security-audit
- **Requested by:** Automated weekly security audit
- **Original message:** Weekly security audit cron — 2026-05-11
- **Prompt(s) used:** general-pr workspace

## Docs updated
- `docs/deployment/environment-variables.md` — added FIELD_ENCRYPTION_KEY to Required table and Production section
- `docs/backend/security/index.md` — updated encryption bullet to reflect mandatory key in production

## Note
This PR only addresses the code-hardening portion of #543. The following require separate work:
- **PII field expansion** (email, firstName, etc.) needs architectural design for blind indexing since `User.email` has a `@unique` index
- **Backfill migration** to encrypt existing plaintext data
- **Verify key is set in Render** — requires manual dashboard access

PR #658, opened 2026-05-25
fix: make missing FIELD_ENCRYPTION_KEY non-fatal in production

shantamg:
## Changes
- Fix: production no longer crashes on boot when `FIELD_ENCRYPTION_KEY` is unset — logs a warning instead
- Add: opt-in `REQUIRE_FIELD_ENCRYPTION=true` to restore fail-fast enforcement before launch

## Why
PR #547 made a missing `FIELD_ENCRYPTION_KEY` a fatal startup error in production. We intentionally run **keyless / plaintext** during the pre-launch test phase so conversation content is directly inspectable in the DB for prompt debugging.

That enforcement has been **crashing every backend deploy since #655** (build succeeds, server exits 1 at startup), so production has been frozen on commit `066c057b` and **five merged PRs never deployed** — including the Stage 3 duplicate-needs fix (#645), Stage 4 share-gate (#656), WebSocket auth (#626), and the partner-name fix (#657).

The encryption middleware already passes all fields through unchanged when no key is set (`prisma-encryption-middleware.ts`), so keyless operation is a supported, deliberate mode — not an accident worth crashing over.

This change self-heals the deploy: the moment it merges, the server boots without a key again and the backlog of merged PRs ships.

## Tradeoff
Plaintext-at-rest for sensitive content. Acceptable for the test phase; set `FIELD_ENCRYPTION_KEY` (and `REQUIRE_FIELD_ENCRYPTION=true`) before real users.

## Provenance
- **Original message:** While debugging why a new session's Stage 0 opener still says "someone" instead of the partner's name, traced it to failed backend deploys; user confirmed they deliberately want plaintext in the test phase for debuggability.
- **Prompt(s) used:** manual investigation (Render API deploy logs) + `/pr`

## Docs updated
- No doc changes needed

### DIFF
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -61,12 +61,25 @@ if (process.env.E2E_AUTH_BYPASS === 'true' && process.env.NODE_ENV === 'producti
   );
 }
 
-// Fail fast: FIELD_ENCRYPTION_KEY is required in production
+// Field encryption in production.
+// During the pre-launch test phase we deliberately run without a key so that
+// conversation content stays plaintext and can be inspected directly in the DB
+// for prompt debugging. When FIELD_ENCRYPTION_KEY is unset the encryption
+// middleware passes all fields through unchanged (see prisma-encryption-middleware.ts).
+// Set REQUIRE_FIELD_ENCRYPTION=true to re-enable fail-fast enforcement before launch.
 if (!process.env.FIELD_ENCRYPTION_KEY && process.env.NODE_ENV === 'production') {
-  throw new Error(
-    'FATAL: FIELD_ENCRYPTION_KEY is not set in a production environment. ' +
-    'All sensitive fields (messages, empathy drafts, conversation summaries) would be stored as plaintext. ' +
-    'Generate a key with: openssl rand -base64 32'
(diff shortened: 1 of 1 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Converts the missing `FIELD_ENCRYPTION_KEY` startup check from a hard crash to a warning, with an opt-in `REQUIRE_FIELD_ENCRYPTION=true` flag to restore fail-fast behavior before launch.

## Feedback
- **Correctness**: Logic is sound — the nested `REQUIRE_FIELD_ENCRYPTION` check correctly preserves the ability to enforce encryption when needed, while defaulting to a warn-only path during the test phase.
- **Security**: The tradeoff (plaintext-at-rest during pre-launch) is explicitly documented in the comment and the PR body. The opt-in enforcement flag is a clean pattern for re-enabling the guard rail later.
- **Operational**: This unblocks five stalled deploys, which is a meaningful prod impact fix.

No issues found. Clean, minimal, well-documented change.

## Verdict
Approve — straightforward, self-contained fix with clear reasoning and an explicit path back to enforcement.

PR #659, opened 2026-05-25
fix: field encryption passes through plaintext when no key (runtime guard)

shantamg:
## Changes
- Fix: `getKey()` in `field-encryption.ts` no longer throws in production when `FIELD_ENCRYPTION_KEY` is unset — it returns `null`, so `encrypt()`/`decrypt()` pass values through as plaintext (their documented keyless behavior)
- Add: throw is now gated behind the same opt-in `REQUIRE_FIELD_ENCRYPTION=true` introduced in #658
- Update: tests for the new passthrough-vs-enforce behavior

## Why (production was broken)
#658 softened the **startup** fail-fast so the server boots keyless. But #547 added a **second** guard — a runtime throw in `getKey()` — that fires on every sensitive-field read/write. So once #658 let the server boot, every conversation operation 500'd:

```
[getInitialMessage]      -> 500   ← new sessions get NO opener →
                                     UI falls back to "Start the Conversation"
[getSessionState]        -> 500
[getConversationHistory] -> 500
```

This is why a brand-new session showed the empty-state placeholder and the typing dots never resolved. It also broke reading existing sessions.

`encrypt()`/`decrypt()` already short-circuit to plaintext when `getKey()` returns `null`, so returning `null` (instead of throwing) restores the intended keyless/plaintext test-phase behavior.

## Provenance
- **Original message:** New session with Darryl showed "Start the Conversation" empty state after typing dots; traced to 500s from the runtime FIELD_ENCRYPTION_KEY guard missed by #658.
- **Prompt(s) used:** Render runtime log analysis + `/pr`

## Docs updated
- No doc changes needed

### DIFF
--- a/backend/src/utils/__tests__/field-encryption.test.ts
+++ b/backend/src/utils/__tests__/field-encryption.test.ts
@@ -149,7 +149,7 @@ describe('field-encryption', () => {
     });
   });
 
-  describe('production enforcement (no key)', () => {
+  describe('production without key (test-phase passthrough)', () => {
     beforeEach(() => {
       _resetForTesting();
       delete process.env.FIELD_ENCRYPTION_KEY;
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes a runtime crash (500) caused by `getKey()` throwing when `FIELD_ENCRYPTION_KEY` is unset in production. #658 allowed the server to boot without the key, but the runtime guard in `getKey()` still threw on every encrypt/decrypt call, breaking all session operations.

## Feedback
- **Correctness**: Logic is sound. `encrypt()`/`decrypt()` already handle `null` key by passing through plaintext (line 64/line ~88), so returning `null` from `getKey()` instead of throwing restores the intended behavior.
- **Consistency**: The opt-in `REQUIRE_FIELD_ENCRYPTION=true` flag cleanly mirrors what #658 introduced for the startup check — good alignment.
- **Tests**: Updated tests cover both the default passthrough path and the opt-in enforcement path. Each test calls `_resetForTesting()` after setting the env var, which correctly clears the cached key. Clean and thorough.
- **Docstring**: Header comment updated to reflect the new behavior — appreciated.

## Verdict
**Approve** — Small, focused fix that resolves a production 500. The passthrough behavior is explicitly documented as a temporary pre-launch test-phase decision, and the enforcement opt-in provides a clean upgrade path when the team is ready to require encryption.

PR #660, opened 2026-05-25
fix: strip leaked tool-call XML from visible AI response

shantamg:
## Changes
- Fix: `parseMicroTagResponse` now strips `<tool_calls>...</tool_calls>` and stray `<invoke>...</invoke>` blocks from the user-facing response text
- Add: regression test for a leaked `update_session_state` tool-call block

## Why
`update_session_state` is a real tool the model should call via the tool-use API (two-pass: state in one pass, prose in another). But the model sometimes emits the tool call as **literal XML text** in the prose pass. The parser stripped every other control tag (`<thinking>`, `<draft>`, `<needs>`, `<stage4_*>`, `<dispatch>`) but not tool-call blocks — so the raw XML rendered on screen above the Stage 0 greeting.

This is a display-layer, defense-in-depth fix: real state still flows through the proper tool-use pass; we just guarantee tool-call markup never reaches the user.

## Provenance
- **Original message:** New Darryl session rendered a raw `<tool_calls><invoke name="update_session_state">...` block above the opener text.
- **Prompt(s) used:** manual investigation + `/pr`

## Docs updated
- No doc changes needed

### DIFF
--- a/backend/src/utils/__tests__/micro-tag-parser.test.ts
+++ b/backend/src/utils/__tests__/micro-tag-parser.test.ts
@@ -246,6 +246,28 @@ Response without closing tag.`;
       expect(result.response).toBeTruthy();
     });
 
+    it('strips a leaked update_session_state tool-call XML block from the response', () => {
+      const raw = `<tool_calls>
+<invoke name="update_session_state">
+<parameter name="mode">ONBOARDING</parameter>
+<parameter name="ready_share">false</parameter>
+<parameter name="draft_topic"></parameter>
+<parameter name="memory_items">[]</parameter>
+</invoke>
+</tool_calls>
+
+Hey Shantam. First, let's write a short topic for Darryl so they know what this is about. What's going on?`;
+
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Strips leaked `<tool_calls>` and `<invoke>` XML blocks from user-facing AI response text, with a regression test covering the exact failure scenario.

## Feedback
- **Correctness**: The two new regexes follow the same `[\s\S]*?` (non-greedy) pattern as every other tag-stripping rule in the parser. Placement at the top of the chain is correct — tool-call blocks should be stripped before any inner tags are processed.
- **Test quality**: Good regression test that asserts both the positive (expected text) and negative (no leaked markup) cases.
- **Defense-in-depth**: Comment clearly explains that the real fix path is the tool-use API; this is a display-layer safety net. Appropriate framing.

## Verdict
**Approve** — Small, focused, well-tested fix for a real user-facing bug. No concerns.

PR #663, opened 2026-05-25
fix: stream reply when model omits the leading <thinking> tag

shantamg:
## Problem

Repeated failed chat submits in production. Sentry showed a cluster of `sendMessageStream` errors (~10 in 5 minutes) on a real session:

```
warn :: Stream ended while still in thinking trap. Buffer has 297 chars. Keeping it hidden.
error :: Empty AI response after tag stripping
error :: Stream failed, cleaning up user message
```

## Root cause

The streaming handler hard-coded `isInsideThinking = true` and buffered all output until it saw `</thinking>`, assuming every response opens with `<thinking>`. When the model skipped the opener and went straight to the visible reply (more often on deep, long-context turns), the entire reply landed in the thinking buffer, the stream-end flush kept it hidden, and the empty-response guard threw — deleting the user message and surfacing as a failed submit.

The 297-char buffer (~75 tokens) vs. the 1536 `maxTokens` cap confirms the stream ended **naturally**, not by truncation — i.e. those chars were the actual reply, hidden by the trap.

## Fix

- **Detect the opener** on the first non-whitespace chunk. If the output can't become a `<thinking>` opener, bail out of the trap and stream the text normally. Ambiguous partial prefixes (`<th`) keep buffering so a genuine opener split across chunks still works.
- **Safer end-of-stream flush**: only hide the buffer when a real `<thinking>` opener was confirmed; otherwise flush it as the visible reply instead of erroring.

## Tests

- New regression test: a no-`<thinking>` stream now streams the reply, emits no empty-response error, keeps the user message, and persists the AI reply.
- Updated existing streaming mocks to include the literal `<thinking>` opener — matching real output shape (all 63 fixtures open with `<thinking>`; `bedrock.ts` does no assistant prefill).

`npm run check` clean; full `messages.test.ts` suite (22 tests) passes.

### DIFF
--- a/backend/src/controllers/messages.ts
+++ b/backend/src/controllers/messages.ts
@@ -2064,9 +2064,14 @@ export async function sendMessageStream(req: Request, res: Response): Promise<vo
     let metadata: SessionStateToolInput = {};
     let streamError: Error | null = null;
 
-    // Tag trap state - Claude outputs <thinking>...</thinking> first, which we hide
-    // After thinking, there may be <draft>, <dispatch>, or <needs> tags that we also hide
+    // Tag trap state - Claude USUALLY outputs <thinking>...</thinking> first, which we hide.
+    // After thinking, there may be <draft>, <dispatch>, or <needs> tags that we also hide.
+    // The model occasionally skips the leading <thinking> and goes straight to visible prose
+    // (more often on deep, long-context turns). We start in the thinking trap optimistically,
+    // but the first chunk decides: if the output does not actually begin with <thinking>, we
+    // bail out of the trap immediately so the reply is not swallowed and hidden.
     let isInsideThinking = true;
+    let sawThinkingOpener = false; // Becomes true only once we confirm a real <thinking> opener
     let isTrappingTags = false; // After thinking, buffer to check for hidden semantic tags
     let thinkingBuffer = '';
     let tagTrapBuffer = ''; // Buffer for checking semantic tags after thinking
@@ -2272,6 +2277,30 @@ Write only the user-facing conversational response. Do not include tool JSON, XM
           if (isInsideThinking) {
             thinkingBuffer += event.text;
 
+            // The model is supposed to open with <thinking>, but it sometimes skips
+            // straight to the visible reply (more often on deep, long-context turns).
+            // Decide as soon as we have a non-whitespace prefix: if it cannot become a
+            // "<thinking>" opener, bail out of the trap so the reply is streamed instead
+            // of being swallowed and hidden (which would surface as an empty-response error).
+            if (!sawThinkingOpener) {
+              const trimmedStart = thinkingBuffer.replace(/^\s+/, '');
+              const opener = '<thinking';
+              if (trimmedStart.length > 0) {
+                if (trimmedStart.startsWith(opener)) {
+                  sawThinkingOpener = true;
+                } else if (!opener.startsWith(trimmedStart.slice(0, opener.length))) {
+                  // Definitely not a <thinking> opener — treat everything as visible output.
+                  logger.warn(`[sendMessageStream:${requestId}] Response did not open with <thinking>; routing ${thinkingBuffer.length} buffered chars as visible response.`);
+                  isInsideThinking = false;
+                  isTrappingTags = true; // Run it through the tag trap so any hidden tags are still caught
+                  tagTrapBuffer = thinkingBuffer;
+                  thinkingBuffer = '';
+                  continue;
+                }
+                // else: still ambiguous (e.g. "<th") — keep buffering.
+              }
+            }
+
             // Check for closing tag
             const closingTagIndex = thinkingBuffer.indexOf('</thinking>');
             if (closingTagIndex !== -1) {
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes a production bug where the streaming handler swallowed the entire AI reply when the model skipped the leading `<thinking>` tag — the thinking trap buffered everything as hidden reasoning and then threw an empty-response error, deleting the user's message.

## Feedback

- **Correctness**: The prefix-matching logic at `messages.ts:2283–2301` is sound. Ambiguous partial prefixes (e.g. `<th`) correctly keep buffering, while non-matching starts bail out immediately. The `continue` after bail-out correctly re-enters the loop into the tag trap phase (PHASE 2), so any semantic tags in non-thinking output are still caught. Clean design.

- **Belt-and-suspenders (end-of-stream)**: The dual-path flush at `messages.ts:2452–2468` is a nice safety net — if PHASE 1 doesn't bail out before stream end (shouldn't happen, but edge cases exist), it still does the right thing based on `sawThinkingOpener`.

- **Tests**: The existing mock streams now include `<thinking>` openers, matching real output shape (all 63 fixtures open with `<thinking>` per the PR description). The new regression test is thorough — covers SSE output, no empty-response error, user message retention, and AI reply persistence.

- **Minor observation (non-blocking)**: `trimmedStart.startsWith('<thinking')` would also match a hypothetical `<thinkingXYZ...` tag, but this is practically impossible from the model and the fallback behavior (hiding it) matches the old behavior, so no concern.

## Verdict

Approve — well-diagnosed root cause, minimal targeted fix, good test coverage, and sensible defensive handling. The Sentry error cluster should clear once this ships.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes a production bug where the streaming handler assumed every model response starts with `<thinking>`, but the model occasionally skips the opener (especially on deep, long-context turns). The entire reply was being swallowed as hidden reasoning, causing an "empty AI response" error that deleted the user message — surfacing as a failed submit.

## Feedback

- **Correctness**: The prefix-matching logic is sound. The `opener.startsWith(trimmedStart.slice(0, opener.length))` check correctly handles ambiguous partial prefixes like `<th` that could still become `<thinking>`, while immediately bailing out when the first non-whitespace chars are clearly not a `<thinking>` opener. The end-of-stream belt-and-suspenders flush (lines +2449–2462 in the diff) is a good safety net.

- **Defense-in-depth**: Routing the non-thinking buffer through the tag trap (`isTrappingTags = true; tagTrapBuffer = thinkingBuffer`) ensures semantic tags (`<draft>`, `<needs>`, etc.) are still caught even when the model skips `<thinking>`. This reuses the existing machinery well.

- **Tests**: The regression test directly reproduces the failure scenario (no `<thinking>`, visible prose split across chunks) and asserts the right outcomes: reply streamed, no empty-response error, user message preserved, AI reply persisted. The existing mock updates to include `<thinking>` openers are correct — they were silently relying on the old "always in thinking trap" assumption.

- **Security**: No risk of leaking chain-of-thought — the `sendCleanText` helper already strips `<thinking>` tags as defense-in-depth (line 2173), and `sawThinkingOpener` ensures that only confirmed `<thinking>` blocks are hidden at end-of-stream.

## Verdict

Approve — clean, targeted fix for a real production bug with solid regression coverage. The approach is conservative (optimistic trap + bail-out) which minimizes risk of breaking the happy path.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes a production bug where the streaming handler's "thinking trap" swallowed the entire AI reply when the model skipped the leading `<thinking>` tag — causing empty-response errors and deleted user messages. The fix detects on the first non-whitespace chunk whether the output is actually a `<thinking>` opener and bails out of the trap if not.

## Feedback

- **Correctness**: The prefix-matching logic in the opener detection is solid. `opener.startsWith(trimmedStart.slice(0, opener.length))` correctly handles full matches, ambiguous partial prefixes (`<th`), and definite non-matches. The `continue` after bail-out correctly routes the buffer through the tag trap on subsequent chunks.

- **Safety (nice)**: The end-of-stream flush now branches on `sawThinkingOpener` — hiding the buffer only when a genuine `<thinking>` was confirmed. This prevents both chain-of-thought leaks *and* the empty-response error that was hitting production. Good defense-in-depth with the belt-and-suspenders comment.

- **Tests**: Existing mock streams updated to include the literal `<thinking>` opener (matching real output shape), and the new regression test is thorough — verifies the reply streams, no empty-response error, user message preserved, and AI reply persisted.

- **Root cause analysis in the PR description**: Excellent. The 297-char buffer vs. 1536 `maxTokens` observation confirming natural stream end (not truncation) is the kind of forensic detail that makes these fixes trustworthy.

## Verdict

Approve — clean, well-scoped fix with solid test coverage and clear production evidence.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes a production bug where the streaming handler assumed every model response opens with `<thinking>`, causing replies to be swallowed when the model skips straight to visible prose — surfacing as repeated failed chat submits.

## Feedback
- **Correctness**: The prefix-matching logic at `messages.ts:2283-2302` is well-designed. The ambiguous-prefix case (`<th` still buffering) correctly handles chunks split across the `<thinking` boundary. The `continue` after bail-out routes accumulated text through the tag trap, so hidden semantic tags (`<draft>`, `<dispatch>`, etc.) are still caught even on no-thinking responses.
- **End-of-stream safety**: The `sawThinkingOpener` guard at `messages.ts:2449-2466` is a good belt-and-suspenders layer — even if PHASE 1 somehow doesn't bail, the flush path handles it rather than throwing an empty-response error.
- **Tests**: Existing mocks updated to include the literal `<thinking>` opener (matching all 63 fixtures) — good hygiene. The new regression test at `messages.test.ts:813-904` covers the exact failure scenario end-to-end: no `<thinking>`, multi-chunk visible reply, verifies streaming output, no empty-response error, no user-message deletion, and AI reply persistence.
- **Sentry alignment**: The PR body's analysis of the 297-char buffer vs. 1536 `maxTokens` cap confirming a natural stream end (not truncation) is a nice forensic detail.

No issues found. Clean, well-scoped fix with solid test coverage.

## Verdict
**Approve** — the root cause is clearly identified, the fix is minimal and correct, and the regression test prevents recurrence.

PR #664, opened 2026-05-25
chore(check-db): guard against silently querying the localhost DB

shantamg:
## Why

During the PR #663 investigation, the `check-db` skill silently fell back to `backend/.env`'s `DATABASE_URL` — which points at a **local** dev DB (a stale snapshot ending ~05-21), not production. That produced a false "telemetry stopped on 05-22" conclusion before I noticed the host was `localhost`. Production is a separate Render Postgres (`be-heard-db`), which had today's data the whole time.

## Change

Adds a pre-query guard to the skill that:
- resolves the DB URL and prints the **host/db only** (credentials redacted) so it's obvious which database is being hit,
- warns loudly when the URL is `localhost`/`127.0.0.1` and the intent was production,
- documents the real prod instance (`be-heard-db`, `dpg-d58660shg0os73bkkpmg-a`) and how to fetch its connection string from the Render API instead of querying localhost.

Skill-only change (`.claude/commands/check-db.md`); no app code touched.

### DIFF
--- a/.claude/commands/check-db.md
+++ b/.claude/commands/check-db.md
@@ -21,6 +21,23 @@ Default to **production read-only** for investigation. A readonly role should on
 
 **SECURITY**: NEVER print full connection strings. Write operations will be rejected by PostgreSQL even if attempted.
 
+### ⚠️ Guard against silently querying the wrong database
+
+`backend/.env`'s `DATABASE_URL` is usually a **local** dev DB (`localhost`), which holds a stale snapshot — NOT production. If you fall back to it without noticing, every result is misleading (this once produced a false "telemetry stopped" conclusion). Before running any query, resolve the URL and confirm the host. Abort and tell the user if it points at localhost when you intended production:
+
+```bash
+DB_URL="${READONLY_DATABASE_URL:-$(grep '^READONLY_DATABASE_URL=' backend/.env 2>/dev/null | cut -d= -f2-)}"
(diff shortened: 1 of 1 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Adds a pre-query guard to the `check-db` skill that prints the resolved DB host (credentials redacted), warns loudly on localhost, and documents the production Render Postgres instance — preventing a repeat of the false "telemetry stopped" conclusion from PR #663.

## Feedback

- **Correctness**: The bash guard snippet correctly resolves `READONLY_DATABASE_URL` first, falls back to `DATABASE_URL`, and the `sed` regex properly extracts `host / db` from a standard Postgres URL. The `case` pattern catches both `localhost` and `127.0.0.1`. Looks solid.
- **Security**: Good — only host/db are printed, credentials stay hidden. The Render instance ID (`dpg-...`) is an internal identifier, not a secret, so it's fine to document here.
- **Placement**: Well-positioned under "Load credentials" and before "Running queries" — the guard runs before any query execution, which is the right flow.
- **Edge case (minor, non-blocking)**: The guard doesn't catch `::1` (IPv6 localhost) or socket connections, but those are unlikely in practice for this project.

## Verdict

**Approve** — Clean, targeted guardrail that prevents a real class of misdiagnosis. No app code touched, just skill documentation.

PR #665, opened 2026-05-25
[codex] Gate status dashboard with Clerk

shantamg:
## Summary
- gates the status dashboard UI with Clerk when a publishable key is configured
- sends Clerk bearer tokens to the brain dashboard API and keeps browser-exposed dashboard secrets dev-only
- adds backend allowlisting by Clerk user ID, keeps email allowlisting, and documents the production monitor deployment

## Validation
- npm --workspace tools/status-dashboard run check
- npm --workspace tools/status-dashboard run build
- npm --workspace backend run test -- brain-auth.test.ts

## Notes
- Production access is intended to be restricted to Shantam and Darryl via DASHBOARD_ALLOWED_USER_IDS and DASHBOARD_ALLOWED_EMAILS.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -63,13 +63,17 @@ ENABLE_AUDIT_STREAM=false
 # =============================================================================
 # Neural Monitor Dashboard (Optional)
 # =============================================================================
-# Secret for dashboard API auth via X-Dashboard-Secret header
+# Secret for dashboard API auth via X-Dashboard-Secret header.
+# Local/dev only; do not set this in production because browser bundles can leak it.
 # Generate a strong random string: openssl rand -hex 32
 # DASHBOARD_API_SECRET=
 # Comma-separated list of emails allowed to access the dashboard (optional)
 # If set, Clerk-authenticated users must have a matching email
 # DASHBOARD_ALLOWED_EMAILS=admin@example.com,dev@example.com
+# Comma-separated list of Clerk user IDs allowed to access the dashboard (optional)
+# Prefer this in production so access does not depend on custom Clerk token claims.
+# DASHBOARD_ALLOWED_USER_IDS=user_...
 # URL of the deployed dashboard (used for CORS)
-# DASHBOARD_URL=https://your-dashboard.example.com
+# DASHBOARD_URL=https://monitor.meetwithoutfear.com
 
 ASSEMBLYAI_API_KEY=
--- a/backend/src/routes/__tests__/brain-auth.test.ts
+++ b/backend/src/routes/__tests__/brain-auth.test.ts
@@ -70,7 +70,19 @@ describe('brain dashboard auth', () => {
       .set('Authorization', 'Bearer valid-token');
 
     expect(res.status).toBe(403);
-    expect(res.body.error).toMatchObject({ code: 'FORBIDDEN', message: 'Email not authorized for dashboard access' });
+    expect(res.body.error).toMatchObject({ code: 'FORBIDDEN', message: 'Not authorized for dashboard access' });
+  });
+
+  it('rejects Clerk tokens whose sub is not in DASHBOARD_ALLOWED_USER_IDS when only user ids are configured', async () => {
+    process.env.DASHBOARD_ALLOWED_USER_IDS = 'user_allowed';
+    verifyTokenMock.mockResolvedValue({ sub: 'user_blocked' } as any);
+
+    const res = await request(createApp())
+      .get('/brain/ably-token')
+      .set('Authorization', 'Bearer valid-token');
+
+    expect(res.status).toBe(403);
+    expect(res.body.error).toMatchObject({ code: 'FORBIDDEN', message: 'Not authorized for dashboard access' });
   });
 
   it('rejects requests without a token when Clerk auth is configured', async () => {
--- a/backend/src/routes/__tests__/brain-auth.test.ts
+++ b/backend/src/routes/__tests__/brain-auth.test.ts
@@ -1,0 +1,85 @@
+import express from 'express';
+import request from 'supertest';
+import brainRoutes from '../brain';
+import { verifyToken } from '@clerk/express';
+
+jest.mock('../../lib/prisma');
+jest.mock('@clerk/express', () => ({
+  verifyToken: jest.fn(),
+}));
+
+const verifyTokenMock = verifyToken as jest.MockedFunction<typeof verifyToken>;
+
+function createApp() {
+  const app = express();
+  app.use('/brain', brainRoutes);
+  return app;
+}
+
+describe('brain dashboard auth', () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    process.env = {
+      ...originalEnv,
+      NODE_ENV: 'production',
+      CLERK_SECRET_KEY: 'sk_test_mock',
+      DASHBOARD_API_SECRET: '',
+      DASHBOARD_ALLOWED_EMAILS: '',
+      DASHBOARD_ALLOWED_USER_IDS: '',
+      ABLY_API_KEY: '',
+    };
+  });
+
+  afterAll(() => {
+    process.env = originalEnv;
+  });
+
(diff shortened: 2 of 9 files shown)


slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary
Gates the status dashboard with Clerk auth (browser side) and adds a `DASHBOARD_ALLOWED_USER_IDS` allowlist on the backend. Good overall direction — Clerk Bearer tokens replace the browser-leaked `VITE_DASHBOARD_SECRET`, docs are updated, and there's test coverage.

## Feedback

- **[Security — blocking]** `backend/src/routes/brain.ts:58-68`: The user-ID allowlist is implemented as a **fast-pass only** — if the ID matches, access is granted immediately. But if it does *not* match, execution falls through. When `DASHBOARD_ALLOWED_EMAILS` is not set, the code reaches the unconditional `next()` (line ~70 in the original), granting access to **any valid Clerk token**, regardless of the user-ID allowlist. This is an authorisation bypass.

  The `.env.example` comment says *"Prefer this in production so access does not depend on custom Clerk token claims"*, implying `DASHBOARD_ALLOWED_USER_IDS` alone is sufficient. But with the current logic, setting only user IDs is insecure.

  **Fix**: After both allowlist checks, if *either* allowlist was configured and the user matched neither, return 403. For example:
  ```ts
  const allowedUserIds = process.env.DASHBOARD_ALLOWED_USER_IDS;
  const allowedEmails = process.env.DASHBOARD_ALLOWED_EMAILS;

  if (allowedUserIds) {
    const ids = allowedUserIds.split(',').map(id => id.trim()).filter(Boolean);
    if (payload.sub && ids.includes(payload.sub)) { next(); return; }
  }

  if (allowedEmails) {
    const emails = allowedEmails.split(',').map(e => e.trim().toLowerCase());
    const email = (payload as any)?.email || (payload as any)?.sub_email;
    if (email && emails.includes(email.toLowerCase())) { next(); return; }
  }

  // If any allowlist was configured but user matched none → deny
  if (allowedUserIds || allowedEmails) {
    errorResponse(res, 'FORBIDDEN', 'Not authorized for dashboard access', 403);
    return;
  }

  // No allowlists configured → allow (already gated by valid Clerk token)
  next();
  ```

- **[Testing — blocking]** `backend/src/routes/__tests__/brain-auth.test.ts`: The rejection test sets *both* `DASHBOARD_ALLOWED_USER_IDS` and `DASHBOARD_ALLOWED_EMAILS`, so it passes. Please add a test where only `DASHBOARD_ALLOWED_USER_IDS` is set and the user's ID is **not** in the list — this will expose the bypass described above.

- **[Nit]** The 403 error message still says *"Email not authorized…"* even when the rejection is based on user ID. Consider a generic message like *"Not authorized for dashboard access"*.

## Verdict
**Request changes** — the user-ID-only allowlist bypass is a real production auth gap. The fix is straightforward (see suggestion above). Everything else looks solid: the Clerk gate in the frontend, the `setAuthTokenProvider` pattern, the `!import.meta.env.PROD` guard on the secret header, and the doc updates are all clean.

slam-paws:
On it — reviewing now.

slam-paws (changes requested):
## Summary
Gates the status dashboard with Clerk auth in the browser, sends Bearer tokens to the backend, adds a `DASHBOARD_ALLOWED_USER_IDS` env var for backend allowlisting, and documents the production deployment. Solid approach overall — the Clerk integration pattern on the frontend is clean.

## Feedback

- **[Security / Blocking] Authorization bypass when only `DASHBOARD_ALLOWED_USER_IDS` is set** (`backend/src/routes/brain.ts:58-68`): If `DASHBOARD_ALLOWED_USER_IDS` is configured but `DASHBOARD_ALLOWED_EMAILS` is not, a user whose ID is *not* in the allowlist silently falls through to the email check — which is unconfigured, so it's a no-op — and reaches `next()` on line 71. Any valid Clerk token gets dashboard access regardless of the user ID allowlist.

  The fix: when the user ID check doesn't match and there's no email allowlist to fall back to, reject with 403. One clean approach:

  ```ts
  const allowedUserIds = process.env.DASHBOARD_ALLOWED_USER_IDS;
  const allowedEmails = process.env.DASHBOARD_ALLOWED_EMAILS;

  if (allowedUserIds || allowedEmails) {
    const userIdAllowList = allowedUserIds?.split(',').map(id => id.trim()).filter(Boolean) ?? [];
    const emailAllowList = allowedEmails?.split(',').map(e => e.trim().toLowerCase()).filter(Boolean) ?? [];

    const userId = payload.sub;
    const email = (payload as any)?.email || (payload as any)?.sub_email;

    const userIdMatch = userId && userIdAllowList.includes(userId);
    const emailMatch = email && emailAllowList.includes(email.toLowerCase());

    if (!userIdMatch && !emailMatch) {
      errorResponse(res, 'FORBIDDEN', 'Not authorized for dashboard access', 403);
      return;
    }
  }
  ```

- **[Testing / Blocking]** The test suite doesn't cover the bypass scenario. Add a test where `DASHBOARD_ALLOWED_USER_IDS` is set (without `DASHBOARD_ALLOWED_EMAILS`) and the token's `sub` is *not* in the list — this should return 403 but currently returns 503 (passes through to the route handler).

- **[Nit]** `backend/src/routes/brain.ts:22` — The JSDoc still says "Clerk JWT Bearer token (verified via @clerk/express)" on the first line of the list; the update on lines 21-22 adds the allowlist detail, but you might also want to drop "If neither DASHBOARD_API_SECRET nor CLERK_SECRET_KEY is configured, skips auth (dev mode)." from line 23 since the middleware now rejects in production when auth isn't configured — the comment is slightly stale.

## Verdict
**Request changes** — the user-ID-only allowlist bypass is a real auth gap for the intended production config (where the PR body says `DASHBOARD_ALLOWED_USER_IDS` + `DASHBOARD_ALLOWED_EMAILS` are set, but only user IDs might be set in some deployments). The fix is small; everything else looks good.

shantamg:
Addressed the requested changes in `d645fdc1`:

- fixed the allowlist logic so if either `DASHBOARD_ALLOWED_USER_IDS` or `DASHBOARD_ALLOWED_EMAILS` is configured, the Clerk token must match one of them
- changed the 403 copy to `Not authorized for dashboard access`
- added a regression test for the user-ID-only non-match case
- updated the middleware comment for prod vs local auth behavior

Validated locally:
- `npm --workspace backend run test -- brain-auth.test.ts`
- `npm --workspace tools/status-dashboard run check`
- `npm --workspace tools/status-dashboard run build`

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Gates the status dashboard with Clerk auth in production (Bearer tokens to the backend, `<SignedIn>`/`<RedirectToSignIn>` in the browser) and adds `DASHBOARD_ALLOWED_USER_IDS` alongside the existing email allowlist. Prevents `VITE_DASHBOARD_SECRET` from being sent in production builds.

## Feedback
- **Security (good)**: `api.ts` now blocks `VITE_DASHBOARD_SECRET` in prod (`!import.meta.env.PROD` guard) — solid protection against accidental secret leakage in browser bundles.
- **Auth logic (good)**: Backend allowlist is OR-based (user ID *or* email match) with proper `filter(Boolean)` to handle trailing commas. Case-insensitive email comparison preserved.
- **Tests (good)**: `brain-auth.test.ts` covers allow-by-ID, allow-by-email, reject-neither, reject-wrong-ID, and no-token scenarios — solid coverage.
- **Note (non-blocking)**: If `CLERK_SECRET_KEY` is set but *neither* `DASHBOARD_ALLOWED_USER_IDS` nor `DASHBOARD_ALLOWED_EMAILS` is configured, any valid Clerk user gets through. Pre-existing behavior and the docs clearly say to set the allowlists in production, so this is fine — just flagging for awareness.
- **Frontend (clean)**: `ClerkGate` pattern with `useLayoutEffect` to register the token provider before first render is the right approach. Graceful fallback to unauthenticated routes when no Clerk key is set keeps local dev ergonomic.

## Verdict
Approve — well-structured auth hardening with good test coverage, clear docs, and no security concerns.

PR #666, opened 2026-05-25
[codex] fix Stage 4 need-id coverage

shantamg:
## Summary
- Treat explicit Stage 4 need IDs in proposal.needsAddressed as full coverage matches.
- Include StrategyProposalNeed link rows when refreshing Stage 4 need coverage.
- Add regression coverage for both need-ID and join-table coverage paths.

## Root cause
The share gate reads Stage4NeedCoverage.coveringProposalIds, but coverage refresh only matched proposal needs by text. A proposal could display as addressing a need because it stored the need ID, while the coverage audit still kept the same need OPEN with no covering proposals.

## Production repair
Refreshed Stage 4 coverage for session cmpc3ra1g0097nw1yu40bb1q6 after applying the fixed classifier locally. Result: covered=2, partial=1, open=0.

## Validation
- npm test -- --runInBand backend/src/services/__tests__/stage4-coverage.service.test.ts
- npm run check

### DIFF
--- a/backend/src/services/__tests__/stage4-coverage.service.test.ts
+++ b/backend/src/services/__tests__/stage4-coverage.service.test.ts
@@ -142,6 +142,65 @@ describe('stage4-coverage.service', () => {
     expect(result).toEqual({ covered: 0, partial: 0, open: 2, total: 2 });
   });
 
+  it('covers a need when a proposal stores the need id in needsAddressed', async () => {
+    (prisma.identifiedNeed.findMany as jest.Mock).mockResolvedValue([
+      need({
+        id: 'need-authenticity',
+        need: "I need authenticity in our relating — the knowledge that I'm honoring him and he's honoring me",
+      }),
+    ]);
+    (prisma.strategyProposal.findMany as jest.Mock).mockResolvedValue([
+      proposal({
+        id: 'proposal-honesty',
+        description: 'When we talk, Shantam will commit to being honest about where things stand',
+        needsAddressed: ['need-authenticity'],
+      }),
+    ]);
+
+    const result = await refreshStage4NeedCoverage(sessionId);
+
+    expect(prisma.stage4NeedCoverage.createMany).toHaveBeenCalledWith({
+      data: [
+        expect.objectContaining({
+          needId: 'need-authenticity',
+          coverageStatus: 'COVERED',
+          coveringProposalIds: ['proposal-honesty'],
+        }),
+      ],
+    });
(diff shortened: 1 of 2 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes Stage 4 need coverage classification to recognize proposals that reference needs by ID (via `needsAddressed` array or `StrategyProposalNeed` join table), not just by text similarity matching.

## Feedback
- **Correctness**: The fix correctly adds two coverage paths — direct need-ID match in `needsAddressed` and join-table `needLinks` lookup. The `linkedByNeedId` variable is computed but then the existing `addressed` check _also_ picks up `label === need.id` (line ~157 in the patched file). This means need-ID matches via `needsAddressed` are detected in both `linkedByNeedId` and `addressed` — redundant but not harmful, just belt-and-suspenders.
- **Types**: `needLinks?: Array<{ needId: string }>` optional field on `ProposalRow` is clean and matches the Prisma `include` shape.
- **Tests**: Two solid regression tests covering both the need-ID-in-array and join-table paths. Good use of realistic test data.
- **Database**: Uses `include` on an existing query (no migration needed) — appropriate.
- **Performance**: The `include: { needLinks: { select: { needId: true } } }` is well-scoped; only fetches what's needed.

## Verdict
Approve — targeted fix with good test coverage for the root cause. Ship it.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes Stage 4 need coverage classification to recognize explicit need IDs in `proposal.needsAddressed` and `StrategyProposalNeed` link rows as full coverage matches (score = 1), rather than relying solely on text similarity.

## Feedback
- **Correctness**: Logic is sound. The `linkedByNeedId` flag correctly checks both `needsAddressed.includes(need.id)` and the `needLinks` join table. The `addressed` check also adds `label === need.id` for belt-and-suspenders safety — minor overlap with `linkedByNeedId`, but harmless and arguably clearer.
- **Types**: `needLinks?: Array<{ needId: string }>` on `ProposalRow` is properly optional, matching the Prisma `include` shape. The cast to `ProposalRow[]` on the query result is consistent with the existing pattern.
- **Tests**: Two new test cases cleanly cover both paths (need-ID in `needsAddressed` array, and `StrategyProposalNeed` link row). Good use of the existing `need()` / `proposal()` helpers.
- **Database**: No migration needed — this is a read-path fix using an existing join table via `include`.

## Verdict
Approve — clean, focused fix with good regression coverage. No issues found.

PR #668, opened 2026-05-26
Fix session waiting state UI

shantamg:
## Summary
- hide Stage 2 chat input while waiting after empathy validation, even during optimistic cache windows
- clear the empathy validation typing indicator once the follow-up AI message is present
- show guided action panels above the composer and suppress stale share offers outside Stage 2

## Verification
- `npm test -- --runTestsByPath src/utils/__tests__/getWaitingStatus.test.ts src/utils/__tests__/chatUIState.test.ts src/components/__tests__/ShareTopicPanel.test.tsx src/components/__tests__/ChatInterface.test.tsx`
- `npm run check` in `mobile`
- `npm test -- --runTestsByPath src/controllers/__tests__/stage2-copy.test.ts`
- `npm run check` in `backend`

### DIFF
--- a/backend/src/controllers/notifications.ts
+++ b/backend/src/controllers/notifications.ts
@@ -70,15 +70,25 @@ export async function getPendingActionsHandler(
 
     const actions: PendingAction[] = [];
 
-    // 1. Pending share offers (reconciler suggested sharing)
-    const shareOffers = await prisma.reconcilerShareOffer.findMany({
-      where: {
-        userId: user.id,
-        result: { sessionId },
-        status: { in: ['PENDING', 'OFFERED'] },
-      },
-      include: { result: true },
+    const userStageProgress = await prisma.stageProgress.findFirst({
+      where: { sessionId, userId: user.id },
+      orderBy: { stage: 'desc' },
+      select: { stage: true },
     });
+    const userMaxStage = userStageProgress?.stage ?? 0;
+
+    // 1. Pending share offers (reconciler suggested sharing). These are only
+    // actionable while the user is still in Stage 2.
+    const shareOffers = userMaxStage === 2
+      ? await prisma.reconcilerShareOffer.findMany({
+          where: {
+            userId: user.id,
+            result: { sessionId },
+            status: { in: ['PENDING', 'OFFERED'] },
+          },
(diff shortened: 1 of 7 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes three related UI issues: (1) hides chat input during the empathy validation waiting state even when optimistic cache lags behind, (2) clears the empathy validation typing indicator once the follow-up AI message arrives, and (3) moves guided action panels above the composer while suppressing stale share offers outside Stage 2.

## Feedback

- **Correctness (backend)**: Hoisting the `stageProgress` query in `notifications.ts` to reuse it for both the share-offer gate and the shared-context gate is a clean refactor — eliminates a duplicate DB query while adding the Stage 2 gate logic. Same pattern applied correctly in `getBadgeCountHandler`.

- **Correctness (backend)**: The new guard in `stage2.ts:getShareSuggestion` (`progress?.stage !== 2`) properly returns `{ hasSuggestion: false }` instead of querying the reconciler for users who have moved past Stage 2. The `status: { in: ['IN_PROGRESS', 'GATE_PENDING'] }` filter is appropriately scoped.

- **Correctness (mobile)**: The relaxed `getWaitingStatus` condition (replacing `myAttemptStatus === 'REVEALED'` with `hasPartnerEmpathy`) is well-justified — during the optimistic cache window, `myAttemptStatus` can still be `'READY'` even though the user has already validated. The surrounding conditions (`myStage === PERSPECTIVE_STRETCH`, `myValidation?.validated`, `partnerValidated === false`) are sufficient to keep this precise.

- **Correctness (mobile)**: The `hasEmpathyValidationFollowUp` memo in `UnifiedSessionScreen.tsx` correctly detects AI messages that arrive after the validation timestamp, and the `useEffect` properly clears the awaiting state. The `Number.isFinite` guards are a good defensive touch.

- **Tests**: New test case for the optimistic-validation race condition covers the exact scenario being fixed. Existing test updated to match the `auxiliaryControls` reordering.

## Verdict
Approve — focused, well-tested fix for a real UI race condition. The backend stage-gating is a good defensive layer that prevents stale share offers from leaking into later stages.

PR #681, opened 2026-05-30
Fix invitation follow-up recovery

### LINKED ISSUE #677: AI message doesn't arrive after invitation sent; typing indicator stuck
## Bug Report

After starting a new session and sending the invitation, then closing the popup:
• The "invitation ready" line appears
• The typing indicator (3 dots) shows up
• But the AI message never arrives

The user had to leave and re-enter the session for it to refresh and show the message. This suggests the Ably realtime channel may not be delivering the message to the client, or the client isn't processing it while the popup is being dismissed.

Shantam notes this is an intermittent issue — "the only time in a while" he's seen it.

## What We Know So Far

• Likely related to Ably subscription timing — the client may miss the message event if the popup dismiss and the Ably publish happen in a narrow window
• Could also be a React Query cache issue where the optimistic update from closing the popup interferes with the incoming Ably message
• No immediate leads from codebase — needs full investigation

## Provenance

- **Channel**: #bugs-and-requests
- **Reporter**: Shantam (U0AD3TF2U7L)
- **Original message**: "Tested starting a new session and after the invitation sending, I closed the popup and it showed the invitation ready line and then I saw the 3 dots but the message never came in. I had to go out and come back into the session for it to refresh. That's the only time in a while that I've seen that front end issue. Everything else has been smooth."
- **Timestamp**: 1780170920.878379

shantamg:
## Summary

- Update invitation confirmation cache state so the session status moves from `CREATED` to `INVITED` after the backend advances the inviter to Stage 1
- Add bounded retry recovery for missed fire-and-forget AI responses instead of relying on a single delayed refetch
- Add a regression test proving the cached session status changes to `INVITED` after invitation confirmation

## Why

PR #678 mitigates the stuck typing indicator by doing a one-shot refetch after 10 seconds. That helps only when the AI message has already been persisted by then. The deeper issue is that the mobile cache can keep `session.status` as `CREATED`, which disables realtime subscriptions while the client is waiting for the Ably `message.ai_response` event.

This fix keeps realtime enabled by updating the cached status immediately, and retains a bounded persisted-message recovery path for missed Ably events or slow background generation.

## Validation

- `npm --workspace mobile test -- --runTestsByPath src/hooks/__tests__/useSessions.test.ts --runInBand --forceExit`

Fixes #677

### DIFF
--- a/mobile/src/hooks/__tests__/useSessions.test.ts
+++ b/mobile/src/hooks/__tests__/useSessions.test.ts
@@ -16,9 +16,10 @@ import {
   useInvitation,
   useAcceptInvitation,
   useDeclineInvitation,
+  useConfirmInvitationMessage,
   sessionKeys,
 } from '../useSessions';
-import { SessionStatus } from '@meet-without-fear/shared';
+import { SessionStatus, StageStatus } from '@meet-without-fear/shared';
 
 // Import mocked functions
 import * as api from '../../lib/api';
@@ -50,6 +51,11 @@ function createWrapper(): React.FC<{ children: React.ReactNode }> {
       },
     },
   });
+  return ({ children }) =>
+    React.createElement(QueryClientProvider, { client: queryClient }, children);
+}
+
+function createWrapperWithClient(queryClient: QueryClient): React.FC<{ children: React.ReactNode }> {
   return ({ children }) =>
     React.createElement(QueryClientProvider, { client: queryClient }, children);
 }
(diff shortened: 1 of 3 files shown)


slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary

Fixes the stuck typing indicator after invitation confirmation by (1) updating cached `session.status` to `INVITED` so realtime subscriptions stay enabled, and (2) adding bounded polling (up to 6 attempts at 10s intervals) to recover missed fire-and-forget AI responses.

## Feedback

- **Correctness (cache fix)**: The root cause is well-identified — `session.status` staying `CREATED` disables Ably subscriptions. Adding `status: SessionStatus.INVITED` in the `useConfirmInvitationMessage` `onSuccess` handler (`useSessions.ts:693`) is the right fix. Good that it's gated behind `data.advancedToStage !== undefined`.

- **Correctness (recovery polling)**: The bounded retry in `UnifiedSessionScreen.tsx:1396-1420` is well-structured. The existing latch-clearing effects (`hasInvitationTransitionResponse` / `hasEmpathyValidationFollowUp`) will clear the awaiting flags as soon as the AI message lands in cache, which stops the typing indicator and — since the `useEffect` depends on those flags — the cleanup function will fire and clear the interval. The max-attempts cap at 6 is a good safety net.

- **Minor observation**: When max attempts is reached (line 1410-1412), both `isAwaitingInvitationFollowUp` and `isAwaitingEmpathyValidationFollowUp` are cleared regardless of which one triggered the effect. This is fine in practice since both being stuck simultaneously is unlikely, and clearing both is the safer default. Just noting for awareness.

- **Tests**: Good regression test covering the cache status update. The test correctly seeds `CREATED` status, mocks the confirmation response with `advancedToStage: 1`, and asserts the cached status becomes `INVITED`.

## Verdict

Approve — clean, well-scoped fix with good test coverage. The two-pronged approach (immediate cache fix + bounded recovery polling) is solid.

slam-paws:
On it — reviewing now.

slam-paws (approved):
## Summary
Fixes the stuck typing indicator after invitation confirmation by (1) immediately updating the cached session status from `CREATED` → `INVITED` so Ably subscriptions stay active, and (2) adding a bounded retry (6 × 10s) to recover missed fire-and-forget AI responses.

## Feedback
- **Correctness**: The one-line cache fix in `useSessions.ts:693` is the right approach — it aligns cached state with what the backend has already done, keeping realtime channels open. Good catch on the root cause.
- **Recovery logic**: The bounded polling in `UnifiedSessionScreen.tsx:1396-1419` is well-designed — constants are extracted, the interval is properly cleaned up, and it caps at 60s to avoid indefinite polling.
- **Minor note**: When max attempts are reached (line 1413-1414), both `isAwaitingInvitationFollowUp` *and* `isAwaitingEmpathyValidationFollowUp` are cleared together. If both flags were true simultaneously, the recovery for the second scenario also terminates. This is very unlikely in practice but worth noting.
- **Tests**: Good regression test covering the cache state transition after mutation.

## Verdict
Approve — clean, focused fix with good test coverage. The root cause (cached status blocking realtime) is correctly addressed.

